PR #274, opened 2014-09-12
File compression support in ArticleStorageFS

pboguslawski:
https://dev.ib.pl/ib/otrs-ib/issues/5

Please merge to master if you find it useful (we don't use master for devel now and will try to be more compatible in the future).


xhoy:
+1!

PR #332, opened 2014-11-07
[Master Bug 10853] State and StateType search attributes in customer info are not shown in Search screen

s7design:
Hi,
There is fix for master, I will make another one for 3.3.


s7design:
This PR is closed, it will be rebased and done with more tests

Regards
Zoran

PR #339, opened 2014-11-17
[Master - Bug 9460] Under some circumstances OTRS does not join Tickets to the Customernumber

s7design:
Hi Martin,

There is added CustomerIDRaw and CustomerLoginUserRaw too. I believe that all module work well with new search attributes.
Because there is CustomerUserLoginRaw, now is not allowed wildcard hacking the customer information center.
If is QA ok here, we will be able to solve two bugs with this new features (#9460 and #10534). 

Regards
Zoran


s7design:
I close this PR, it will be solved on master with more tests

Regards
Zoran

PR #346, opened 2014-11-22
New frontend agent ticket zoom menu module cluster

nilei:
This is especially useful if you have a lot of ticket actions enabled. The ticket actions can be clustered into free configurable menus. The most important actions may stay as single buttons while the less frequently actions are available in a separated menu (e.g. called "More..."). The OpenMenuOnHover setting is also considered to ease the access to the sub menus without additional clicks.

Example screenshot:.

screen shot 2014-11-22 at 02 37 11

-- Cheers, Nils


nilei:
Wouldn't it be okay to start with my simple manually configurable approach instead of let this PR die? As soon as this is implemented, we can receive and review further PRs to optimise the behaviour, add optional automatic clustering based on XYZ (maybe a module layer) and also add configuration options on user level later.

mrcbnsls:
Hi @nilei,

You're right, we shouldn't let this die. The idea is still very nice. So we'd suggest to do it like you suggested. We should try to keep it as simple as possible for the first approach, so we should just have it configurable. Do you think there is a useful default configuration for modules which should go into the submenu which is valid for most OTRS users? If not, please leave the default empty so users (admins) can decide on their own.

Additionally, I'd ask you to change the technique for opening the submenu, please. It should just open on hovering by using pure CSS.

Marc

nilei:
Hi @mrcbnsls, as suggested I have remove the java script dependency and added some small CSS to display the sub menu on hover. However, I can't get the z-index fixed. The sub menu is always hidden by the process information or the article list. Can you have a look into the CSS to see where to place the z-index correctly?

-- Cheers, Nils

nilei:
@mrcbnsls, could you have a look why the drop down menu is hidden by the other elements?

TIA

-- Cheers, Nils

mrcbnsls:
This has now been merged in 8ce6b49 and will be part of OTRS 5. Thanks a lot for this nice contribution! I like it a lot that the ticket zoom looks much cleaner due to this change.

I did some small changes to the look and feel and we configured some default groups. You will anyway be able to specify your own groups as the administrator, but we thought it would be a good idea to have default ones.

<a href="https://cloud.githubusercontent.com/assets/4690172/8743135/ce73e54e-2c6c-11e5-8f17-008ae8f9b6de.png" target="_blank">menu</a>

Marc

nilei:
Hi @mrcbnsls, 

the clusters are not displayed in mobile mode. May you consider to enable them for mobile mode or implement another solution for the mobile mode. I could think of a single button like the one for the main navigation which then displays all possible ticket actions vertically.

-- Cheers, Nils

PR #352, opened 2014-12-01
Add ticket owner view in custom queues

noritnk:
Ticket queue view is very useful, but we also want to look at number of tickets ownered by each member in custom queues.  Especially, a manager of a queue will need it.  So I wrote Ticket Owner Overview to fill that by customization of Ticket Queue Overview.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -6016,7 +6016,7 @@
             </Hash>
         </Setting>
     </ConfigItem>
-     <ConfigItem Name="DashboardBackend###0270-TicketQueueOverview" Required="0" Valid="1">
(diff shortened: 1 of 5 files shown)


noritnk:
Hi,

I found issue on permission, and fixed it.

Best Regards,
@noritnk

mgruner:
Hello @noritnk, thanks a lot for your contribution. At this point I would not want to merge it, because I am not fully convinced that this is a very common use case (other people might want to look at tickets by owner without the custom queue limitation or even by queue). Besides, the PR is not fully complete (links in dashboard don't contain the Queue limitation; search screen GUI does not understand about a "my queues" limitation). I am also not sure if such a feature could be turned on by default as there could be privacy/data protection problems in some countries like Germany; this would have to be clarified.

I'm not totally against merging such a feature, but we would have to take some more steps to get there.

Again, thanks for your help so far!

Best regards, Martin

PR #359, opened 2014-12-03
[Bug 10720] No "Available Columns" 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10720

Hi Martin,

I have fixed this in Dashboard screen. As I have  mentioned before, there is not a problem in AgentTicketSearch and the others screens with "Settings filter options". There was a problem only for Dashboard filter settings.  
There is a little different logic for update those preferences, because it works with AJAX, and I found  some solution, please follow-up this.
If I can improve it, let me know.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/DashboardTicketGeneric.pm
+++ b/Kernel/Output/HTML/DashboardTicketGeneric.pm
@@ -344,6 +344,11 @@ sub Preferences {
         }
     }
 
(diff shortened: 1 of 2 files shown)


s7design:
I close it, it will be solved and more tested on master

Regards
Zoran

PR #365, opened 2014-12-05
[Rel-4_0 - Bug 10816] Apache2 with mod_deflate enabled will not start

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10816


mgruner:
Merged, thanks!

PR #366, opened 2014-12-05
[rel-4_0_Bug_10737] Fixed bug with adding CustomerCompany to second backend

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10737


s7design:
This is solved in:

https://github.com/OTRS/otrs/commit/3e50777f7e34c6f917d17ccbc12a8752b876603e

Zoran

PR #370, opened 2014-12-07
make minimum log level configurable. 

reneeb:
This helps to reduce logfile size.


mgruner:
Merged with small fix in 43f81e8. Thanks!

PR #371, opened 2014-12-07
Let LWP::UserAgent handle https-proxy commands

mbeijen:
This also makes it possible to do proper SSL certifcate verification when https URLs are used and a proxy is configured.

For accessing HTTPS resources via a proxy server, HTTP CONNECT is used only LWP did not support this properly. See https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_tunneling This is why we had the workaround of using Net::SSL in place.

Only, since LWP 6.06 LWP now supports this properly. See https://metacpan.org/changes/distribution/libwww-perl 
OTRS already ships LWP 6.06. This means the whole Net::SSL workaround, and the cruft around this, can be removed!

### DIFF
--- a/Kernel/System/WebUserAgent.pm
+++ b/Kernel/System/WebUserAgent.pm
@@ -137,77 +137,75 @@ sub Request {
 
     my $Response;
 
-    {
+    # init agent
+    my $UserAgent = LWP::UserAgent->new();
 
-        # init agent
-        my $UserAgent = LWP::UserAgent->new();
-
-        # In some scenarios like transparent HTTPS proxies, it can be neccessary to turn off
-        #   SSL certificate validation.
-        if ( $Kernel::OM->Get('Kernel::Config')->Get('WebUserAgent::DisableSSLVerification') ) {
-            $UserAgent->ssl_opts(
-                verify_hostname => 0,
-            );
-        }
-
-        # set credentials
-        if ( $Param{Credentials} ) {
-            my %CredentialParams    = %{ $Param{Credentials} || {} };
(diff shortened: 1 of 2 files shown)


mbeijen:
I've left the DisableSSLVerification setting now. I'm not sure in what scenario you'd need it though as transparent proxies are really supposed to leave SSL certificates in place!

jensoliver:
There are several setups in high security areas using this kind of setup to scan even https requests.

mbeijen:
Cool! I was not aware of that.

Kernel/System/WebUserAgent.pm
```diff
@@ -137,109 +137,75 @@ sub Request {
...
-        if ( $Kernel::OM->Get('Kernel::Config')->Get('WebUserAgent::DisableSSLVerification') ) {
-            my $Loaded = $Kernel::OM->Get('Kernel::System::Main')->Require(
-                'Net::SSLeay',
-                Silent => 1,
-            );
-            if ($Loaded) {
-                $UserAgent->ssl_opts(
-                    verify_hostname => 0,
-                );
-                $UserAgent->ssl_opts(
```
mgruner: Hello @mbeijen , did you drop this on purpose?

mbeijen: Yeah, there's no need for Net::SSLeay anymore so it's OK just to set ssl_opts' `verify_hostname`.

mgruner:
Merged in 33a8df7 (master only). Thanks a lot, Mike!

PR #372, opened 2014-12-08
show 'add agent' and 'add group' action in UserGroup overview

reneeb:
I find it always annoying when I have to add several agents via web-gui that I always have to go the "full path" for adding agents. With these to actions shown in the overview, it's just one click...


mrcbnsls:
Merged in 094bd62, thanks.

PR #374, opened 2014-12-09
[Rel-4_0 - Bug 10895] Dynamic Field shown in ticket information although it isn' t used

s7design:
Hi men,

I fixed it here for customer like Carlos suggested. Now, we have unique behavior for all screens.
There is the same situation in Ticket overview medium  and large as well.
I believe that in this moment it is better solution.
If we have any suggestion let me know.

Regards
Zoran


carlosfrodriguez:
Hi @s7design 

This pull request was merged in https://github.com/OTRS/otrs/commit/72ff7aa1e34f59d64b7424922bb60ff3e8cd73e5

Many thanks for your help.

((enjoy))
cr

PR #381, opened 2014-12-12
Ported to OM 12_12

s7design:


mrcbnsls:
Merged in 69ba9fe, excellent! I really love the fact that you added the UT! That eases up testing a lot. I had to make some smaller changes on some places, maybe you want to compare my merge commit with your work to see where.

PR #384, opened 2014-12-15
Ported to OM 15_12

s7design:


mrcbnsls:
Merged in 9db8f66, good job!

PR #390, opened 2014-12-17
Master port 17 12

s7design:


mrcbnsls:
Merged in acbc65d with only one small change. Thank you!

PR #393, opened 2014-12-19
add new config option and change redirect after a user was created

reneeb:
Currently the "next screen" after a user was created is determined by the activated frontend modules. But this is sometimes not sufficient as an admin wants to create several users before he assigns them to groups/roles or the user<->group frontend is active but permissions are regulated *mainly* with roles.

This patch adds a new config option to define which screen should be redirected to.

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -6833,4 +6833,16 @@ via the Preferences button after logging in.
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="AddUserNextScreen" Required="0" Valid="0">
+        <Description Translatable="1"></Description>
+        <Group>Framework</Group>
+        <SubGroup>Core</SubGroup>
+        <Setting>
+            <Option SelectedID="UserGroup">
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Hi @reneeb, I totally agree that the current redirect is not really helpful in most situations. I like your approach, but we currently try to actually *avoid* new config settings if they're not really needed. What do you think about a solution where the admin would just be redirected to the overview of agents (AdminUser without any subaction)? This would be similar to other screens.

And to help with the further navigation for the admin: In AdminCustomerUser we have the nofitication bar on top of the screen which tells you if the customer has been added successfully. It contains two links for accessing AgentTicketEmail or AgentTicketPhone for this customer. How about adding something similar for agents to either open AdminUserGroup or AdminRoleUser for this agent?

reneeb:
Sounds good for me.

mrcbnsls:
Fine! Are you willing to implement it? :-D

reneeb:
I won't have time to implement it until February...

mrcbnsls:
This should be okay. Please feel free to open a new PR whenever you find the time for it. Thanks in advance!

PR #394, opened 2014-12-21
OTRS plugin for GitHub

cemtopkaya:
Is there any plugin for GitHub which provide to create issue with the OTRS ticket id inside GitHub?
I think there is Redmin plugin for this but GitHub and OTRS integration would be great ...


mrcbnsls:
Hi @cemtopkaya, sorry, I don't really understand this request. Also, there seems to be something wrong with this PR. Please use the PR mechanism only for real changes to the OTRS codebase. If you have any more questions, please contact sales@otrs.com or file bug reports at bugs.otrs.org. Thanks.

PR #396, opened 2014-12-22
[Rel-4_0 - Bug 10117] Filter in queue view for customer userID is not completed

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10117

### DIFF
--- a/Kernel/Modules/AgentCustomerInformationCenterSearch.pm
+++ b/Kernel/Modules/AgentCustomerInformationCenterSearch.pm
@@ -69,16 +69,32 @@ sub Run {
     my $LikeEscapeString   = $Self->{DBObject}->GetDatabaseFunction('LikeEscapeString');
 
     my $MaxResults = $AutoCompleteConfig->{MaxResultsDisplayed} || 20;
+    my $IncludeUnknownTicketCustomers = int( $Self->{ParamObject}->GetParam( Param => 'IncludeUnknownTicketCustomers' ) || 0 );
     my $SearchTerm = $Self->{ParamObject}->GetParam( Param => 'Term' ) || '';
 
     if ( $Self->{Subaction} eq 'SearchCustomerID' ) {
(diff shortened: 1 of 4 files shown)


mgruner:
Hi S7,
thanks for your proposal. I believe the bug is actually about something different. I'll close this PR for now and ask for more clarification in the bug.
Regards, mg

mgruner:
Hi S7,
thanks so far for your proposal. Adding this globally to AgentCustomerSearch/AgentCIC could have unwanted side effects for us, so I suggest that you add a new optional mode "IncludeUnknownTicketCustomers" that you only send from the few filter places where we actually want it.
Your approach to unifying the lists will not scale well, grep will get slower and slower the more customers are found. In CIC you can see that first the customer list and customer id list are fetched and then merged with a hash lookup. I'd suggest that you fetch the ticket customers (if mode active) before that too and then just unify the three lists at once. That will also fix the broken sorting. Please optimize this for both modules.
Regards, mg

s7design:
Hi @mgruner 

I have changed code as you suggested. Now, unknown customer, that there are in tickets, will be added only for filters on dashboard and Ticket view screens (Status, Queue...). Actually, I wanted such effect. I hope it is good. If you think that I can improve code more, let me know please.

Regards
Zoran

s7design:
Hi @mgruner 

I saw that there are a bug here with CustomerID filters and MinQueryLenght, I think that there was typo. You can see the fix in the last commit. I have some ideas for improvement this more, but your opinion will be welcomed in this moment. Maybe we can move DB logic for fetching Customer data from CIC and AgentCustomerSearch to System module for example Ticket.pm
What do you think about that?

Regards
Zoran

s7design:
Hi @mgruner,

I make proposal PR there  https://github.com/s7design/otrs/pull/16. Maybe on this way it could be better fixed.

Regards
Zoran

s7design:
I close this PR, it will be rebased and solved on master branch.

Regards
Zoran

s7design:
.

PR #398, opened 2014-12-23
[Rel-4_0 Bug 10837 ] Reply mail into"cc" instead of "to"

s7design:


carlosfrodriguez:
Hi @s7design,

Thank you for this pull request, it was merged in https://github.com/OTRS/otrs/commit/80fbe691cfde5732dd60c9add51fea2f06e83e9a.

I just change the first RegEx to a direct compare, as I believe that the RegEx is not really needed since the value is gather from the DB.

((enjoy))
cr

PR #399, opened 2014-12-23
Dashlet: Filter Attributes with more then one CustomerID doesn't work

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10968

### DIFF
--- a/Kernel/Output/HTML/DashboardTicketGeneric.pm
+++ b/Kernel/Output/HTML/DashboardTicketGeneric.pm
@@ -1978,6 +1978,10 @@ sub _SearchParamsGet {
         next STRING if !$String;
         my ( $Key, $Value ) = split /=/, $String;
 
+        if ( $Key eq 'CustomerID' ) {
+            $Key = "CustomerIDRaw";
+        }
+
         # push ARRAYREF attributes directly in an ARRAYREF
         if (
             $Key


Kernel/Output/HTML/DashboardTicketGeneric.pm
```diff
@@ -1978,6 +1978,10 @@ sub _SearchParamsGet {
         next STRING if !$String;
         my ( $Key, $Value ) = split /=/, $String;
 
+        if ( $Key eq 'CustomerID' ) {
+            $Key = "CustomerIDRaw";
+        }
+
```
mrcbnsls: Hi @s7design, thanks for the fix. Could you please explain a bit more how this change is related to the linked bug? Thanks.

s7design:
Hi @mrcbnsls,

There was a problem with quote CustomerIDs as @carlosfrodriguez mentioned in previous PR for this bug. I suppose that you have seen our conversation ( https://github.com/OTRS/otrs/pull/397 )

Without my fix there was problem here on line 542:
        # quote all CustomerIDs
        if ( $TicketSearch{CustomerID} ) {
            $TicketSearch{CustomerID} = $Kernel::OM->Get('Kernel::System::DB')->QueryStringEscape(
                QueryString => $TicketSearch{CustomerID},
            );
        }

and because in reporter's test case there are two customer, TicketSearch{CustomerID}  become quoted Array reference  instead quoted each value in array, actuary we get wrong result.

There it was possible fix this if I chage cod above as something like

        # quote all CustomerIDs
        if ( $TicketSearch{CustomerID} ) {
            if (ref $TicketSearch{CustomerID}) {
                my @TicketSearchCustomerID = @ {$TicketSearch{CustomerID}};
                delete $TicketSearch{CustomerID};
                for my $Item  ( @TicketSearchCustomerID ) {

                    $Self->{LogObject}->Log(
                        Priority => "error",
                        Message  => ">>> $Item"
                    );
              
                    push @{ $TicketSearch{CustomerID} }, $Kernel::OM->Get('Kernel::System::DB')->QueryStringEscape(
                        QueryString => $Item );     
                }
              
            }
            else{
                $TicketSearch{CustomerID} = $Kernel::OM->Get('Kernel::System::DB')->QueryStringEscape(
                QueryString => $TicketSearch{CustomerID},
            );
            }
            
        }

carlosfrodriguez:
Hi @s7design 
This pull request was merged in https://github.com/OTRS/otrs/commit/fb8638c36a9edf19791fcec82717c1a4b034aebd, unfortunately it was not possible to implement on rel-3_3 so I had to go for a similar solution from the last comment for that branch.

((enjoy))
cr

PR #403, opened 2014-12-31
[Rel-4_0 bug 10961 ] Adding email recipients via addressbook does not update customer information

s7design:

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.TicketAction.js
+++ b/var/httpd/htdocs/js/Core.Agent.TicketAction.js
@@ -91,12 +91,21 @@ Core.Agent.TicketAction = (function (TargetNS) {
      */
     function AddMailAddress($Link) {
         var $Element = $('#' + $Link.attr('rel')),
-            NewValue = $Element.val();
+            NewValue = $Element.val(),
+            NewData;
         if (NewValue.length) {
             NewValue = NewValue + ', ';
         }
         NewValue = NewValue + Core.Data.Get($Link.closest('tr'), 'Email');
         $Element.val(NewValue);
+        if (typeof $Element.attr("data-customerkey") !== 'undefined' ){
+            NewData = $Element.attr("data-customerkey");
+            if(NewData.length){
+                NewData = NewData + ', ';
+            }
+            NewData = NewData + $Link.data('customerkey');
+            $Element.attr("data-customerkey",NewData);
+        }
     }
 
     /**
@@ -269,8 +278,9 @@ Core.Agent.TicketAction = (function (TargetNS) {
                 $Bcc = $('#BccCustomer', parent.document);
 
                 $.each($('#ToCustomer').val().split(/, ?/), function(Index, Value){
-                    $To.val(Value);
-                    parent.Core.Agent.CustomerSearch.AddTicketCustomer( 'ToCustomer', Value );
+                        $To.val(Value);
+                        var CustomerKey = $('#ToCustomer').data("customerkey").split(/, ?/);
+                        parent.Core.Agent.CustomerSearch.AddTicketCustomer( 'ToCustomer',Value , CustomerKey[Index] );
                 });
 
                 $.each($('#CcCustomer').val().split(/, ?/), function(Index, Value){
--- a/var/httpd/htdocs/js/Core.Agent.TicketAction.js
+++ b/var/httpd/htdocs/js/Core.Agent.TicketAction.js
@@ -103,7 +103,7 @@ Core.Agent.TicketAction = (function (TargetNS) {
             if(NewData.length){
                 NewData = NewData + ', ';
             }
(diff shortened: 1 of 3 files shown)


var/httpd/htdocs/js/Core.Agent.TicketAction.js
```diff
@@ -269,8 +278,9 @@ Core.Agent.TicketAction = (function (TargetNS) {
                 $Bcc = $('#BccCustomer', parent.document);
 
                 $.each($('#ToCustomer').val().split(/, ?/), function(Index, Value){
-                    $To.val(Value);
-                    parent.Core.Agent.CustomerSearch.AddTicketCustomer( 'ToCustomer', Value );
+                        $To.val(Value);
+                        var CustomerKey = $('#ToCustomer').data("customerkey").split(/, ?/);
+                        parent.Core.Agent.CustomerSearch.AddTicketCustomer( 'ToCustomer',Value , CustomerKey[Index] );
```
mrcbnsls: I think the indendation is wrong here, isn't it?

var/httpd/htdocs/js/Core.Agent.TicketAction.js
```diff
@@ -91,12 +91,21 @@ Core.Agent.TicketAction = (function (TargetNS) {
...
         }
         NewValue = NewValue + Core.Data.Get($Link.closest('tr'), 'Email');
         $Element.val(NewValue);
+        if (typeof $Element.attr("data-customerkey") !== 'undefined' ){
+            NewData = $Element.attr("data-customerkey");
+            if(NewData.length){
+                NewData = NewData + ', ';
+            }
+            NewData = NewData + $Link.data('customerkey');
+            $Element.attr("data-customerkey",NewData);
```
mrcbnsls: I think you could use $Element.data('customerkey') here and above, too, like you did in line 282. What do you think?

s7design: Hi @mrcbnsls,

I am sure that you are right, I did not know for this API function. Thanks for follow-up, I will check this and change on this way soon.

Regards
Zoran

var/httpd/htdocs/js/Core.Agent.TicketAction.js
s7design: Yes, you are absolutely right. It is my mistake :)

s7design:
Hi @mrcbnsls,
I fixed code as you suggested. I believe that it is better now. 
I tried to use Core.Data.Set()/Core.Data.Get() for the following lines 
$Element.attr("data-customerkey",NewData);
$Element.attr("data-customerkey")
but I think that it is not possible or maybe I do not know in this moment. Maybe your suggestion could be welcomed. The problem was for me in case there are more then one customer, I got only the first one with line Core.Data.Get($Element , "customerkey" );

Regards
Zoran

mrcbnsls:
@zottto, could I ask you to have a look at this PR, please?

s7design:
Hi @zottto ,

Thanks for your follow-up. You are right in your first comment. I will change this. Also I believe that you could be  right in second one, and I mentioned my problem in the previous comment 7 days ago. 
And my question was how I can  do this with Data API. I tested this and I was not be able to do this with API.
Function  Core.Data.Get() get only the first customer user if there are more selected customer user.

Regards
Zoran
S7

zottto:
Hi Zoran,

I had a deeper look into this and here is my feedback:

1. In the addressbook the input fields for To, Cc and Bcc have an active autocompletion itself, but clicking on an entry suggested by the autocompletion does not do anything. Because we have the search input field above, I think the autocompletion here is not wanted.

2. The problem with the customerkey is not, that the Data-API-call does not work. The problem is, that if you add more than one customer, the search for the second customer reloads the page and than the data-customerkey attribute is empty again. The second customerkey is not appended (because the first customerkey is lost). The Data-API as well as the access via attr() call will only return the latest customerkey. You have to fix that, to assure that the customerkeys will still be available after the reload of the iframe.

3. Probably this is a follow-up to 2.) If I add two customers and close the addressbook, the first customer is selected in the AgentTicketEmail screen, but the customer data of customer 2 is selected in the sidebar

4. You save the customerkeys in a comma-separated string. It would be better to use a JSON data structure to save this as data attribute. The key of the hash could be the customer email and the value the customer key. So you don't have to split up the data-customerkey, when the addressbook is closed.

Hope that helps,
Marc

s7design:
Hi @zottto,

Thanks for the detailed feedback, I will try to implement as you suggested. I believe that using JSON is very good advice.
I will do it soon, and you will be able to see is it well.

Regards
Zoran

s7design:
I am working on it, and I am planning to committ it in new PR soon.
I will do that like @zottto suggested.

Regards
Zoran

PR #404, opened 2015-01-02
Replaced hardcoded SQL limit with config parameter

chkal:
The SQL limit value is hardcoded to 50000 although there is a corresponding config parameter.


mrcbnsls:
Merged in 491cd00, thanks. For future contributions (which are always welcome!), I'd ask you to sign our CLA as described on http://otrs.github.io/. Thanks!

PR #410, opened 2015-01-12
Added frontend selenium test for AdminType

s7design:
Hi @mrcbnsls ,

Here is frontend test for AdminType. We added this one at the first. Please check it. If it is OK, we will add new ones for the others modules.
We are planning to add Selenium tests for modules that was completed with porting to OM, and in the following PR's we will add ported module and Selenium test together in one PR.

Also, there are some mistake in existing Selenium test, I have seen this when I analyzed Selenium testing. 
I am planning to fix this and make PR for this.
If you have any suggestion, please let me know.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Standard/AdminType.tt
+++ b/Kernel/Output/HTML/Standard/AdminType.tt
@@ -28,7 +28,7 @@
 
 [% RenderBlockStart("ActionAdd") %]
                     <li>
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Thanks, merged in 4e932f9. Looks pretty good to me!

PR #411, opened 2015-01-13
Added selenium frontend tests for AdminPriority and AdminRole

s7design:


scripts/test/Selenium/Agent/Admin/AdminRole.t
```diff
@@ -0,0 +1,166 @@
...
+        );
+
+        my $ScriptAlias = $ConfigObject->Get('ScriptAlias');
+
+        $Selenium->get("${ScriptAlias}index.pl?Action=AdminRole");
+
+        $Selenium->find_element( "table",             'css' );
+        $Selenium->find_element( "table thead tr th", 'css' );
+        $Selenium->find_element( "table tbody tr td", 'css' );
+
```
mrcbnsls: Same issue here as in the other UT I commented some minutes ago. This will not work on vanilla OTRS systems because there are no roles defined by default. Please consider this. Thanks!

s7design: Thanks for follow-up. I fixed this. I hope now there is much better UT for AdminRole.
Zoran

mrcbnsls:
Thanks for the fixes, works fine now. Merged in 865d1e4.

PR #413, opened 2015-01-15
Converting email addresses to html

schtritoff:
I had problems with recipient emails being converted to HTML. For example, email address in format like `"John Smith <john.smith@domain.tld>"` got converted to `"John Smith &lt;john.smith@domain.tld&gt;"` in table.field ARTICLE.A_TO using event notification. This PR solved that problem.

vco7txi

### DIFF
--- a/Kernel/System/Ticket/Event/NotificationEvent.pm
+++ b/Kernel/System/Ticket/Event/NotificationEvent.pm
@@ -643,6 +643,9 @@ sub _SendNotification {
         for my $Key ( sort keys %Recipient ) {
 
             next KEY if !$Recipient{$Key};
+            
+            # do not convert email addresses to HTML
+            next KEY if ($Key eq 'Email');
 
             $Recipient{$Key} = $HTMLUtilsObject->ToHTML(
                 String => $Recipient{$Key},


carlosfrodriguez:
Hi @schtritoff, @vupv,

In fact I think that the whole html conversion block is not needed, don’t you think?

**Remove**
```perl
    # convert values to html to get correct line breaks etc.
    if ( $Notification{Type} =~ m{text\/html} ) {

        KEY:
        for my $Key ( sort keys %Recipient ) {

            next KEY if !$Recipient{$Key};

            $Recipient{$Key} = $HTMLUtilsObject->ToHTML(
                String => $Recipient{$Key},
            );
        }
    }
```

((enjoy))
cr

schtritoff:
I'm not sure, didn't dig that much into code - I was in hunt for an email bug which led me to this solution.

carlosfrodriguez:
Hi @schtritoff,

Well the thing is that this block was introduced when we add the RichText functionality to this screen, but the recipient is not really related to the notification body or subject (where RichText changed apply), and it does not need any special conversion to HTML, as it was working before the RichText functionality was added, that's why I decided to remove the block, please take a look into the [bug 11038](http://bugs.otrs.org/show_bug.cgi?id=11038) for more information.

I will have to close this pull request, thank you very much for your contribution, I'm sorry that your patch was not integrated into the branch but it was a great help to find out the root cause, I will be looking forward for more contributions from your side.

((enjoy))
cr

PR #414, opened 2015-01-16
Fixed some mistakes in Selenium tests and add checks screens in Admin.t

s7design:
Hi @mrcbnsls ,

I saw while we tested existed Selenium test that there are some mistakes. I feel free to fix it. I fix following:

-Admin.t : add check for screens that is not exist
-AdminACL: change that default language is English. In my opinion it is better because this is case in the rest of system
- AdminQueue.t: There was a copy/paste mistake from AdminState.t
-AdminSelectBox.t: add Test User instead fixed root user. In this case there is unique test like in another Selenium tests, and also maybe in some OTRS test systems there is no root@localhost agent.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminACL.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminACL.t
@@ -48,7 +48,7 @@ JAVASCRIPT
 
         my $TestUserLogin = $Helper->TestUserCreate(
             Groups   => ['admin'],
-            Language => 'de',
+            Language => 'en',
         ) || die "Did not get test user";
 
         $Selenium->Login(
@@ -148,9 +148,9 @@ JAVASCRIPT
         # now we should not be able to add the same element again, an alert box should appear
         $Selenium->execute_script($CheckAlertJS);
         $Selenium->find_element( ".ItemAddLevel1 option[value='Properties']", 'css' )->click();
-        my $LanguageObject = Kernel::Language->new(
-            UserLanguage => 'de',
-        );
+
+        my $LanguageObject = $Kernel::OM->Get('Kernel::Language');
+
         $Self->Is(
             $Selenium->execute_script(
                 "return window.getLastAlert()"
--- a/scripts/test/Selenium/Agent/Admin/AdminACL.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminACL.t
@@ -46,9 +46,12 @@ $Selenium->RunTest(
 }());
 JAVASCRIPT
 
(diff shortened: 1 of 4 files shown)


scripts/test/Selenium/Agent/Admin/AdminACL.t
```diff
@@ -148,9 +148,9 @@ JAVASCRIPT
         # now we should not be able to add the same element again, an alert box should appear
         $Selenium->execute_script($CheckAlertJS);
         $Selenium->find_element( ".ItemAddLevel1 option[value='Properties']", 'css' )->click();
-        my $LanguageObject = Kernel::Language->new(
-            UserLanguage => 'de',
-        );
+
+        my $LanguageObject = $Kernel::OM->Get('Kernel::Language');
+
```
mrcbnsls: Why did you remove the 'de' stuff? I think this is to test if the message is being translated correctly.

s7design: Hi @mrcbnsls ,

I think that it is not case here. I believe that there was "de" because someone, who create this script, wanted to check in his native language.
In my opinion it is better if we have unique behavior and default language in general. I agree with you that we can set specific language if we want to test specific situation.

This is only my opinion. If you want we can leave as it was before.
Please, let me know what you mean about it.

Regards
Zoran

mrcbnsls: I tend to disagree. I still think the reason is to see if this message is being translated correctly, because the translation is done in JS at this place. The 'de' itself doesn't matter, it could be any other language. But for 'de' we know for sure we have the translation available. If you find a better solution for checking the correct translation, I'm open for it.

s7design: OK, I understand you. 
I will make another solution for it tomorrow, and there by default will be checked English and also  German translation will be checked (or any language).
We can discus tomorrow when I commit my solution.

Regards
Zoran

s7design:
Hi @mrcbnsls ,

I changed a little  AdminACL.t test. I added variable for language on the beginning of script. Maybe it could be good if someone want to test in more languages in the same time or only in his native language.
Maybe in AdminACL it was not hard to change language on two place in script, but it could be standard for the other scripts that we will make. If there will be more messages to check in some complicated module, we can use this way to test translating message of system.

In any case we can leave as it was before, but maybe this change is more configurable for any language.

Regards
Zoran

mrcbnsls:
Thanks, nice. Merged in 3ef6eba.

PR #416, opened 2015-01-16
Added selenium tests for Admin

s7design:
There are the following selenium tests:

AdminAutoResponse
AdminEmail
AdminEmailAccount
AdminNotification

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminEmail.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminEmail.t
@@ -27,7 +27,7 @@ $Selenium->RunTest(
     sub {
 
         my $Helper = Kernel::System::UnitTest::Helper->new(
-            RestoreSystemConfiguration => 0,
+            RestoreSystemConfiguration => 1,
         );
 
         my $Language = 'de';
@@ -116,13 +116,6 @@ $Selenium->RunTest(
         $Self->True(
             index( $Selenium->get_page_source(), $Expected ) > -1,
             "$RandomID admin notification was sent",
-        );
-
-        # restore RichText config
-        $Kernel::OM->Get('Kernel::System::SysConfig')->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'Frontend::RichText',
-            Value => 1
         );
 
         }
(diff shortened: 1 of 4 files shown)


scripts/test/Selenium/Agent/Admin/AdminEmail.t
```diff
@@ -0,0 +1,115 @@
...
+        );
+
+        my $ScriptAlias = $ConfigObject->Get('ScriptAlias');
+
+        $Selenium->get("${ScriptAlias}index.pl?Action=AdminEmail");
+
+        # check page
+        for my $ID (
+            qw(From UserIDs GroupIDs GroupPermissionRO GroupPermissionRW
+            NotifyCustomerUsers RoleIDs Subject RichText)
```
mrcbnsls: Checking RoleIDs is not working on vanilla OTRS instanced, because by default there are no roles in the system which means that the role selection is not being displayed in AdminEmail. The test should either consider if there are any roles or not check for the role selection.

scripts/test/Selenium/Agent/Admin/AdminEmail.t
```diff
@@ -0,0 +1,115 @@
...
+            my $Element = $Selenium->find_element( "#$ID", 'css' );
+            $Element->is_enabled();
+            $Element->is_displayed();
+        }
+
+        $Self->Is(
+            $Selenium->find_element( '#From', 'css' )->get_value(),
+            "admin\@example.com",
+            "#From stored value",
+        );
```
mrcbnsls: This check fails on my system. Does it work for you?

scripts/test/Selenium/Agent/Admin/AdminEmail.t
```diff
@@ -0,0 +1,115 @@
...
+        $Selenium->find_element( "#Subject",                         'css' )->submit();
+
+        # check if test admin notification is success
+        my $Expected = "Your message was sent to: $TestUserLogin\@localunittest.com";
+
+        $Self->True(
+            index( $Selenium->get_page_source(), $Expected ) > -1,
+            "$RandomID admin notification was sent",
+        );
+
```
mrcbnsls: This check also fails for me.

scripts/test/Selenium/Agent/Admin/AdminEmail.t
s7design: Thanks @mrcbnsls for follow-up. We will fix it soon. Your advice is very good, I did not think about it, because we had some test role in current test system.

Regards
Zoran

mrcbnsls:
Thx, merged in 28bd6b2. I only removed CustomerUserGroups from the elements to find on the page, because this is only visible if CustomerGroupSupport is enabled (you already know it :-D). But the UT is also fine without this check.

PR #418, opened 2015-01-22
Formated filter input fields

s7design:
Hi @mrcbnsls,

While we made some selenium test, we saw some filters that are not formatted like the the others ones.
I felt free to fix it. I did not open new bug for it. I believe that it could be useful that there are the same filters in all screens.
It is only my proposal, if you think that it is good you can merge it.

Regards
Zoran


mrcbnsls:
Thanks, merged in d384cec (rel-4_0) and 8ee5339 (master).

PR #419, opened 2015-01-22
[Rel-4_0 Bug 10984] Issues when add customer user by "Customer user" option

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10984
Hi @mrcbnsls,

I fixed here format style in AgentTicketEmail screen. There was a problem when it is added more customer user by "Customer user" options.
I also fixed problem with reload customer info data when radio button in MultipleCustomer. I am not sure that it is the best solution. I saw when I was solving this that if statement in  $('.CustomerTicketRadio').bind('change', function ()  is not necessary. It is works now, but I believe that your help could be useful with your experience in JS and OTRS generally.

Regards
Zoran


s7design:
It will be closed and solved on master branch

s7design:
.

PR #422, opened 2015-01-24
allow article dynamic fields in templates

reneeb:
When you reply in a ticket, you might want to use an article dynamic field in the response. Currently this is not possible...


mrcbnsls:
@reneeb, unfortunately this does not work for me. First of all, I get an error because `%Article` is not declared. I fixed this locally, but the replacement does not work, too. Could you please have a look?

mrcbnsls:
@reneeb, since there have not been any updates on this for a while, I'll close it for now. Please feel free to open another PR once the mentioned issue have been fixed. Thank you!

PR #424, opened 2015-01-27
[Rel-4_0 - BUG 10962 ]Invalid or wrong session on FAQ dasboard link

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10962

### DIFF
--- a/Kernel/Modules/AgentDashboardCommon.pm
+++ b/Kernel/Modules/AgentDashboardCommon.pm
@@ -827,7 +827,7 @@ sub _Element {
     }
 
     if ( $Config{CacheTTL} ) {
-        my $Content = $Self->{CacheObject}->Get(
+        $Content = $Self->{CacheObject}->Get(
             Type => 'Dashboard',
             Key  => $CacheKey,
         );
--- a/Kernel/Modules/AgentDashboardCommon.pm
+++ b/Kernel/Modules/AgentDashboardCommon.pm
@@ -832,9 +832,11 @@ sub _Element {
             Key  => $CacheKey,
         );
 
-          if ( $PreviewContent =~ /$Self->{SessionName}.'='.$Self->{SessionID}/i){
+        if ( $Self->{'HttpType'} eq 'https' && $PreviewContent =~ /$Self->{SessionName}/ ) {
+            if ( $PreviewContent =~ /$Self->{SessionName}.'='.$Self->{SessionID}/i ) {
                 $Content = $PreviewContent;
-          }
+            }
+        }
     }
 
     # execute backends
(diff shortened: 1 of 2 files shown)


carlosfrodriguez:
Hello @s7design,

For now I believe that we can implement something less aggressive as I'm not sure the side effects on this. It also looks that it is not exactly a problem with https, if you disable the storage of session in cookies it also happens with http.

My proposal is to replicate "somehow" the DashboardTicketGeneric, in DashboardFAQ, change the CacheTTL setting to CacheTTLLocal, then the cache from AgentDashboardCommon will not apply, manage the cache local in DashboardFAQ.

Now FAQShowLatestNewsBox() does not return something cacheable so we need to modify it:

Option1 Split the first part including the FAQSearch, into a separated function that returns a reference to the @ItemIDs, then just call  FAQShowLatestNewsBox() with the \@ItemIDs as param.

Option2 Make FAQShowLatestNewsBox() able to receive a \@ItemIDs in such case it does not need to calculate them again, then modify the output to send back the \@ItemIDs

What do you think? maybe and Option3?

((enjoy))
cr

s7design:
Hi @carlosfrodriguez,

Thanks for follow-up. I believe that your suggestion would be nice for better fix. To be honest I don't know how to implement it in this moment. I am trying to do it, and I am sure that we will be able to fix it, your help will be welcomed. We can discus about it to find the best solution.

Regards
Zoran

carlosfrodriguez:
Hi @s7design 

For sure let me know how we can get in touch.

((enjoy))
cr

s7design:
Hi @carlosfrodriguez,

We can get in touch by Skype.  My nick at office is zilibasic.s7desgin.
And of course by email, I think that you have my email in office. 

Regards 
Zoran

carlosfrodriguez:
Hi @s7design,

I have been re-thinking about this, and I believe that the only problem is the catching of the FAQ dashboard widget, but the information inside the widget is also used in other places like in the FAQ Explorer on all its interfaces and it is not cached there, so I think that we can also remove the cache of the dashboard widget too until a much better solution can be implemented. (if is really needed).

What do you think about it?

((enjoy))
cr

s7dsucurovic:
Hi @carlosfrodriguez,

Yes I agree with you it would be good solution in this moment.
I will provide a commit soon.

Regards
Zoran

s7design:
Hi @carlosfrodriguez,

Yes I agree with you it would be good solution in this moment.
I will provide a commit soon.

Regards
Zoran

s7design:
Hi @carlosfrodriguez,

I fixed this. It is my proposal. If I well understand you, we don't want use cache data for FAQ widget, because FAQ explore doesn't use it as well.
I fixed it so that I don't set cache data for these widgets. Please let me know if you have any suggestions.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

Normally we don't use any package specific code in the framework, I my opinion the only change that is needed is in the FAQ module to delete the CacheTTL from the Widget in the SysConfig xml file.

If you agree please close this pull request and open another against FAQ:master

((enjoy))
cr

s7design:
Hi @carlosfrodriguez,

I agree with you it could be good solution. I will make another PR for this soon.

Regards
Zoran

PR #425, opened 2015-01-28
[Rel-4_0 - Bug 10988] Queues will not be visible in CustomerTicketMessage

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10988

### DIFF
--- a/scripts/test/Selenium/Customer/UpdateQueueGroup.t
+++ b/scripts/test/Selenium/Customer/UpdateQueueGroup.t
@@ -1,0 +1,171 @@
+# --
+# UpdateQueueGroup.t - UpdateQueueGroup tests update 'To' in CustomerTicketMessage on Add/Update Group
+# Copyright (C) 2001-2015 OTRS AG, http://otrs.com/
+# --
+# This software comes with ABSOLUTELY NO WARRANTY. For details, see
+# the enclosed file COPYING for license information (AGPL). If you
+# did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
+# --
+
+use strict;
+use warnings;
+use utf8;
+
+use vars (qw($Self));
+
+# get selenium object
+$Kernel::OM->ObjectParamAdd(
+	'Kernel::System::UnitTest::Selenium' => {
+		Verbose => 1,
+	}
+);
+my $Selenium = $Kernel::OM->Get('Kernel::System::UnitTest::Selenium');
+
+$Selenium->RunTest(
+    sub {
+
+        # get helper object
+		$Kernel::OM->ObjectParamAdd(
+			'Kernel::System::UnitTest::Helper' => {
+				RestoreSystemConfiguration => 0,
+			}
+		);
+
+		my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
+
+		# create and login test customer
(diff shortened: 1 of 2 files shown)


s7design:
Hi @mgruner ,

Hereby is update fix with adding clear cache when Group is updated. Furthermore there is a unit test which work both before and after fix, so you can test different scenarios right off the bat. Soon there will be PR for master branch as well.

Regards,
S7

scripts/test/Selenium/Customer/UpdateQueueGroup.t
```diff
@@ -0,0 +1,171 @@
...
+        $Selenium->refresh();
+
+        # check if test queue is available to select
+        my $Success;
+        eval {
+            $Success = index( $Selenium->get_page_source(), $QueueName ),
+        };
+
+        if ( $Success > -1 ) {
+            print "\nPatch is success $QueueName is available to select with new group $GroupName\n\n";
```
mgruner: Usually print() should not be used in UTs. Use $Self->True/False instead.

scripts/test/Selenium/Customer/UpdateQueueGroup.t
```diff
@@ -0,0 +1,171 @@
...
+        }
+        else {
+            print "\n$Queue{Name} is no longer available to select with new group $GroupName\n\n";
+
+            # clear cache
+            print "\nClearing cache\n\n";
+            $CacheObject->CleanUp(
```
mgruner: I don't understand why you need to clear the cache here, what is the purpose of this?

scripts/test/Selenium/Customer/UpdateQueueGroup.t
```diff
@@ -0,0 +1,171 @@
...
+
+            # refresh page
+            $Selenium->refresh();
+
+            # verify that test queue is now available for test group after clearing cache
+            $Self->True(
+                $Selenium->find_element( "#Dest option[value='$QueueID||$QueueName']", 'css' ),
+                "$Queue{Name} is available to select after clearing cache"
+            );
+        }
```
mgruner: I think this whole section could be replaced by a simple execute_script call. Just look with jquery for the option element you are looking for, and use .length() to check if it exists. No need to use if/else and eval. Same below. The whole test can be simplified a lot.

scripts/test/Selenium/Customer/UpdateQueueGroup.t
s7design: Hi @mgruner ,

This clearing of cache is here only for testing behavior that is motioned in the bug report.
With this you can test behavior before and after fixing. I can agree that it is not necessary if you want to merge our fix, this was only for testing in this PR. Maybe I did not understand what a unit test you wanted. I will change this and it will be ready for merging.

Regards
Zoran

scripts/test/Selenium/Customer/UpdateQueueGroup.t
s7design: Yes I agree with you. Now we use execute_script call in such situation. We have not know for this before.

s7design:
Hi @mgruner 

I updated selenium test, I hope it is well now.

Regards
Zoran

mgruner:
Merged, thank you!

PR #426, opened 2015-01-28
[Master - Bug 10988] Queues will not be visible in CustomerTicketMessage

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10988

Hi @mrcbnsls ,

There is PR for master, I saw while I tested this bug that there is difference between rel-4_0 and master branch of OTRS system.
It is small fix, but I think that it will be easier for you while merging if there are two separate PR.

Regards
Zoran

### DIFF
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -293,6 +293,9 @@ sub GroupUpdate {
         Type => 'Group',
         Key  => 'GroupList::1',
     );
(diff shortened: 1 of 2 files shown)


mgruner:
I have some comments about the selenium test in the other PR. I'll close this one here for now. Maybe a separate one for master is not really needed if the other one can be cherry-picked? Otherwise you can reopen it later.

PR #427, opened 2015-01-28
Master port and selenium test for AdminCustomerUserGroup

s7design:


mrcbnsls:
@s7design, this PR seems wrong to me. You put it against rel-4_0, but the branch name indicates that it is for master. Please check and open another one. Thanks!

PR #428, opened 2015-01-28
Added port and selenium test for AdminCustomerUserGroup

s7design:
Hi @mrcbnsls.

There is new PR for AdminCustomerUserGroup. You are right, PR that you closed, was wrong.
I believe that it is good now.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminCustomerUserGroup.pm
+++ b/Kernel/Modules/AdminCustomerUserGroup.pm
@@ -12,8 +12,7 @@ package Kernel::Modules::AdminCustomerUserGroup;
 use strict;
 use warnings;
 
-use Kernel::System::CustomerUser;
-use Kernel::System::CustomerGroup;
+our $ObjectManagerDisabled = 1;
 
 sub new {
     my ( $Type, %Param ) = @_;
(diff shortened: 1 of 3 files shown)


Kernel/Output/HTML/Standard/AdminCustomerUserGroup.tt
```diff
@@ -253,7 +253,7 @@
             </div>
             <div class="Content LayoutGrid ColumnsWithSpacing">
                 <div class="Size1of2">
-                    <ul class="Tablelike" id="Always Groups">
+                    <ul class="Tablelike" id="AlwaysGroups">
```
s7design: Hi @mrcbnsls, 

I think that there was mistake, because of that we felt free to fix it. I think that space in ID is typo.
But please follow-up it.

Regards
Zoran

Kernel/Modules/AdminCustomerUserGroup.pm
```diff
@@ -22,36 +21,32 @@ sub new {
...
 sub Run {
     my ( $Self, %Param ) = @_;
 
+    my $LayoutObject        = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
+    my $ConfigObject        = $Kernel::OM->Get('Kernel::Config');
+    my $ParamObject         = $Kernel::OM->Get('Kernel::System::Web::Request');
+    my $CustomerUserObject  = $Kernel::OM->Get('Kernel::System::CustomerUser');
+    my $CustomerGroupObject = $Kernel::OM->Get('Kernel::System::CustomerGroup');
+    my $GroupObject         = $Kernel::OM->Get('Kernel::System::Group');
+
```
mrcbnsls: Same issue here as in my other comment. The 4 latter objects can be pulled after line#51. So if CustomerGroupSupport is not active at all, these objects dont need to get pulled. I fixed this in my merge commit.

scripts/test/Selenium/Agent/Admin/AdminCustomerUserGroup.t
```diff
@@ -0,0 +1,195 @@
...
+
+my $Selenium = Kernel::System::UnitTest::Selenium->new(
+    Verbose => 1,
+);
+
+$Selenium->RunTest(
+    sub {
+
+        my $Helper = Kernel::System::UnitTest::Helper->new(
+            RestoreSystemConfiguration => 0,
```
mrcbnsls: In this case, RestoreSystemConfiguration needs to be 1, because otherwise your changes (enabling/disabling CustomerGroupSupport) will not get undone after finishing the test. That means that CustomerGroupSupport will be disabled after running the UT although you had enabled it before. I also fixed this in my merge commit.

mrcbnsls:
Merged in 99c3515. Thank you! Please see my comments and consider these in your future work!

PR #429, opened 2015-01-28
TicketSearch: Implement search by article flag

moritz:
... including negation.
For example this could be used to search for tickets containing important
articles, or those without.

Tests included.

### DIFF
--- a/Kernel/System/Ticket/TicketSearch.pm
+++ b/Kernel/System/Ticket/TicketSearch.pm
@@ -132,6 +132,9 @@ To find tickets in your system.
             SmallerThanEquals => '2002-02-02 02:02:02',
         }
 
+        # User ID for searching tickets by ticket flags (defaults to UserID)
+        TicketFlagUserID => 1,
+
         # search for ticket flags
         TicketFlag => {
             Seen => 1,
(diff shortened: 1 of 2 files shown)


Kernel/System/Ticket/TicketSearch.pm
```diff
@@ -1095,6 +1142,27 @@ sub TicketSearch {
...
     }
 
+    if ( $Param{NotArticleFlag} ) {
+        my $ArticleFlagUserID = $Param{ArticleFlagUserID} // $Param{UserID};
+        my $Index = 1;
+        for my $Key ( sort keys %{ $Param{NotArticleFlag} } ) {
+            my $Value = $Param{NotArticleFlag}->{$Key};
+            return if !defined $Value;
+
+            $SQLExt .= " AND NOT EXISTS (SELECT na$Index.id "
```
mgruner: Thanks for this nice suggestion, @moritz!
Using sub selects would be a performance killer, though, I think this is no option. It would fire a SELECT per ticket. Perhaps it could be done like NotTicketFlag with a LEFT JOIN instead?

moritz: @mgruner I've tried it with a LEFT JOIN, but that doesn't work because it goes through two tables, and in the second table you have lost too much information (basically the distinction between the WHERE and the ON clauses). Though someone smarter than me might find a better solution.

Would you consider merging the ArticleFlag search while leaving out the NotArticleFlag?

mgruner:
Yes, I think that would be good.

moritz:
@mgruner I've now update the pull request to include only the positive search for article flags.

mgruner:
Merged in 28b4d6f! Thanks a lot!

PR #430, opened 2015-01-29
Selenium test package manager

s7design:
Hi @mrcbnsls,

There is a selenium test and porting to OM for PackageManager.
I added TestPackage.opm file in test/sample folder. I used it for Selenium testing. I hope that it is OK.

I can use String value in format like xml file, or temporarily local file in AdminPackageMenager.t, but with added test package module I am able to test installing package from file that is selected from file system.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminPackageManager.pm
+++ b/Kernel/Modules/AdminPackageManager.pm
@@ -2031,7 +2031,6 @@ sub _GetFeatureAddonData {
     my ( $Self, %Param ) = @_;
 
     my $Language    = $Kernel::OM->Get('Kernel::Output::HTML::Layout')->{UserLanguage};
-    my $CacheObject = $Kernel::OM->Get('Kernel::System::Cache');
 
     # cleanup main language for languages like es_MX (es in this case)
     $Language = substr $Language, 0, 2;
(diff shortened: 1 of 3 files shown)


Kernel/Modules/AdminPackageManager.pm
```diff
@@ -29,28 +24,20 @@ sub new {
...
     my ( $Self, %Param ) = @_;
 
+    my $PackageObject     = $Kernel::OM->Get('Kernel::System::Package');
+    my $LayoutObject      = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
+    my $ParamObject       = $Kernel::OM->Get('Kernel::System::Web::Request');
+    my $DBObject          = $Kernel::OM->Get('Kernel::System::DB');
+    my $MainObject        = $Kernel::OM->Get('Kernel::System::Main');
+    my $ConfigObject      = $Kernel::OM->Get('Kernel::Config');
+    my $UploadCacheObject = $Kernel::OM->Get('Kernel::System::Web::UploadCache');
+
```
mrcbnsls: Objects should be pulled from the OM just before they are actually needed. There are returns in the code (e.g. line#62 or line#77) before some of these objects you created here are used. Please go through the code again and create objects later if possible.

Example: The ParamObject ist used first in line#84, so it should be pulled from OM in line#78 (before if ( $Self->{Subaction} eq 'ViewDiff' )).

Please let me know if you have more questions about this.

mrcbnsls:
Hi @s7design, please see my comment on the code and update the PR accordingly. Apart from that, the UT looks good and works fine for me! I will merge everything together once you fixed the other things.

mrcbnsls:
Merged in 9bdfe80 with only small changes. Thanks!

PR #432, opened 2015-01-29
Added ports and Selenium tests

s7design:
Added ports for some modules and Selenium tests.
Selenium tests for AdminSelectbox and AdminSession are already included in master.
In Ticket.xml we changed regex for PostMasterFilter so PostmasterHeaderFieldCount cannot be set as 0.


mrcbnsls:
Merged in cbee08c7, thank you!

PR #433, opened 2015-01-30
fixed typo for "exept=>except" and improved datepicker for validation future date

s7design:
Hi @mrcbnsls,

I again saw something wrong while I tested code by Selenium and ported to OM. In my opinion there was type with "exept", it should be "except".
And when we tested AdminSystemMaintenance screen, we saw that it could be a problem with datepicker. There was only current year in datepicker. If someone want to create system SystemMaintenance on the end of year, it will not be possible.Also we add validation for future year. 
Maybe it is not so important, because of that it is only my proposal for fix.

What do you think about it? If you mean that it have a sense, I can make another PR for rel-4_0.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminSystemMaintenance.pm
+++ b/Kernel/Modules/AdminSystemMaintenance.pm
@@ -243,7 +243,7 @@ sub Run {
             # add notification
             push @NotifyData, {
                 Priority => 'Notice',
-                Info     => 'All sessions has been killed, exept the current one!',
+                Info     => 'All sessions has been killed, except the current one!',
             };
 
             # set class for expanding sessions widget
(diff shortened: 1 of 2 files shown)


s7design:
Hi @mrcbnsls,
I am not sure if I did need change language files.
Regards
Zoran

mrcbnsls:
Hi @s7design, thanks. No, the language files don't need to be touched, they are generated using a script (which can be done by one of us). So please remove all Kernel/Language* files and i18n/* files from the PR.

s7design:
Hi @mrcbnsls,
Thanks for follow-up, I will do that soon.
I will leave only SysMaintance files, and I think that it is better that you call this script for generating language files.
Regards
Zoran

s7design:
I corrected this, I believe that it is good now.

mrcbnsls:
Merged, thx. I though have dumped the change regarding the date validation, because that would hinder users from changing already running maintenance periods. They would need to adapt the date everytime (which don't want to).

PR #436, opened 2015-01-31
Create rhel7-otrs-systemd.spec

hisa0211:
Created as RPM spec file for using systemd.

This file depands on some other systemd script files (Also I created).
When reviewing this, please evaluate them (list below) at the same time.
- scripts/redhat-systemd-otrs.dist
- scripts/redhat-systemd-otrs.mariadb
- scripts/redhat-systemd-otrs.oracle
- scripts/redhat-systemd-otrs.postgresql

### DIFF
--- a/scripts/rhel7-otrs-systemd.spec
+++ b/scripts/rhel7-otrs-systemd.spec
@@ -72,6 +72,8 @@ install -m 644 scripts/apache2-httpd.include.conf $RPM_BUILD_ROOT/etc/httpd/conf
 if test -e /opt/otrs/RELEASE; then
     cat /opt/otrs/RELEASE|grep VERSION|sed 's/VERSION = //'|sed 's/ /-/g' > /tmp/otrs-old.tmp
 fi
+
+%post
 # useradd
 export OTRSUSER=otrs
 echo -n "Check OTRS user ... "
@@ -82,8 +84,6 @@ if id $OTRSUSER >/dev/null 2>&1; then
 else
     useradd $OTRSUSER -d /opt/otrs/ -s /bin/false -g apache -c 'OTRS System User' && echo "$OTRSUSER added."
 fi
-
-%post
 # run OTRS permission setting.
 /opt/otrs/bin/otrs.SetPermissions.pl --web-group=apache
 # if it's a major-update backup old version templates (maybe not compatible!)
--- a/scripts/rhel7-otrs-systemd.spec
+++ b/scripts/rhel7-otrs-systemd.spec
@@ -1,0 +1,156 @@
+# --
+# RPM spec file for RHEL7 of the OTRS package with systemd
+# Copyright (C) 2001-2012 OTRS AG, http://otrs.org/
+# --
+# This software comes with ABSOLUTELY NO WARRANTY. For details, see
+# the enclosed file COPYING for license information (AGPL). If you
+# did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
+# --
+#
+# please file bugfixes or comments on http://bugs.otrs.org
+#
+# --
+Summary:      OTRS Help Desk.
+Name:         otrs
+Version:      0.0
+License:    GNU AFFERO GENERAL PUBLIC LICENSE Version 3, 19 November 2007
+#Copyright:    GNU AFFERO GENERAL PUBLIC LICENSE Version 3, 19 November 2007
+Group:        Applications/Mail
+Provides:     otrs
+# mod_perl removed for now, not available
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @hisa0211, thank you for your contribution. Could you please give some more information on your PRs (this one and the connected ones), especially about which things/issues are being improved through the changes. Also, I need to ask you to sign our CLA (as described on http://otrs.github.io/) before we are able to integrate your changes into OTRS. Thank you!

hisa0211:
Hi @mrcbnsls, thank you for your comment.
Yes, I'll show you some information about my work for #436-#440.
It may be redundant because all are collectively writing, please ask if there is something what you cannot understand.


[#436] "scripts/rhel7-otrs-systemd.spec"
A spec file for RHEL/CentOS7. As you know, latest distribution of RHEL/CentOS uses systemd instead of initd. So, it's better to rush the transition to systemd.

This file creates RPM with following rules.
Most of "/opt/otrs" and files/directories under thatare kept previous ones.
But some changes exists. Following is it.

- Excules some redundant files under scripts directory. 
 - Configurations named "rechat-*", but renamed as "otrs.service(.*)".
 - Configurations named "suse-*.
 - RPM rule file named "*.spec".

- "/opt/otrs/scripts/otrs.service.{dist,mariadb,oracle,postgresql}"
  Includes new systemd scripts for replacing.

- "/usr/lib/systemd/system/otrs.service"
  This is new systemd script file. For details, please read descriptions following files.

- "/etc/httpd/conf.d/zzz_otrs.conf"
- "/etc/sysconfig/otrs"
  Files from previous one.

- In accordance with using systemd commands, I rewrite also instructions that are displayed after installation.

- And when re-installing/updating, RPM with this spec won't replace "/etc/sysconfig/otrs" and "/usr/lib/systemd/system/otrs.service", if exists.


[#437/#438/#439/#440] Common description about "scripts/redhat-systemd-otrs.*".
This is systemd script file for start/stop OTRS scheduler.
Former scripts named "scripts/redhat-rcotrs" has many many operations other than start/stop OTRS sceduler. But, systemd not recommends.
So, New files focus to only for start/stop OTRS scheduler and start Cron/Web/DB service before starting.
This file nominates "/etc/sysconfig/otrs" as a configuration file ("EnvironmentFile" parameter is this), but these file not use it. Just took over the previous definition. If you judged unnecessary, and may be removed it.


[#437] "scripts/redhat-systemd-otrs.dist"
When installing OTRS via RPM with "rhel-otrs-systemd.spec" I wrote, this file is installed as a standard "/usr/lib/systemd/system/otrs.service" and "/opt/otrs/scripts/otrs.service.dist". The latter one is a backup to keep default setting. 

It doesn't operate any RDBMS because we don't know what/where RDBMS uses with OTRS has installed. But this file may be replaced with other "otrs.service" files if using some RDBMS on current node.
But it doesn't mean this file is useless. If user uses OTRS with external DBs, it's a good convenience more of this file.


[#438] "scripts/redhat-systemd-otrs.mariadb"
[#439] "scripts/redhat-systemd-otrs.oracle"
[#440] "scripts/redhat-systemd-otrs.postgresql"
These are systemd script files as same as #437. But it operates DBs (named respectively) before starting otrs.service via systemd.
User can replace it as "/usr/lib/systemd/system/otrs.service".


That's all. Thanks for reading.

hisa0211:
I caught that my file has an error.
Modified file is "rhel7-otrs-systemd.spec" (#436) only.

Please note that.

mgruner:
Hello @hisa0211, thank you very much for your CLA and for this proposal. I started working on this and will provide feedback when it goes into the master branch (this will be part of OTRS 5, not 4). It is likely that I will not include the database specific files as I believe that the installation should work as automatically as possible. That depends too much on the setup, the DB could also live remotely etc.

mgruner:
Hello @hisa0211, and thank you very much for your submission, again. We decided to actually drop the OTRS service from the RPM files. It has never been a real part of the upgrading instructions, for example.
With OTRS 5, there is now only one cron job managing the otrs Daemon. You might want to manage this daemon with a systemd unit and not via cron job any more. I tend to not want to do that in general, though, as the upgrading procedure would be different in this case compared to all other systems.
Please feel free to continue the discussion if you have other items to consider.
Thank you very much and best regards, mg

hisa0211:
My first trigger for make this, the existing init-script (NOT cron). Because that operates many other daemons like mail and DBs. I only wanted to solve that.
If another solution is better one, I don't mind.
Thanks for reporting the result.

PR #441, opened 2015-02-02
Added port and selenium tests

s7design:


mrcbnsls:
Thx, merged in 5a40acc. I removed TicketWatcher from the list of links to be found in the UT, because its only found for the term 'admin' if Ticket::WatcherGroup is enabled, which is not the case by default.

PR #400, opened 2014-12-24
Ported to OM 23_12

s7design:

PR #442, opened 2015-02-04
Master port 03 02

s7design:


Kernel/Modules/AdminRegistration.pm
```diff
@@ -27,30 +21,24 @@ sub new {
...
 
+    my $LayoutObject       = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
+    my $ParamObject        = $Kernel::OM->Get('Kernel::System::Web::Request');
+    my $RegistrationObject = $Kernel::OM->Get('Kernel::System::Registration');
+    my $ConfigObject       = $Kernel::OM->Get('Kernel::Config');
+    my $OTRSBusinessObject = $Kernel::OM->Get('Kernel::System::OTRSBusiness');
+    my $EnvironmentObject  = $Kernel::OM->Get('Kernel::System::Environment');
+    my $RegistrationState  = $Kernel::OM->Get('Kernel::System::SystemData')->SystemDataGet(
+        Key => 'Registration::State',
+    ) || '';
```
mrcbnsls: @s7design, same issue here as in other PRs. Objects should be pulled once they are needed. Please go through your PRs and check for this topic. Thanks!

mrcbnsls:
@s7design, AdminSMIME does not work for me. Both using the module directly and the UT fail. This seems to be caused by the fact that you create an instance of the crypt object (wich seems to be right) but don't 'use' it on top of the file. Please check. After I fixed this locally, the UT still fails because the delete link can't be found. You used the filename to find the right element, but this filename seems to be different everytime. Please verify and fix, thanks!

Kernel/Modules/AdminSMIME.pm
```diff
@@ -22,50 +21,46 @@ sub new {
...
-        qw(ParamObject DBObject LayoutObject ConfigObject LogObject MainObject EncodeObject)
-        )
-    {
-        if ( !$Self->{$Needed} ) {
-            $Self->{LayoutObject}->FatalError( Message => "Got no $Needed!" );
-        }
-    }
-
-    $Self->{CryptObject} = Kernel::System::Crypt->new( %Param, CryptType => 'SMIME' );
-    $Self->{CustomerUserObject} = Kernel::System::CustomerUser->new(%Param);
```
mrcbnsls: Btw, as suggested by Moritz, please use the same syntax here as mentioned here: https://github.com/OTRS/otrs/pull/400#discussion-diff-22269249

s7design: Thanks @mrcbnsls ,
I look this right now. We tried to use this as Moritz suggested, but we had some problem to implement here.
I will test it a little bit more, and then I am planning to find better solution.

Regards
Zoran

s7design:
Hi @mrcbnsls ,

I am testing this, but it seems a little tricky. 
What was a problem while you test it. I have in one test case Internal Server Error, but I don't know in which one.
After that I can test in Selenium and manually, because of that I said that it is a little tricky. If you see what is a problem please tell me. In any case, I am doing on it and I believe it will be done soon.

Regards
Zoran

s7design:
When we started to do on porting for this module or on AdminPGP in another PR ,there was a problem that is similar to this one.We got Internal Server Error if SMIME/ PGP is not configured. But if this config item is set all was ok. We saw that we could not use  $Kernel::OM->ObjectParamAdd.

I am not absolutely  sure that it was a problem, but I am sure that there is a problem :)
I believe that we will solve this with your help :)

Zoran

mrcbnsls:
@mgruner, lets talk about this one once you're back.

s7design:
I will close this PR and make new one. There I will do port to OM with new Crypt:SMIME object.
I tried to do rebase  this branch with new master, but it will be easier for me if I do in new PR.

Regards
Zoran

s7design:
I reopen this PR, I think maybe we can leave here porting to OM for 
Kernel/Modules/AdminRegistration.pm

I new PR will be done porting to OM for AdminSMIME.pm and Selenium test for it.

Regards
Zoran

mgruner:
Merged, thanks!

PR #444, opened 2015-02-04
[Master Bug 11001]   Confusing screen output when trying to log in with an invalid CustomerUser

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11001

Here is fixed problem with login messages for agent and customer.
We have fixed it in master, we will be able to make new PR for rel-4_0 if it would be necessary.

Regards
Zoran


mrcbnsls:
@s7design, please see my comment on the bug entry. Thanks!

PR #446, opened 2015-02-05
[Rel-4_0 - Bug 10970] byte sequence errors on notifications

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10970

There was the issue only in OTRS with PostgreSQL. I tested it after fixing and now it works fine in PostgreSQL.
In mysql it works also without fix, and now as well. 

Regards
Zoran

### DIFF
--- a/Kernel/System/NotificationEvent.pm
+++ b/Kernel/System/NotificationEvent.pm
@@ -16,6 +16,7 @@ our @ObjectDependencies = (
     'Kernel::System::DB',
     'Kernel::System::Log',
     'Kernel::System::Valid',
+    'Kernel::System::Encode',
 );
 
 =head1 NAME
@@ -118,8 +119,33 @@ sub NotificationGet {
         );
     }
 
+    # get encode object
+    my $EncodeObject = $Kernel::OM->Get('Kernel::System::Encode');
+
     my %Data;
     while ( my @Row = $DBObject->FetchrowArray() ) {
+
+        # convert body
+        $Row[3] = $EncodeObject->Convert(
+            Text  => $Row[3],
+            From  => $Row[5],
+            To    => 'utf-8',
+            Force => 1,
+        );
+
+        # convert subject
+        $Row[2] = $EncodeObject->Convert(
+            Text  => $Row[2],
+            From  => $Row[5],
+            To    => 'utf-8',
+            Force => 1,
+        );
+
+        # set new charset
+        $Row[5] = 'utf-8';
+
+        # fix some bad stuff from some browsers (Opera)!
+        $Row[3] =~ s/(\n\r|\r\r\n|\r\n|\r)/\n/g;
         $Data{ID}         = $Row[0];
         $Data{Name}       = $Row[1];
         $Data{Subject}    = $Row[2];
--- a/scripts/test/Selenium/Agent/Admin/AdminNotificationEventEncode.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminNotificationEventEncode.t
@@ -1,0 +1,146 @@
+# --
+# AdminNotificationEventEncode.t - frontend tests for AdminNotificationEventEncode
+# Copyright (C) 2001-2015 OTRS AG, http://otrs.com/
+# --
+# This software comes with ABSOLUTELY NO WARRANTY. For details, see
+# the enclosed file COPYING for license information (AGPL). If you
+# did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
+# --
+
+use strict;
+use warnings;
+use utf8;
+
+use vars (qw($Self));
+
+our $ObjectManagerDisabled = 1;
+
+# get needed objects
+my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
+
+# get selenium object
+$Kernel::OM->ObjectParamAdd(
+    'Kernel::System::UnitTest::Selenium' => {
+        Verbose => 1,
+        }
+);
+my $Selenium = $Kernel::OM->Get('Kernel::System::UnitTest::Selenium');
+
+$Selenium->RunTest(
+    sub {
+
+        # get helper object
+        $Kernel::OM->ObjectParamAdd(
+            'Kernel::System::UnitTest::Helper' => {
+                RestoreSystemConfiguration => 0,
+                }
+        );
+        my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
+
+        # do not check RichText
+        $Kernel::OM->Get('Kernel::System::SysConfig')->ConfigItemUpdate(
+            Valid => 1,
+            Key   => 'Frontend::RichText',
+            Value => 1
+        );
+
+        my $TestUserLogin = $Helper->TestUserCreate(
+            Groups => ['admin'],
+        ) || die "Did not get test user";
+
+        $Selenium->Login(
+            Type     => 'Agent',
+            User     => $TestUserLogin,
+            Password => $TestUserLogin,
+        );
+
+        # get notification that ends with &nbsp;
+        my $NotifEventRandomID = "notifevent" . $Helper->GetRandomID();
+        my $NotifEventText     = "Selenium NotificationEvent test";
+
+        # Add NotificationEvent
+        my $NotificationID = $Kernel::OM->Get('Kernel::System::NotificationEvent')->NotificationAdd(
+            Name    => $NotifEventRandomID,
+            Subject => $NotifEventText,
+            Body    => "Test notification&nbsp;",
+            Type    => 'text/html',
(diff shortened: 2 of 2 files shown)


carlosfrodriguez:
Hi @s7design,

I'm having problems to recreate the issue, the reporter have PostgreSQL 8.4.x while I have PostgreSQL 9.3.x

Can you send me the detailed steps to recreate it?

((enjoy))
cr

s7design:
Hi,

We communicated with @carlosfrodriguez by mails about this, and now I saw that maybe this conversation could be interesting for someone else here, and because of that I add this comment. 

  I worked on this bug that is connected with PostgreSQL. My version of PostgreSQL is 9.1.11.
    This is steps how I reproduce bug:
    make notification event
                e.g. - events:TicketQueueUpdate
                Ticket filter: all queues
                Recipient: Customer
                Notification Subject: some subject
                and the most important for reproduction this bug is text of notification: It should be something like "Test notification &nbsp;" in source form or in text form "Test notification 2*<whitespace>"
                (there is attachment, you can see there)
    test this notification, in my case change queue for a ticket
    check syslog. There is error massage. Something like this

        ERROR: invalid byte sequence for encoding "UTF8": 0xa0, SQL: 'INSERT INTO article (ticket_id, article_type_id, article_sender_type_id, a_from, 
        a_reply_to, a_to, a_cc, a_subject, a_message_id, a_message_id_md5, a_in_reply_to, a_references, a_body, a_content_type, content_path, valid_id, 
        incoming_time, create_time, create_by, change_time, change_by) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,...

    I tested this in master and 4_0, and the bug occurs in the both versions. I hope that my explanation is helpful for you. If you have any question feel free to ask.

s7design:
And as I said there is a problem in my system with PostgreSQL  9.1.11.
    I don't know how I can help you i reproducing the bug, maybe there is not a bug with PostgreSQL is 9.3.x or PostgreSQL is 9.4.x. I am planning to test it with newer version of PostgreSQL.

    But in my opinion there is a problem in any case, because another notifications is solved on the similar way.
    For example:

     Notification.pm function NotificationGet() line 92
    AutoResponse.pm function AutoResponseGet line  138

And because of that I fixed it on this way. I believe that it is necessary in this module. Also I could see that I fix also Notification.pm in this PR. This was definitely a bug. I saw this while I was working on this bug, and I thought that it is good if I fix it here.


            $DBObject->Prepare(

                SQL => 'SELECT id, notification_type, notification_charset, '

                    . ' notification_language, subject, text, content_type '

                    . ' FROM notifications WHERE '

                    . ' notification_type = ? AND notification_language = ?',

                Bind  => [ \$Type, \$Language, ],

                Limit => 1,

            );

            # get encode object

            my $EncodeObject = $Kernel::OM->Get('Kernel::System::Encode');

            my %Data;

            while ( my @Data = $DBObject->FetchrowArray() ) {

                # convert body

                $Data[5] = $EncodeObject->Convert(

                    Text  => $Data[5],

                    From  => $Data[2],

                    To    => 'utf-8',

                    Force => 1,

                );

                # convert subject

                $Data[3] = $EncodeObject->Convert(

                    Text  => $Data[3],

                    From  => $Data[2],

                    To    => 'utf-8',

                    Force => 1,

                );


I changed this in =>

            # convert subject


            $Data[4] = $EncodeObject->Convert(


                  Text  => $Data[4],


                From  => $Data[2],


                To    => 'utf-8',


                Force => 1,


            );


Before my fix it was not work well, it was not converting subject. Sombody makes mistake there, and insted subject it is converting notification_language.

And as I mentioned in my opinion it is good fix and it could be nice if there is the same behavior for all notification in OTRS.
And if you think that it is not necessary in this moment, we definitely should fix the issue with NotificationGet() in Notification.pm, it is real bug.
I hope that I help you, if you have any question let me know. 

Regards
Zoran
S7

luca-ianez:
Hi all,
I don't know if it can help but I had the same problem with auto responses.
We recently switch from a custom installation of OTRS to the appliance which is shipped with OTRS 4.0.6 free and PostgreSQL. We configured the system to send an autoresponse when a user open a new ticket, but we had the same error posted on top that says: "FEHLER: ungültige Byte-Sequenz für Kodierung »UTF8«: 0xa0". This error was triggered in the ArticleCreate function of Article.pm when it tries to insert the autoresponse generated by the TemplateGenerator. 
We temporarily fixed this by finding all 0xA0 characters in the subject and in the body of the autoresponse and replacing them with 0x20 characters just before the INSERT query at line 284.

    $Param{Subject}=~tr/\xA0/\x20/;
    $Param{Body}=~tr/\xA0/\x20/;

Maybe it would be better to fix this in the Template Generator code.

Regards
Luca

s7design:
Hi Luca,

Thanks for your sugestion. I will investigate about this. Your opinion is welcomed, and with this we can improve OTRS. I am not sure if this fix is nesessary in TemplateGenerator, but I will check this. With this fix, it will need some change in the others modules. Did you try marge this fix in your system. It would be nice if you could give me any feedback about this fix.
Now, there is encoding in this place, actualy it could be moved somwhere else but in this case we change logic in the others Notification  modules but behavior stay the same. I will check is it posible to do this in _Replace() function for all Article parts if it is realy necessary. I say again your experiance and bechavior of OTRS with PostgreSQl will be welcomed.
The fix in this PR is nessessary, becouse there are some errors in Notification.pm. and we fixed it as well.

We can discuss about this, I am looking forward your feedback.

Regards
Zoran

mgruner:
Hi Zoran. 
Unfortunately I was not able to reproduce this problem on my postgresql 9.4.1. Would you be able to provide a unit test case for this that can demonstrate the problem?
Besides, with your changes the NotificationEvent.t causes errors, but that is probably because of problems in the test (it inserts utf8 data as ISO and then fails on comparing it) that would have to be fixed too.
In general I would first like to understand where exactly things break why. Can you please summarize this?
Thanks and best regards, mg

carlosfrodriguez:
Hi @s7design and @mgruner,

I've extracted the fix for the wrong encoded column from Kernel::System::Notification and I've tracked this as a separate bug in http://bugs.otrs.org/show_bug.cgi?id=11207.

@s7design please rebase this pull request.

I believe this discussion will get easier as now the topic is whenever we need to encode NotificationEvent or not.

((enjoy))
cr

s7design:
Hi,

Thanks @carlosfrodriguez, I think that it is good idea. I rebased this PR, now there is only fix for NotificationEvent. 
I will provide unit test for wrong behavior in PostgreSQL. I am not sure if there is bug with all versions of PostgreSQL. As I mentioned before, I use PostgreSQL 9.1.11.

Regards
Zoran

s7design:
Hi for all,

I have detailed looked into this bug again. I started to make unit test. It should be Selenium test, actually. I can do that, but maybe in this moment it is not so important. 

I found that this bug occurs only if Rich text is enabled, and off course only for PostgreSQL.
The following is my test case:
* Step 1: Enable Rich text config item
* Step 2: Make Notification event: Events(e.g.Ticket create), TicketFilter(e.g. Queue - all), Recipient (Customer), Notification ("Test notification&nbsp; - in text source view)
* Step 3: Create Ticket
* Step 4: Look in SysLog there is error 

ERROR: invalid byte sequence for encoding "UTF8": 0xa0, SQL: 'INSERT INTO article (ticket_id, article_type_id, article_sender_type_id, a_from, a_reply_to, a_to, a_cc, a_subject, a_message_id, a_message_id_md5, a_in_reply_to, a_references, a_body, a_content_type, content_path, valid_id, incoming_time, create_time, create_by, change_time, change_by) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, 

* Step 4: Notification mails is sent 
Customer maybe will not see this bug, because he received a mail, but there is error log is SysLog

As I mentioned, I will be able to make Selenium test for this, but I am quiet sure that this fix is necessary.
Please give me feedback what is behavior of your system with my test case.

regards
Zoran

s7design:
Hi @mgruner,

I have added Selenium test for checking test case that is described in bug report.
As I mentioned before, it works only if PostgreSQL is used for DB. This Selenium is not so good as I wanted, I have a problem with clean up created test notification. If notification is deleted or set to invalid, the notification event will not be triggered. Because of that I commented part of code in the test.  But I hope you can run it on system with PostgreSQL. I could you simple unit test here, but Selenium test is much better now :)

Please give me feedback after testing it.

Regards
Zoran

s7design:
I close it, it will be rebased and solved in master

Regards
Zoran

s7design:
.

PR #447, opened 2015-02-06
[Rel4_0 - Bug 11011] OTRS does not honor customer change

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11011


mrcbnsls:
Thanks! Merged in f6f4ff2 (rel-4_0) and e681f54 (master).

PR #449, opened 2015-02-09
added code for clean up system after testing

s7design:
Hi @mrcbnsls ,

We added clean up part in these selenium tests. These tests  was  earlier, and I think that with our fix there is better state of system after running selenium tests.
There is used such logic in tests which we have added so far.

Regards
Zoran


mrcbnsls:
Good idea, merged in 7609d21. I only added a cache cleanup to the queue UT, otherwise the cache would be incorrect. Thanks!

PR #450, opened 2015-02-11
Master preference language

s7design:
Hi @mrcbnsls,

While we tested modules for porting to OM, we saw that translating in fronted doesn't work well. 
In customer interface it did not work at all, and in agent interface it did not work for  layout messages (e.g login messages)
I have not opened new bug for this issue, if you want I can do it.
Please, let me know what you think about this.

Regards
Zoran

### DIFF
--- a/Kernel/System/Web/InterfaceAgent.pm
+++ b/Kernel/System/Web/InterfaceAgent.pm
@@ -87,15 +87,14 @@ sub new {
         },
     );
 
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Yes, you are right, there were some issues. I fixed it another way here: https://github.com/OTRS/otrs/commit/85395044e43a9616d4e96bb25c1febbabde52a2c

PR #451, opened 2015-02-11
Add some ported modules

s7design:
Hi @mrcbnsls,

There are some ported modules to OM. I am not sure how we can do Selenium test for these modules.
Do you have any suggestions how test cases we can use?
These modules are a little bit specific. Is it necessary Selenium tests for them. 
Please let me know your opinion about it.

Regards
Zoran


mrcbnsls:
It's okay for now to not have any tests for these modules. I think we can add them later as they will be a bit more complex.

Merged in a549446, thanks.

PR #452, opened 2015-02-11
Master user preferences

s7design:
Added port and Selenium test for AgentPreferences and CustomerPreferences.
CustomerPreferences require https://github.com/OTRS/otrs/pull/450 for Selenium test to work.
If there any suggestion or feedback regarding required pull request, let us know, and we will work on it.


mrcbnsls:
Nice, merged in b926164. Thanks!

PR #453, opened 2015-02-11
Changed format for all successful messages in customer login screen

s7design:
Hi @mrcbnsls,

We found again something it could be improved in our opinion. We saw this while we tested ported modules to OM.
I am not sure if it is the best workflow that we add new PR for all this situation when we see something wrong in the code. If you have any suggestion about this please let me know. 
We only want  to improve OTRS generally, but we don't want to make mass in your organization of work.

Regards
Zoran


mrcbnsls:
Hi @s7design, good idea, but it should be done in master.

s7design:
Hi @mrcbnsls 
I can make another PR for master. It is not a problem.
I will do this soon

Regards
Zoran

mrcbnsls:
Thanks!

PR #456, opened 2015-02-18
[Rel-4_0 - Bug 10995] Characters not correctly managed by OTRS scripts.

s7design:
Hi @mrcbnsls,

There was a problem only in command line printing. Service was added well in DB (I use mysql utf-8 encoding, and I tested there).

I tested it with something like this:
     bin/otrs.AddService.pl -n 'Açãošđčćžшђчћжüöä-Test_Service' -c '3 normal'

I hope it is fixed well. If you have any suggestion, please let me know. If you agree with me, we could solve, in the same way, the others scrips in /bin directory (e.gotrs.AddRole, otrs.AddGroup.... ) 
I will make another PR for the others scripts.

Regards
Zoran

### DIFF
--- a/bin/otrs.AddService.pl
+++ b/bin/otrs.AddService.pl
@@ -23,6 +23,7 @@ use strict;
 use warnings;
 
 use File::Basename;
+use Encode;
 use FindBin qw($RealBin);
 use lib dirname($RealBin);
 use lib dirname($RealBin) . '/Kernel/cpan-lib';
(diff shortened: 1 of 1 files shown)


mgruner:
Hi S7,
please take a look at my solution in 25019f7, I think you might not have fully understood the code in Encode->new(). The problem with OTRS 4 is that the EncodeObject is not global any more, so that the code to initialize STD* and ARGV was not executed (or too late). Doing this without the EncodeObject AND also locally in all files is definitely not a good option.
Best regards, mg

PR #459, opened 2015-02-20
Patch Oracle connection in WebInstaller

Giermann:
WebInstaller was unable to create Oracle database.
I found the wrong usage of "$Self->{}" instead of "$DB{}" for DBHost and DBSID, which broke connection to Oracle. On the other hand the correct connection string was stored in ConfigDSN.
Second, the "test database connection" used another naming, requiring the short "SID" in config parameter "DBSID", but the resulting ConfigDSN used the Service Name (in my case the SID extended by or domain name). It should be unique for connection test and real connection and I prefer the Service Name, so I changed the DSN for the connection string.

Maybe someone should care about the naming "SID" now, but I'm fine since it at least works now.

Further, the separate commit will store DBPort and DBSID inside DatabaseHost and Database(name) to avoid having them only in the DSN.

### DIFF
--- a/Kernel/Modules/Installer.pm
+++ b/Kernel/Modules/Installer.pm
@@ -599,7 +599,7 @@ sub Run {
 
             # set DSN for Config.pm
             $DB{ConfigDSN} = 'DBI:Oracle://$Self->{DatabaseHost}:' . $DB{DBPort} . '/' . $DB{DBSID};
-            $DB{DSN}       = "DBI:Oracle://$Self->{DBHost}:$DB{DBPort}/$Self->{DBSID}";
+            $DB{DSN}       = "DBI:Oracle://$DB{DBHost}:$DB{DBPort}/$DB{DBSID}";
             $Self->{ConfigObject}->Set(
                 Key   => 'Database::Connect',
                 Value => "ALTER SESSION SET NLS_DATE_FORMAT = 'YYYY-MM-DD HH24:MI:SS'",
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @Giermann, thank you for your contribution. Could I ask you to sign our CLA as described on http://otrs.github.io/ first? Thank you for your cooperation.

Giermann:
Just sent it!
Sorry, I noticed that I'm forced to accept it, but didn't sign and send it before :)

carlosfrodriguez:
Hi @Giermann,

I fully agree with the changes in line 602 changing $Self to $DB, but i don't understand why to change line 601 at all. 

Additionally, normally in Config.pm in 'DatabaseHost', we use to store only the name of the Host or its IP Address, without any port, as it is not required for other Database connections like MySQL and PostgreSQL, so I trend to prefer to still only use it that way and add the port to the connection string for only the DBs that requires it. Please take a look to: [Kernel/Config.pm.dist](https://github.com/OTRS/otrs/blob/rel-4_0/Kernel/Config.pm.dist)

((enjoy))
cr

Giermann:
Well, I'm fine with both.
I have to admit, that I'm still not that deep in the OTRS sources...

Looking at the examples in Config.pm.dist, you already advise to use a hard coded port, but re-use the "Database" setting for the Service Name. Maybe it would be better to totally remove "DBSID" then and use "DBName" for Oracle too?
Further I found the ENV setting of NLS_LANG and NLS_DATE_FORMAT mandantory, contrary to ORACLE_HOME which is being needed on DBD::Oracle installation, but not while using it.
Is there any easy way to uncomment those two settings with the WebInstaller while configuring an Oracle database?

Giermann:
@carlosfrodriguez,
I changed the DBPort part to be only in the DSN. The "ReConfigure Config.pm" part now only contains the differences "DBSID" vs. "DBName" - should I go on modifying [InstallerDBoracle.tt](https://github.com/OTRS/otrs/blob/rel-4_0/Kernel/Output/HTML/Standard/InstallerDBoracle.tt) to use DBName instead of DBSID along with all references in [Installer.pm](https://github.com/OTRS/otrs/blob/rel-4_0/Kernel/Modules/Installer.pm)?

Giermann:
Is anything missing on my comments?

carlosfrodriguez:
Hi @Giermann 

This pull request was finally merged in https://github.com/OTRS/otrs/commit/7ec24dac7564e4b94156adcb96c8a1012dc1bc16

Thank you for your contribution

((enjoy))
cr

PR #460, opened 2015-02-23
fixed url for link in NoPermission screen

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11059

There was a problem with recognising URL link from variable. With fix it will be  interpolated accordingly.

### DIFF
--- a/Kernel/Output/HTML/Standard/NoPermission.tt
+++ b/Kernel/Output/HTML/Standard/NoPermission.tt
@@ -51,7 +51,7 @@
 [% RenderBlockStart("PossibleNextActionContainer") %]
                 <p class="SpacingTop">
 [% RenderBlockStart("PossibleNextActionRow") %]
-                    <a href="[% Data.Link %]" title="[% Translate(Data.Description) | html %]">[% Translate(Data.Description) | html %]</a>
+                    <a href="[% Data.Link | Interpolate %]" title="[% Translate(Data.Description) | html %]">[% Translate(Data.Description) | html %]</a>
                     <br/>
 [% RenderBlockEnd("PossibleNextActionRow") %]
                 </p>


carlosfrodriguez:
Hi @s7design,

I try to recreate the problem of the URL by trying to zoom into a non existing ticket, and the Interpolate filter is not working for me, can you take a look again if you can find what is going on?

BTW this only fixes this issue and not the referred bug.

((enjoy))
cr

s7design:
Hi @carlosfrodriguez,

Thanks for follow-up. I have just detailed looked at this bug again. In my opinion this fix is well and enough to solve all that is necessary in the bug. I added only interpolate directive and I am absolutely sure that it is needed because of interpolation appropriate value of variable in the link tag for  "Go to dashboard!"

This works well for me. To be honest I am not sure what was problem for you to reproduce the bug.
I also tried that you mentioned "zoom into a non existing ticket". This also works for me. 

And regarding to showing Spam and Delete menu items if agent doesn't have permission, I think that this is not needed to be fixed, and this is not a bug. Access control to show or not show these links can be done by using Key "Group" and Content like "rw:group1;move_into:group2", as it is noted in description of config items.
And in my opinion this is current behavior of system. 
The same situation is with dropdown menu item for move if agent does't have move_into permission. In this case there is menu item "Move" without Queue, and similar situation is with menu item for move function in Large ticket overview screen.

 As I mentioned above I believe that for this bug is enough to solve issue with link, and for the others things I can add comment on bug tracker as info for reporter. Maybe I could add comment before, but I wanted to consult with somebody of you,
Also, I will be able to change all of that if you want to change this behavior. We can discuss about all of that. If you have any suggestion, please let me know.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

I agree that the Interpolate filter is needed, but somehow it is not working correctly for me. I will try again on Monday

Thank you

((enjoy))
cr

carlosfrodriguez:
Hi @s7design,

I'm very sorry I had the setting modified, with a OTRS 3.3 style link, what's why the Interpolate filter was not working for me.

((enjoy))
cr

carlosfrodriguez:
Hi @s7design,

Thanks for your pull request, this has been merged in https://github.com/OTRS/otrs/commit/418abd3503fb64ade18a346e0d6528fb716e646a.

Can you please comment but the other part of the bug in bugzilla?

((enjoy))
cr

PR #461, opened 2015-02-24
[Rel-4_0 - Bug 10809] OTRS tags aren't working within the bounce function

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10809

In bounce function tags are little bit different then in other modules. We think that for TicketNumber it is possible to have right away replaced value, yet that is not the case for <OTRS_BOUNCE_TO> since bounce receiver isn't known until completion of request.


mgruner:
You are right, and the ticket number is also replaced correctly at send time. I think we can keep it like this.

PR #464, opened 2015-02-25
Fix bug#11106 - Owner type is ignored if set previous owner to brank Age...

noritnk:
Fix bug#11106 - Owner type is ignored if set previous owner to brank AgentTicketOwner.
See http://bugs.otrs.org/show_bug.cgi?id=11106 .


carlosfrodriguez:
Hi @noritnk ,

Thanks for your pull request, it was merged in https://github.com/OTRS/otrs/commit/12eec34fbb02b019db7d2f9b1148a6db40f0af9f

((enjoy))
cr

PR #465, opened 2015-02-25
Added messages when no result shown

s7design:
While working on a bug #11067 ( bugs.otrs.org/show_bug.cgi?id=11067 ) we found out that in some modules there was no message when filter was used and no matches were found. 
Hereby is our suggestion for missing parts, so system respond same in all places.

### DIFF
--- a/Kernel/Modules/AdminCustomerUserGroup.pm
+++ b/Kernel/Modules/AdminCustomerUserGroup.pm
@@ -408,6 +408,16 @@ sub _Change {
         );
     }
 
+    # check if there are groups/customers
+    if ( !%Data ) {
+        $LayoutObject->Block(
+            Name => 'NoDataFoundMsgList',
+            Data => {
+                ColSpan => 3,
+            },
+        );
+    }
+
     if ( $NeType eq 'CustomerUser' ) {
 
         # output count block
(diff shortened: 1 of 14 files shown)


carlosfrodriguez:
Hi @s7design,

I like it a lot, but I found 1 disadvantage, when you don't have any element for example in AdminTemplateAttachment, if you don't have any attachment 
1 Click on a Template, the list of attachments is totally empty
2 Search for an attachment, "No matches found" appear correctly
3 Delete all characters from filter, the "No matches found" is still there

This behavior does not happens for example in AgentCustomerService, when you have no services is displays "No data found" if you search it displays "No matches found" if you delete the search characters it returns to "No data found"... I think this should be the behavior also in other screens.

Do you think it is possible to also update the other screens to behave just like AgentCustomerService?

((enjoy))
cr

s7design:
Hi @carlosfrodriguez 

I have fixed all screens as you suggested. I believe that now all screens have the same behaviour. Please check it. If you have any suggestion about this we will be able to change it.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

Thanks for the follow-up, this pull request was merged in https://github.com/OTRS/otrs/commit/8d85d6e27e8eb6ff42c49040f269f9f4341ef04c.

((enjoy))
cr

PR #467, opened 2015-02-26
Fix bug#11113 - Internal server error at importing with IMAP

noritnk:
Fix bug#11113 - Internal server error at importing with IMAP.
See http://bugs.otrs.org/show_bug.cgi?id=11113 .


mrcbnsls:
Merged in 3d21c1f and 94679a1 (master). Thanks!

PR #468, opened 2015-02-26
[Rel-4_0 Bug 10961]   Adding email recipients via addressbook does not update customer information

s7design:

Hi @mrcbnsls and @zottto,

I fixed this bug with JSON data as @zottto suggested. I hope it is well now. I leave CustomrAutoComplete for fields in Address book. @zottto commented about this on previous PR for this bug (https://github.com/OTRS/otrs/pull/403) In my opinion it is ok for forward or reply ticket, but I could agree with you, it maybe is not necessary here. There is have CustomerAutoComplete on the parent screens. 

I use hidden filed for saving customer data, because I have some problem with data-customerdata attribute in HTML tag. I think that it is better option now. (You can see my  incomplete solution in previous commits here)

You can see I use JSON object in AgentBook, and because of that I add  "$ObjectManagerDisabled = 1;" We prepared for porting to OM this module recently in some PR. I can finish porting to OM here in this PR if you think that it is good idea. Also I will be able to make another PR for master if it is necessary.

We can discus about all of that. I tried to do it the best I know now, but in this comment I said my dilemmas and if you have any suggestion for improvement let me know. 

Regards
Zoran
If you want I will be able to make another PR for master.
Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentBook.pm
+++ b/Kernel/Modules/AgentBook.pm
@@ -56,7 +56,10 @@ sub Run {
             User => $_,
         );
         if ( $CustomerUserData{UserEmail} ) {
-            $List{ $CustomerUserData{UserEmail} } = $CustomerUserList{$_};
+            $List{ $CustomerUserData{UserEmail} } = {
+                Email       => $CustomerUserList{$_},
+                CustomerKey => $_
+            };
         }
     }
 
@@ -71,13 +74,14 @@ sub Run {
         );
 
         my $Count = 1;
-        for ( reverse sort { $List{$b} cmp $List{$a} } keys %List ) {
+        for ( reverse sort { $List{$b}->{Email} cmp $List{$a}->{Email} } keys %List ) {
             $Self->{LayoutObject}->Block(
                 Name => 'Row',
                 Data => {
-                    Name  => $List{$_},
-                    Email => $_,
-                    Count => $Count,
+                    Name        => $List{$_}->{Email},
+                    Email       => $_,
+                    Count       => $Count,
+                    CustomerData => $Kernel::OM->Get('Kernel::System::JSON')->Encode( Data => { $_ => $List{$_}->{CustomerKey}  } ),
                 },
             );
             $Count++;
--- a/Kernel/Modules/AgentBook.pm
+++ b/Kernel/Modules/AgentBook.pm
@@ -79,11 +79,11 @@ sub Run {
             $Self->{LayoutObject}->Block(
                 Name => 'Row',
                 Data => {
-                    Name  => $List{$_}->{Email},
-                    Email => $_,
+                    Email => $List{$_}->{Email},
                     Count => $Count,
                     CustomerDataJSON =>
-                        $Kernel::OM->Get('Kernel::System::JSON')->Encode( Data => { $_ => $List{$_}->{CustomerKey} } ),
+                        $Kernel::OM->Get('Kernel::System::JSON')
+                        ->Encode( Data => { $List{$_}->{Email} => $List{$_}->{CustomerKey} } ),
                 },
             );
             $Count++;
--- a/Kernel/Modules/AgentBook.pm
+++ b/Kernel/Modules/AgentBook.pm
@@ -82,7 +82,7 @@ sub Run {
                     Name  => $List{$_}->{Email},
                     Email => $_,
                     Count => $Count,
-                    CustomerData =>
+                    CustomerDataJSON =>
                         $Kernel::OM->Get('Kernel::System::JSON')->Encode( Data => { $_ => $List{$_}->{CustomerKey} } ),
                 },
             );
--- a/Kernel/Modules/AgentBook.pm
+++ b/Kernel/Modules/AgentBook.pm
@@ -13,6 +13,7 @@ use strict;
 use warnings;
 
 use Kernel::System::CustomerUser;
+our $ObjectManagerDisabled = 1;
 
 sub new {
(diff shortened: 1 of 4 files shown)


mrcbnsls:
@zottto, your turn :)

zottto:
Hi Zoran,

thanks for your commits. As long as I use the search field in the adress book to add customers to the ticket, your solution works well. But I have some problems using the auto-completion:

1. If I use the auto-completion on the To field in the adressbook and choose a customer from the autocompletion list, this user is added to the To field, but the customer information in the sidebar is still not updated.

2. If I add a customer via autocompletion and than use the search field to add another customer, only the customer added after the use of the search field is added to the ticket. The customer added via autocompletion is lost.

3. If I add a customer via autocompletion, the entry is "User Name" <user.name@domain.com>, but if I add a user via the search field, only the email adress user.name@domain.com is added to the field.

Please check, that this inconsistencies are fixed.

Marc

s7design:
Hi @zottto,

Thanks for your detailed  follow-up. I totally agree with you, there are  inconsistencies with auto-completion here. The most of that I saw while I did on it, but I was not sure what is the best solution here.
Regarding on:
  1.  If I use the auto-completion on the To field in the adressbook and choose a customer from the autocompletion list, this user is added to the To field, but the customer information in the sidebar is still not updated.

  Yes, you are right. Because of that I mentioned in my previous comment that I leave auto-completion   here, maybe it will be useful for forward and reply screen.

   2. If I add a customer via autocompletion and than use the search field to add another customer, only the customer added after the use of the search field is added to the ticket. The customer added via autocompletion is lost.

   There is a problem that I have not seen before.

   3. If I add a customer via autocompletion, the entry is "User Name" user.name@domain.com, but if I add a user via the search field, only the email adress user.name@domain.com is added to the field.

    I have seen this, and the same behavior is if we use "Customer option". It adds the email address in format: user.name@domain.com 

Now, while I look on this problems with auto-completion here, maybe the best option if we cancel it. As I mentioned in  previous comment, there is auto-completion on the parent screen of all modules that use "Address book" feature.

If we still think that it is necessary for this module, I will try to do that. I suppose that I will have to add some if statements in code where auto-completion is done. I believe if we do that, your help will be welcomed. 

Sorry, because I ask a lot of questions and I have many dilemmas about it, but it is easier for me if I consult with you.
I believe that it will done the best we can.

Regards
Zoran

zottto:
Hi Zoran,

I suggest to remove autocompletion here. The user has autocompletion in the parent screen and can use it there. There is no need to open the adress book for that. This will make it easier for you in the adress book screen.

Still one thing to check is how the user name / email is displayed. If your choice within the adress book is applied to the parent screen, the user is only shown as email adress and not with the user name - like it is when using the autocompletion in the parent screen. It would be nice, if this could look the same.

Marc

s7design:
Hi, @zottto 

Thanks for follow-up. I agree with you, it is the best solution.
It will be done soon.

I believe that I will be able to do a format of an email like: "User Name" user.name@domain.com,

Regards
Zoran

s7design:
Hi @zottto, 

I have finally done this. Now, a format of an email is like: "User Name" user.name@domain.com.
I hope it works well now.

Regards
Zoran
S7

zottto:
Hi @s7design,

the main functionality now works fine. But I have two additional comments:

1. If I add more than one customer to a field (e.g. "To"), the addition of the second customer is not visible anymore, because the string is longer than the width of the input field. It would be nice to always set the cursor to the end of the field for a better UX.

2. in Core.Agent.TicketAction.js you replace some HTML entities, but you do this always on the complete concatenated string. It is only needed for the part that is added in this run. Please also check if you need to convert more special characters or entities or you can use a generic JS function for that.

Marc

s7ncvijanovic:
Hi @zottto, 

Thanks Marc for your detailed follow-up. Your advices are always useful. I have done fix as you suggested.  I hope it is ok now. I added focus on field where email is added. I believe it is good for UX.
Some of that I have not known before, and thanks because I learned something new :)
You mentioned  a generic JS function, but I don't know if exist such function. I browsed about it, but I did not find anything, because of that I use replace function. 

If you have any new suggestion I will be apply it.

Regards
Zoran

s7design:
Hi @zottto,

Thanks Marc for your detailed follow-up. Your advices are always useful. I have done fix as you suggested. I hope it is ok now. I added focus on field where email is added. I believe it is good for UX.
Some of that I have not known before, and thanks because I learned something new :)
You mentioned a generic JS function, but I don't know if exist such function. I browsed about it, but I did not find anything, because of that I use replace function.

If you have any new suggestion I will be apply it.

Regards
Zoran

zottto:
Hi @s7design ,

The PR looks very good now. May I ask you to also make the changes in the OTRS 3.3 branch and send an additional PR for that? I will do the porting to the master branch.

Thanks,
Marc

s7design:
Hi @zottto ,

Thanks for follow-up. Finally it is good :)
But it is much better with your suggestion, thanks.
I will provide PR for OTRS 3.3, it will not be a problem. It will have done soon.

Regards
Zoran

s7design:
Hi,

Here could be a problem during the merging, because in meantime porting to OM has done.
If there is a problem, we will be able to check this.

Regards
Zoran

zottto:
Merged with 2bfa2e1 to rel_4-0.
@s7design: Please provide a working PR for the master branch.

Thanks,
Marc

s7design:
Thanks @zottto ,

I close this PR, and we will provide another PR for master soon.

Regards
Zoran

PR #470, opened 2015-02-27
[Rel-4_0 Bug 11123] DynamicFieldText RegEx message problem on create and edit

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11123

### DIFF
--- a/Kernel/Output/HTML/Standard/AdminDynamicFieldText.tt
+++ b/Kernel/Output/HTML/Standard/AdminDynamicFieldText.tt
@@ -173,16 +173,16 @@
                             <fieldset id="RegExRow_[% Data.EntryCounter | html %]" class="TableLike SpacingTop RegExRow W50pc">
                                 <label class="Mandatory" for="RegEx_[% Data.EntryCounter | html %]"><span class="Marker">*</span>[% Translate("RegEx") | html %]:</label>
                                 <div class="Field">
-                                    <input id="RegEx_[% Data.EntryCounter | html %]" class="W80pc [% Data.RegExServerError | html %]" type="text" maxlength="500" value="[% Data.RegEx | html %]" name="RegEx_[% Data.EntryCounter | html %]" />
+                                    <input id="RegEx_[% Data.EntryCounter | html %]" class="W80pc [% Data.RegExServerError | html %] Validate_Required" type="text" maxlength="500" value="[% Data.RegEx | html %]" name="RegEx_[% Data.EntryCounter | html %]" />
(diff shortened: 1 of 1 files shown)


mgruner:
Please rebase.

s7design:
Hi @mgruner 

I rebased it, I hope you can merged it now.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

I can't reproduce the error at 100% in my system the correct classes are sent from Perl to the tt so they don't need to be explicitly added, on the other hand I agree that the "} is displayed and it should be removed, 

Can you check again and update the PR and the bug is necessary?

s7design:
Hi @carlosfrodriguez 

Thanks for follow-up. I have just looked into this PR and bug report and I am sure that all of my fixes is needed. I will analyze that step by step. You are right that you cannot reproduce all cases. There are three casees in bug report. I will explain each of them

* Regarding to Case 1:
```
  1) Go to AdminDynamicFieldText and fill in all required fields. 
  2) Add RegEx fill in "Error Message" and leave RegEx field empty. 
  3) Click submit, no fronted message appear.
```
It is connected to adding new dynamic field, and is solved in meantime. (solved by @rolfschmidt in https://github.com/OTRS/otrs/commit/158434c224f0c79626b30792d0ce84e036aa1930 - line 199) I rebased this PR recently. Because of that this case is not valid anymore. I will update bug report, thanks for that maybe I should do that before.

* Regarding to Case 2:
```
  1) Fill in RegEx field, submit, and enter to change created DynamicFieldText
  2) Delete RegEx field and submit. No message appear with successful submit, but RegEx is removed from DynamicFieldText
```
It is connected to editing created dynamic field. Without my fix this feature doesn't work properly, and it is definitely a bug. There is a problem if you want to edit dynamic field (text or textarea ) that there is RegEx. In that case if you delete "Error Message:" input field of RegEx. block, whole RegEx. will be deleted. It happened because fronted validation for required field doesn't work. This is solved with adding class "Validate_Required".

* Regarding to Case 3:
```
  1) Create DynamicFieldText with RegEx.
  2) Enter to change it.
  3) Delete "Error Message" and submit. Server error appears: 

         This field is required."}
  4) ( "} ) is added additionally unnecessary.
```
As you said it is necessary because this is type error. It is needed for server side errors.

I hope that my additional explanation is helpful. If you have any question feel free to ask, I am available on Skype as well.

Regards
Zoran

rolfschmidt:
@s7design :+1: :+1:

carlosfrodriguez:
Hi @s7design,

Thank you for your pull request, it was finally merged in https://github.com/OTRS/otrs/commit/86ee2b7cf5197d55866f2caac6ccb6b1d0ea557d.

((enjoy))
cr

PR #476, opened 2015-03-05
Enable to disable to force to rewrap a text input in text area

noritnk:
In recent years, we don't have to rewrap the text in many cases by
following reasons.  This PR enables to disable to force to rewrap a text
in text area as article body.

  - Recent screen is large.  It can display more than 150 characters of
    a line without scroll.
  - Even when send email, body text is encoded quoted-printable or
    base64.  It means that we don't have to mind "Line Length Limits"
    written in RFC 5322 Section 2.1.1.
  - Japanese and Chinese does not use spaces to separate between words
    in locale.  It causes to ignore hard wrap trick in these locales.
    By the way, I confirmed Firefox and Internet Explorer expectly wrap
    a text even in these locales with "word-wrap: break-word".
  - If Frontend::RichText is not "0", a text is not rewrapped.

By the way, I see that the feature to rewrap a text in text area is
introduced in fix for bug#526.  At first "hard" attributes of textarea
is used, it is changed into Perl code in the fix.  However, now, if
"word-wrap: break-word" works correctly, a horizontal scrollbar will be
not caused for a long line, and the line will be wrapped expectedly by
browser's feature.

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -384,6 +384,17 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Frontend::WrapText" Required="1" Valid="1">
+        <Description Translatable="1">Enables or disables to force to rewrap a text input in text area.</Description>
+        <Group>Framework</Group>
+        <SubGroup>Core::Web</SubGroup>
+        <Setting>
+            <Option SelectedID="1">
+                <Item Key="0">No</Item>
+                <Item Key="1">Yes</Item>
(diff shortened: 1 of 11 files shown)


mgruner:
Hello @noritnk. Thanks a lot for your suggestion. I am not sure if this is so simple; changing the ToAscii method would change the behaviour everywhere, also for the emails that OTRS sends. And we don't have control over the mail clients that customers would use.
Besides that, there are merge conflicts, so I'll close the PR for now. @mrcbnsls: FYI, what do you think about this?

PR #480, opened 2015-03-09
Customer user information center

nilei:
In order to not only have a detailed view on one customer, this module allows a detailed view on a customer user. It also changes the customer information center so that a click on one of the users redirects to this customer information center.

Help needed: 
- Filters in dashboard modules don't work


nilei:
@mrcbnsls , could you have a look into this and help me to solve my issues with the filters?

mrcbnsls:
Hi @nilei, 

Thanks, nice! We like this idea and think its very useful. But there seems to be a lot of code duplication. We think it would be better to integrate this into the code of CustomerInformationCenter rather than creating a whole new module with all dependencies. So I'd ask you to re-create this PR with the requested changes. I'll be glad to support you getting the filters working afterwards.

Marc

mgruner:
I'll close this PR for now. @nilei, we are very much looking forward to your contribution!

PR #482, opened 2015-03-11
[Rel-4_0] Added translation for Templates tag

s7design:
While working on bug #11117 and #11118 we found out that there is missing similar functionality for Templates. This is our suggestion for it. If needs be, we can open bug report for this issue and then create appropriate PR for it.

We are only looking forward to improving OTRS experience for all users.

Regards,
S7

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -410,6 +410,19 @@ sub Template {
         );
     }
 
+    # get ticket
+    my %Ticket = $Kernel::OM->Get('Kernel::System::Ticket')->TicketGet(
+        TicketID => $Param{TicketID},
+    );
+
+    # get recipient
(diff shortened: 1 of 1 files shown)


mgruner:
Hi S7,
hm, but the TicketID is optional, isn't it? Also if you see this as a bug, please open a bug report for this and provide instructions to reproduce it. A unit test would be also nice.
Regards, mg

s7design:
Hi @mgruner ,

After some internal analysis we concluded that this functionality is not applicable for all situations where templates could be used. You are right about TickedID being optional.
Internal communication between agents, while using templates, shouldn't depend on customer user language preference. There are certain specific situation where this functionality could be used in a desired way. 
We will close this PR. If you have any further advice on this situation we can come back to it and work further on.

Regards,
S7

s7design:
.

PR #484, opened 2015-03-12
Master port to OM 12_03

s7design:
Hi @mrcbnsls and @mgruner ,

There are ported modules to OM. We are tested it manually, and I hope it was done ok.
These modules is big, and maybe we did not test some cases, your feedback will be welcomed if we made some mistakes.
If you have any idea how we can make Selenium test for these modules, we will be able to add it. 

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminProcessManagement.pm
+++ b/Kernel/Modules/AdminProcessManagement.pm
@@ -14,19 +14,9 @@ use strict;
 use warnings;
 use Data::Dumper;
 
-use Kernel::System::YAML;
-
-use Kernel::System::JSON;
-use Kernel::System::DynamicField;
-use Kernel::System::ProcessManagement::DB::Entity;
-use Kernel::System::ProcessManagement::DB::Activity;
-use Kernel::System::ProcessManagement::DB::ActivityDialog;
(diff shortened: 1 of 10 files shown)


mrcbnsls:
Merged in 98b7d4d. Thank you, that was a huge one :-D

Some notes on the selenium tests. There were several problems with the process delete action in all tests which were caused by the fact that the selenium scenario would be shut down actually *before* the ajax request would be finished which deleted the process. That would result in a ajax error which would cause phantomjs to crash. I now fixed that by just deleting the process on database directly.

For the future we should take special care for cases where we need to wait for ajax requests to finish. You could do it for example like this:

```perl
my $ElementPresent = 1;

ACTIVESLEEP:
for my $Second (1 .. 20) {
	# check for the result of the ajax request (in this case,
	# the element should not longer be present)
	if ( !$Selenium->execute_script("return \$('#Name').length") ) {
		$ElementPresent = 0;
		last ACTIVESLEEP;
	}
	sleep 1;
}
```

s7design:
Thanks @mrcbnsls for follow-up.

We will use your suggestion for waiting for Ajax request.  
I think it is much better. 

Regards
Zoran

PR #485, opened 2015-03-13
Fix bug#11161 - ticket_history table is always joined in TicketSearch

noritnk:
This PR is fix for bug#11161.
See http://bugs.otrs.org/show_bug.cgi?id=11161 .

### DIFF
--- a/Kernel/System/Ticket/TicketSearch.pm
+++ b/Kernel/System/Ticket/TicketSearch.pm
@@ -500,7 +500,7 @@ sub TicketSearch {
     # use also history table if required
     ARGUMENT:
     for my $Key ( sort keys %Param ) {
-        if ( $Key =~ /^(Ticket(Close|Change)Time(Newer|Older)(Date|Minutes)|Created.+?)/ ) {
+        if ( $Param{$Key} && $Key =~ /^(Ticket(Close|Change)Time(Newer|Older)(Date|Minutes)|Created.+?)/ ) {
             $SQLFrom .= 'INNER JOIN ticket_history th ON st.id = th.ticket_id ';
             last ARGUMENT;
         }
--- a/Kernel/System/Ticket/TicketSearch.pm
+++ b/Kernel/System/Ticket/TicketSearch.pm
@@ -500,7 +500,7 @@ sub TicketSearch {
     # use also history table if required
     ARGUMENT:
     for my $Key ( sort keys %Param ) {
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @noritnk,

I'm not sure for this pull request, what happens if the value is 0? should it check in the history table or not?

((enjoy))
cr

noritnk:
Hi carlosfrodriguez,

Thanks for the review.

> what happens if the value is 0? 

The condition is ignored.  It defines not here but in the codes wich build where condition.
So if the value is 0, history table should not be joined.

Regards,
@noritnk

carlosfrodriguez:
Hi @noritnk,

I'm doing a 2nd review on this and I found that apparently the ticket_history table is used in another places, I can be wrong, but my feeling is that it might be that the history_table is always needed and the line you change in this pull request might be needed to be changed to include ticket history always (possibly in another place), please take a look at line 590

```perl
    # created type ids
    if ( $Param{CreatedTypeIDs} ) {

        # lookup history type id
        my $HistoryTypeID = $Self->HistoryTypeLookup(
            Type => 'NewTicket',
        );

        if ($HistoryTypeID) {

            # create sql part
            $SQLExt .= $Self->_InConditionGet(
                TableColumn => 'th.type_id',
                IDRef       => $Param{CreatedTypeIDs},
            );
            $SQLExt .= " AND th.history_type_id = $HistoryTypeID ";
        }
    }
```

This CreatedTypeIDs are normal search parameters and if the ticket_history table is not joint then the SQL should be wrong isn't it?

And as this there are other places in the code that are not really related to the Ticket times and needs the ticket_history table ;-( as fas as I can see.

Please let me know what do you think.

((enjoy))
cr

noritnk:
Hi @carlosfrodriguez,

Thanks for the second review.
After applying this PR, conditions that ticket_history table is joined is below.

+ if ( $Param{$Key} && $Key =~ /^(Ticket(Close|Change)Time(Newer|Older)(Date|Minutes)|Created.+?)/ ) {

So if $Param{CreatedTypeIDs} is defined and it is not emptry string, even if this PR is applied, ticket_history table is joined. And other conditions are also same as it.

Best regards,
@noritnk

carlosfrodriguez:
Hi @noritnk 

I think I missed the |Created.+?... sorry about that!

((enjoy))
cr

carlosfrodriguez:
Hi @noritnk,

I merge your pull request in https://github.com/OTRS/otrs/commit/65f5bdb67343b0570c32998952e871d5342ba60b, and added an small unit tests, just to be sure for now on that the ticket_history table is join when is needed.

Sorry for the long delay and the back and forward

((enjoy))
cr

PR #487, opened 2015-03-13
Port some modules for generic interface to OM 13_03

s7design:

PR #490, opened 2015-03-17
  Port rest of modules for generic interface to OM 17_03

s7design:


Kernel/Modules/AdminGenericInterfaceMappingSimple.pm
```diff
@@ -24,32 +22,21 @@ sub new {
...
-            $Self->{LayoutObject}->FatalError( Message => "Got no $_!" );
-        }
-    }
-
-    # create additional objects
-    $Self->{ValidObject} = Kernel::System::Valid->new( %{$Self} );
-    $Self->{WebserviceObject} =
-        Kernel::System::GenericInterface::Webservice->new( %{$Self} );
-
-    $Self->{DeletedString} = '_GenericInterface_Mapping_Simple_DeletedString_Dont_Use_It_String_Please';
```
carlosfrodriguez: Hi @s7design,

This  $Self->{DeletedString} can remain in the constructor.

((enjoy))
cr

carlosfrodriguez:
Hi @s7design,

Can you please update this pull request with the change in Kernel/Modules/AdminGenericInterfaceMappingSimple.pm please?


((enjoy))

s7design:
Hi @carlosfrodriguez,

Thanks for follow-up. I update this PR with your suggestion about $Self->{DeletedString}.

I saw your changes in another PR for porting to OM (https://github.com/OTRS/otrs/pull/487#event-263113165). I think it is useful and good changes. To be honest, some of these changes I didn't see, and the main work on these PRs is porting to Object manager. Sometimes we don't go trough whole code line by line, we only change objects with ObjectManager, because of that we didn't see some of your changes. It is not hard for us to go line by line, but sometimes it is not possible. In that case it would last too much time, you know that some modules is longer than 5000 lines. 
If we will see such inconsistent in the further work we try to correct this as you did.
I also see on that PR that you add comment like "get layout object" or "get param object", it is good suggestion. I will apply that in the further work, but we have already merged some modules. Maybe it could be nice one day, that we go trough these merged modules and add  such comments. In this moment I believe that it is more important that we continue on porting new modules. I think that you agree with me.

If you have any suggestion it is always welcomed.

Regards
Zoran

s7design:
Hi @carlosfrodriguez,

 As you can see, I added some comments for created object with OM. Also I changed "Webservice" to "web service" in comments and I changed "admins" with "administrators". All of that I saw in your merged code for porting to OM.

Regards and thanks
Zoran

carlosfrodriguez:
Hi @s7design,

Thank you for your changes, I merge this pull request here: https://github.com/OTRS/otrs/commit/0369d087e20f8ec00722211bc2751c36e272f41c

In general I like it so much, i think you are doing a great job and I appreciate it a lot. I made small changes, some of them leftovers from earlier versions nothing to do with the porting itself.

But I have 2 comments:

1 The selenium test is not working at all, apparently it imports some YAML files but they does not exists in the directory can you take a look on it? please create a separated Pull request, currently this part of the unit test is disabled

2 I found lines like 
```perl
# get needed objects
my $InvokerType = $Kernel::OM->Get('Kernel::System::Web::Request')->GetParam( Param => 'InvokerType' );
```

for OTRS 4 and beyond when we instantiate a new object variable we normally use this kind of comment
```perl
# get param object
my $ParamObject = $Kernel::OM->Get('Kernel::System::Web::Request');
```

for sure if they are multiple objects we can use the comment *# get needed objects*, but in the first example we are just getting a value from the web request so the comment is not accurate ;-)
 (just keep this for your records)

((enjoy))
cr

s7design:
Hi @carlosfrodriguez ,

Thanks for your detailed follow-up. I will use your suggestion in my further work.
All of that is welcomed, I like this because in this way we can improve code and my work generally as well.

Regardds
Zoran

PR #495, opened 2015-03-20
Fix bug#11185 - Mismatch between shown columns order and default columns...

noritnk:
This PR fixes bug#11185.
See http://bugs.otrs.org/show_bug.cgi?id=11185 .


mgruner:
Merged, thank you very much!

PR #496, opened 2015-03-20
[Master - Bug 11163] Uninitialized value warning in Excel output in AgentTicketSearch

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11163

### DIFF
--- a/Kernel/System/CSV.pm
+++ b/Kernel/System/CSV.pm
@@ -119,7 +119,9 @@ sub Array2CSV {
         my @ColumnLengths;
         my $Row = 0;
         for my $DataRaw ( \@Head, @Data ) {
+            DATARAW:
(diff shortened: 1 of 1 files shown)


mgruner:
Hm. When I do a simple ticket search and output to Excel, your patch destroys the excel file. It is cut after the first emtpy cell, no further cells are being populated. Does it work for you?

Kernel/System/CSV.pm
```diff
@@ -119,7 +119,9 @@ sub Array2CSV {
         my @ColumnLengths;
         my $Row = 0;
         for my $DataRaw ( \@Head, @Data ) {
+            DATARAW:
```
mgruner: This has the wrong name, it should be COL: if it is placed correctly here.

s7design:
Hi @mgruner ,

I hope that it is works now.

Regards
Zoran

mgruner:
Merged, thanks!

PR #497, opened 2015-03-20
Master customer ticket overview modul - port to OM

s7design:


mrcbnsls:
Merged in 215f153, thanks. Same thing here regarding check for ticket type and service, but that only a small issue which I fixed :)

PR #499, opened 2015-03-20
Master port to OM (customer ticket attachment )

s7design:


mrcbnsls:
Merged in 1e1a448, thanks. Please note that I added some code in the test to disable Ticket::Type and Ticket::Service in order to make sure these fields are not validated.

PR #435, opened 2015-01-31
Improvement of QueryCondition.

noritnk:
I see that QueryCondition is some potential bugs.  This PR makes a change to simplify QueryCondition.
This PR includes fix for bug#11022.

By the way, the change enables to parse an input with not the character unit but  the word unit.

PR #500, opened 2015-03-21
Improvement of QueryCondition

noritnk:
I port following PR for master, and added some comments.
https://github.com/OTRS/otrs/pull/435

### DIFF
--- a/Kernel/System/DB.pm
+++ b/Kernel/System/DB.pm
@@ -1064,30 +1064,35 @@ sub QueryCondition {
     $Param{Value} =~ s/^\s+//g;
     $Param{Value} =~ s/\s+$//g;
 
-    # add base brackets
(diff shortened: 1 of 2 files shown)


noritnk:
I added an change to previos PR which is for 4.0.  Both && || and || && expressions are replaced with ||.  they are treated as an error in OTRS 4.0, but I think that it is wrong.

If we write a || b && c, it is replaced a OR b AND c in SQL.  It is equal to a || (b && c).  So both && || and || && should be treated as || expression, as a && || b is equal to (a && ) b, a || && b is equal to a || (&& b).

mgruner:
Hello @noritnk. I honestly fail to understand what you are trying to achieve here. a && || b would be an invalid input that should cause failure IMHO. Please feel free to add more information!

PR #501, opened 2015-03-21
Fixed duble sessionID parameter in some links

s7design:
Hi,

While I worked on the bug http://bugs.otrs.org/show_bug.cgi?id=10962 I saw that there is issue on some links if it is set https protocol. The problem is that in some link is added double SessionID value. There are wrong links on Dashboard and AgentTicketOverview  screens (Queue, Status, Escalation...).
I had fixed this when I made the first PR for the bug 10962. Becouse of that I have just made new PR for this bug in FAQ repository, I fixed here the issue with double SessionID value.

Regards
Zoran


mgruner:
Very nice, thank you! Merged in master and rel-4_0.

PR #503, opened 2015-03-24
Master port om AdminSMIME

s7design:
Hi,

This is not my final version of porting AdminSmime.pm to  OM. 
There is a problem if SMIME is not enabled, or if SMIME environment is not set correctly. 

I have advice that we change AdminSmime and make something as it was in  AdminPGP screen. In AdminPGP if PGP is not enabled there was message box "This feature is disabled! ...
After @mrcbnsls  today's changes in commit https://github.com/OTRS/otrs/commit/f05d85f29f17db52fab73dc9355d03eac4cc5f3a this screen is a little bit changed.

What do you think about this. I will be able to fix AdminSMIME, and in another PR AdminPGP as well.
In that case we would have better UX.
It is only my opinion. Let me know if I can do that.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminSMIME.pm
+++ b/Kernel/Modules/AdminSMIME.pm
@@ -14,8 +14,6 @@ use warnings;
 
 our $ObjectManagerDisabled = 1;
 
-use Kernel::System::CustomerUser;
-
 sub new {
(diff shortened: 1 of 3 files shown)


mrcbnsls:
Merged in a49ef16, thanks.

PR #504, opened 2015-03-24
Master  CustomerProcessManagement port to OM

s7design:


mrcbnsls:
Merged in 2517baa, thanks! One very important note: Please make sure to use Kernel::System::CustomerGroup in customer modules. I found that you replaced those occurences with Kernel::System::Group, which obviously doesn't work. The selenium test showed that, which is nice. So I fixed it and would ask you to take some special care about that.

PR #505, opened 2015-03-24
Selenium test for GenericInterface

s7design:
Hi @carlosfrodriguez,

I fixed problem with web service yml files. I hope that Selenium test works now.
The problem was because yesterday these yml files were deleted, but they are necessary for my Selenium test and I added again them.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminGenericInterfaceWebservice.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminGenericInterfaceWebservice.t
@@ -70,71 +70,71 @@ $Selenium->RunTest(
             webserviceconfig_2 => 'Connector to send and receive date from Nagios 2.',
         );
 
-        # for my $Webservice (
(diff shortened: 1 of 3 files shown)


carlosfrodriguez:
Hi @s7design,

Thank you very much it works great, this pull request was merged in https://github.com/OTRS/otrs/commit/4b6ca4a6bed7008c3f6eeac1dc10436f505997b2.

By the way are you using a graphical web browser for you selenium environment like FireFox?
I trying to use PhantomJS but as soon as it tries to read the YAML it gets stuck and I need to use 'ctl c' 2 times to continue, for sure the test fail as the browser dies, but with FF everything works just fine.

((enjoy))
cr

s7design:
Hi @carlosfrodriguez 

I use PhantomJS for all ~Selenium test. This test works for me, I have just checked it again.
I checked it in more VMs, and all is OK. If you found what is the issue with phantom in your environment please let me know, maybe there is some test case that make problem with PhantomJS.

Regards
Zoran

PR #507, opened 2015-03-25
Master port to OM - AdminPGP

s7design:
Hi @mrcbnsls and @mgruner ,

I do here porting to OM for AdminPGP. As you can see in code, I delete checking if PGP is enabled in constructor, and now there is situation that it had been before.
In my opinion it is more user-friendly.

Regards
Zoran


mrcbnsls:
Merged in c0f32ae, thanks.

PR #508, opened 2015-03-26
Master port OM - Dachboard/CIC

s7design:


Kernel/Modules/AgentCustomerInformationCenterSearch.pm
```diff
@@ -24,59 +24,49 @@ sub new {
...
+            'Kernel::System::DB' => {
+                LogObject    => $Param{LogObject},
+                ConfigObject => $Param{ConfigObject},
+                MainObject   => $Param{MainObject},
+                EncodeObject => $Param{EncodeObject},
+                DatabaseDSN  => $ConfigObject->Get('Core::MirrorDB::DSN'),
+                DatabaseUser => $ConfigObject->Get('Core::MirrorDB::User'),
+                DatabasePw   => $ConfigObject->Get('Core::MirrorDB::Password'),
+            },
+        );
```
mgruner: I think this will not have any effect as Kernel::System::DB is already instantiated. You probably have to create an instance directly with DB->new() instead.

Kernel/Modules/AgentCustomerInformationCenterSearch.pm
```diff
@@ -24,59 +24,29 @@ sub new {
...
-
-        if ( $Self->{SlaveDBObject} ) {
-
-            $Self->{SlaveTicketObject} = Kernel::System::Ticket->new(
-                %Param,
-                DBObject => $Self->{SlaveDBObject},
-            );
-        }
-    }
-
```
mrcbnsls: Where has this part gone?

s7design: Hi @mrcbnsls,

I omitted this part of code, I  think that it is not used in the rest of code.  
$Self->{SlaveTicketObject} is defined in constructor, but I cannot see where it is used. 
maybe I made mistake. We can change this if you think that it is necessary.

Regards
Zoran

s7design:
Hi @mrcbnsls,

I have looked into these modules again. As I mentioned in the previous comment, in my opinion 
* $Self->{SlaveDBObject} and
* $Self->{SlaveTicketObject} 
are not needed in AgentCustomerInformationCenterSearch.pm.

It is needed in AgentDashboardCommon, and I think there was a copy/paste problem when AgentCustomerInformationCenterSearch was created. We searched in this module and these object are not used.

We ported SlaveDBObject and SlaveTicketObject objects to OM. We use default objects if 'Core::MirrorDB::DSN' is not defined. If these config items there are, real slave objects will be defined.  

Your opinion will be welcomed, because maybe we have done something wrong here.

Regards
Zoran

mgruner:
This looks good in general. However, the slave connection will not work in the dashlets because they also use the OM to get the ticket object.
I need to clarify how we will proceed here and will tell you as soon as I know.

mgruner:
Hi @s7design, please update this PR. I changed the handling of the SlaveDB in OTRS4 and 5. Now it is much easier to use, please see the current state in the master branch.

s7design:
Hi @mgruner 

I followed your changes with SlaveDB previous days, I will do as soon as possible.

Regards
Zoran

s7design:
Hi @mgruner ,

I fixed SlaveDBObject as you sugested. I hope it is well now. I rebased PR to master.
Please let me know if I could improve or change something.

Regards
Zoran

Kernel/Modules/AgentDashboardCommon.pm
```diff
@@ -756,13 +756,14 @@ sub _Element {
...
+
     # load backends
     my $Module = $Configs->{$Name}->{Module};
-    return if !$Self->{MainObject}->Require($Module);
+    return if !$Kernel::OM->Get('Kernel::System::Main')->Require($Module);
     my $Object = $Module->new(
         %{$Self},
-        DBObject              => $Self->{SlaveDBObject},
-        TicketObject          => $Self->{SlaveTicketObject},
         Config                => $Configs->{$Name},
```
s7design: HI @mgruner ,
I am not sure is this well, because I deleted $Self->{SlaveDBObject} and $Self->{SlaveTicketObject}.
I think if there are in Self it is enough  %{$Self}, but as always your follow up is welcome.

regards
Zoran

scripts/test/Selenium/Agent/AgentCustomerInformationCenter.t
```diff
@@ -0,0 +1,88 @@
...
+        );
+
+        # create test customer
+        my $TestCustomerUserLogin = $Helper->TestCustomerUserCreate(
+            Groups => ['admin'],
+        ) || die "Did not get test user";
+
+        my $ScriptAlias = $Kernel::OM->Get('Kernel::Config')->Get('ScriptAlias');
+        $Selenium->get("${ScriptAlias}index.pl?Action=AgentCustomerInformationCenter");
+
```
mgruner: I had to add a sleep 1; here as it makes an ajax call.

mgruner:
Merged, thank you very much!
BTW, do you also test the stuff with selenium/firefox as well? I recommend that as it has slightly different results than PhantomJS due to timing issues.

s7design:
I started to use the both browser in selenium tests. We used to test only in phantomjs before.
I saw that there are some differences. Generally phantomjs is much better and faster, but the Firefox is very nice because I can look in browser during the execution of test.

From now, we use double browser for testing, it will be better for discovering more potential bugs. 

Regards
Zoran

PR #509, opened 2015-03-26
[Master Bug - 11156 ] - No translated headers in Excel output in AgentTicketSearch 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11156

Hi @mgruner ,

In the bug report, reporter mentioned problem in exporting to Excel with "Artictle Tree" and 
"Customer Realname". 
Actualy, "ArtictleTree" is translated, and here I wanted to fix only translation for "Customer Realname". I did not call translate update comand.

I am not sure is it the best solution, let me know if I can do on different way.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Standard/AAATicket.tt
+++ b/Kernel/Output/HTML/Standard/AAATicket.tt
@@ -169,6 +169,7 @@
 [% Translate("Phone-Ticket") | html %]
 [% Translate("Search Tickets") | html %]
 [% Translate("Customer") | html %]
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @s7design 

For me it looks OK, this pull request was merged in https://github.com/OTRS/otrs/commit/db0fa1ffca4552de23ef653c5d9af955596af17e.

The fix was tested with Spanish language, both strings are correctly translated now.

((enjoy))
cr

PR #514, opened 2015-03-31
TicketSearch by array reference of TicketIDs

moritz:
Most other search options for IDs already accept array refs, so it is only
consistent to accept array refs for TicketID too. Plus, we have a good use
case at my employer for that (filtering a pre-existing list of tickets).

### DIFF
--- a/Kernel/System/Ticket/TicketSearch.pm
+++ b/Kernel/System/Ticket/TicketSearch.pm
@@ -520,7 +520,7 @@ sub TicketSearch {
     if ( $Param{TicketID} ) {
         $SQLExt .= $Self->_InConditionGet(
             TableColumn => 'st.id',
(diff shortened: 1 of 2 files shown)


carlosfrodriguez:
Hi @moritz,

I see a potencial issue if the TicketIDs param is a hash ref instead of an array ref, could it be possible to prevent this with a check?

((enjoy))
cr

moritz:
hi @carlosfrodriguez,

I've added the check for the reference type, as requested.

Cheers,
Moritz

carlosfrodriguez:
Hi @moritz,

Thank you for your pull request, it was merged in https://github.com/OTRS/otrs/commit/0a43815af670d3ebaab91f8ad00ff1f2efcd834a, with very small style changes and a typo fix ;-)

((enjoy))
cr

PR #517, opened 2015-04-01
Master port to OM - AgentTicketBulk

s7design:
Hi @mrcbnsls and/or @mgruner,

You can see in our ported to OM AgentTicketBulk.pm that we omitted:
* $Self->{NoteFormID} and
* $Self->{EmailFormID} 

We saw that these variable are not used in the code. Because of that we omitted hidden fields for that from AgentTicketBulk.tt 
We think that it is good, but because we aren't sure we ask you for meaning.

Regards
Zoran


mgruner:
Updated to latest master and merged in 28bb8ae, thanks!

PR #520, opened 2015-04-02
Spelling check: false string is not an error

moritz:
Spell-checking a text area with an empty string or the string "0"
should not be considered an error worthy of a log file entry.


carlosfrodriguez:
Hi @moritz,

Thank you for your pull request, it was merged in https://github.com/OTRS/otrs/commit/1671d4df6bd0369a2254fb372f1c43fc86803ed1 with small changes.

((enjoy))
cr

PR #521, opened 2015-04-02
Master port to OM - AgentTicketZoom

s7design:
Hi @mrcbnsls and @mgruner 

As you can see there are porting to OM for AgentTicketZoom module and Selenium test for it.
I would like to ask you about constructor for AgentTicketZoom, there are defined many Self parameters, I left almost all in constructor. I left them because they are used in many functions, or for some of them I not sure how to test it. I am not sure is it the best solution, because of that if you have any suggestion about it please let me know and we will be able to change it.

Regards
Zoran


mgruner:
Merged, very nice!

The $Self parameters can stay, removing them would cause too many code changes.

PR #523, opened 2015-04-03
[Rel-3_3 Bug 11080] Ticketfilter doesn' t work - maximum number of expressions reach

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11080

Hi,
In this PR is maybe fixed the bug (11080). I had a problem with reproducing and configuring environment for Oracle DB. While I was comparing code in the newer versions of OTRS, I saw that this issue is solved in TicketSearch(). Because of that I made this PR, maybe it will be helpful, but QA and follow-up is necessary and welcomed.
I am looking forward for any feedback.

Regards
Zoran


s7design:
I close this PR, it is solved in newer version of OTRS

Regards
Zoran

PR #524, opened 2015-04-06
Master port to OM - AgentTicketCustomer

s7design:


mrcbnsls:
Merged in b15150f, thx. Please note that I also added an active sleep before the ticket search here to make sure the ticket has actually been created correctly.

PR #527, opened 2015-04-06
Master port to OM - AgentTicketForward

s7design:


s7design:
Hello @mgruner,

While porting AgentTicketForward.pm we left in constructor $Self->{GetParam}. This is called multiple time through out whole module, so we were are not certain how should we handle it. Specially since it carry many different information that are called separately in the rest of module. 

Please if you could provide us with any feedback regarding that part, and we will implement changes right away.

Regards,
S7

mrcbnsls:
Merged in 3a713ac, thx. GetParam in new() is fine :)

PR #531, opened 2015-04-08
Master port to OM - AgentTicketWatch

s7design:
Hi,

Maybe we need a little help here.
We left  $Self->{JSONObject} = $Kernel::OM->Get('Kernel::System::JSON'); in constructor. There is a problem in TicketOverviewMedium.pm and  TicketOverviewPreview.pm because there is used  
$Self->{JSONObject}.

There are several solution for this:
* to merge this as we do here
* to port $Self->{JSONObject} to OM  in  TicketOverviewMedium.pm and  TicketOverviewPreview.pm and TicketOverviewSmall.pm and the rest of porting we will be able to port when we do these modules.
* and maybe the best solution; to port all modules (TicketOverviewMedium.pm and  TicketOverviewPreview.pm and TicketOverviewSmall.pm ) to  OM

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketWatchView.pm
+++ b/Kernel/Modules/AgentTicketWatchView.pm
@@ -12,8 +12,8 @@ package Kernel::Modules::AgentTicketWatchView;
 use strict;
 use warnings;
 
-use Kernel::System::JSON;
-use Kernel::System::DynamicField;
+our $ObjectManagerDisabled = 1;
+
 use Kernel::System::VariableCheck qw(:all);
 
 sub new {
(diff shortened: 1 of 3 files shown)


s7design:
Hi,

I tried on local to port $Self->{JSONObject} to OM in TicketOverviewMedium.pm and TicketOverviewPreview.pm and TicketOverviewSmall.pm. And it works fine for me.

If you think that it is OK solution I will be able to add this commit here or make another PR .

Regards
Zoran

mrcbnsls:
Merged in e98ed35, thx. We decided that - for now - we only change the way the JSON object is retrieved in TicketOverview(Medium|Preview) to OM as you suggested, which works perfectly fine.

PR #532, opened 2015-04-08
Master port to OM -  AgentTicketViews

s7design:
There is the same situation with JSONObject like in https://github.com/OTRS/otrs/pull/531


mrcbnsls:
Merged in fe5cb06, thx! Had to fix some things in the escalation view test, but great work after all!

PR #533, opened 2015-04-09
Fix bug#11226 - Update Base.pm

Fess-AKA-DeadMonk:
1. why use __|__ for case in __?:__ instead of just __?i:__ ?
2. Without following changes the tag is replaced only if there is nothing but it in the field.
 1. changed regexp (deleted __\A__ and __\z__), so it would work also if there something except a tag in the field. Without this change the tag is replaced only if there is nothing but it in the field.
 2. replaced the __if__ statement with a __while__ cycle, so multiple tags may be processed
    1. added __g__ key in __s///__ in case some tags are repeating multiple times.

* http://forums.otterhub.org/viewtopic.php?f=62&t=22916&p=116147#p116147 - related forum thread
* http://otrs.ru/forum/viewtopic.php?f=5&t=3042&p=15896#p15896 - related russian forum thread
* http://bugs.otrs.org/show__bug.cgi?id=11226 - related bug

### DIFF
--- a/Kernel/System/ProcessManagement/TransitionAction/Base.pm
+++ b/Kernel/System/ProcessManagement/TransitionAction/Base.pm
@@ -80,13 +80,14 @@ sub _ReplaceTicketAttributes {
         # replace ticket attributes such as <OTRS_Ticket_Dynamic_Field_Name1> or
         # <OTRS_TICKET_Dynamic_Field_Name1>
         # <OTRS_Ticket_*> is deprecated and should be removed in further versions of OTRS
-        while (
-            $Param{Config}->{$Attribute}
-            && $Param{Config}->{$Attribute} =~ m{<OTRS_(?i:TICKET)_([A-Za-z0-9_]+)>}msx
-            )
+        for ( my $replacement = 0;
+            $replacement <= 1000 
+            && $Param{Config}->{$Attribute}
+            && $Param{Config}->{$Attribute} =~ m{<OTRS_TICKET_([A-Za-z0-9_]+)>}misx;
+            $replacement++ )
         {
             my $TicketAttribute = $1;
-            $Param{Config}->{$Attribute} =~ s/<OTRS_(?i:TICKET)_$1>/$Param{Ticket}->{$TicketAttribute}/msxg // '';
+            $Param{Config}->{$Attribute} =~ s/<OTRS_TICKET_$1>/$Param{Ticket}->{$TicketAttribute}/misxg // '';
         }
     }
 
--- a/Kernel/System/ProcessManagement/TransitionAction/Base.pm
+++ b/Kernel/System/ProcessManagement/TransitionAction/Base.pm
@@ -80,13 +80,13 @@ sub _ReplaceTicketAttributes {
         # replace ticket attributes such as <OTRS_Ticket_Dynamic_Field_Name1> or
         # <OTRS_TICKET_Dynamic_Field_Name1>
         # <OTRS_Ticket_*> is deprecated and should be removed in further versions of OTRS
-        if (
+        while (
             $Param{Config}->{$Attribute}
-            && $Param{Config}->{$Attribute} =~ m{\A<OTRS_(?:Ticket|TICKET)_([A-Za-z0-9_]+)>\z}msx
+            && $Param{Config}->{$Attribute} =~ m{<OTRS_(?i:TICKET)_([A-Za-z0-9_]+)>}msx
             )
         {
             my $TicketAttribute = $1;
-            $Param{Config}->{$Attribute} = $Param{Ticket}->{$TicketAttribute} //= '';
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @Fess-AKA-DeadMonk "//=" is the defined equals operator, 
```perl
$Param{Config}->{$Attribute} = $Param{Ticket}->{$TicketAttribute} //= ''
```
In this case it looks wrong it should be:
```perl
$Param{Config}->{$Attribute} = $Param{Ticket}->{$TicketAttribute} // ''
```
Which means that the $Param{Config}->{$Attribute} = $Param{Ticket}->{$TicketAttribute} if $Param{Ticket}->{$TicketAttribute} is defined, otherwise it takes an empty value.

carlosfrodriguez:
Hi @Fess-AKA-DeadMonk,

Tank you for your pull request, in order to continue with the merging it is necessary to get a signed copy of the CLA as stated in http://otrs.github.io.

Can you please send us your signed CLA?

((enjoy))
cr

Kernel/System/ProcessManagement/TransitionAction/Base.pm
```diff
@@ -80,13 +80,13 @@ sub _ReplaceTicketAttributes {
         # replace ticket attributes such as <OTRS_Ticket_Dynamic_Field_Name1> or
         # <OTRS_TICKET_Dynamic_Field_Name1>
         # <OTRS_Ticket_*> is deprecated and should be removed in further versions of OTRS
-        if (
+        while (
```
mgruner: I think some kind of loop protection would be nice (not more than 1000 replacements, for example).

Kernel/System/ProcessManagement/TransitionAction/Base.pm
```diff
@@ -80,13 +80,13 @@ sub _ReplaceTicketAttributes {
...
             $Param{Config}->{$Attribute}
-            && $Param{Config}->{$Attribute} =~ m{\A<OTRS_(?:Ticket|TICKET)_([A-Za-z0-9_]+)>\z}msx
+            && $Param{Config}->{$Attribute} =~ m{<OTRS_(?i:TICKET)_([A-Za-z0-9_]+)>}msx
```
mgruner: Here I think adding the i modifier to the regex at the end would be better than just in the parens. Same below.

Fess-AKA-DeadMonk: so the tag would be completely case insensitive? i like it, because don't see any use of case sensivity in tags at all.
but i was not completely sure, so changed just a _Ticket_ part

Kernel/System/ProcessManagement/TransitionAction/Base.pm
Fess-AKA-DeadMonk: do not see the use of it. i don't see the vulnerabilities in this code, so it would need over 1000 _different_ tags to make more than 1000 iterations.
but i never developed a project of such scale, so would we leave it like this?
```perl
for (my $replacement = 0; $replacement <= 1000 && $Param{Config}->{$Attribute}
            && $Param{Config}->{$Attribute} =~ m{<OTRS_(?i:TICKET)_([A-Za-z0-9_]+)>}msx; $replacement++ )
```

Fess-AKA-DeadMonk:
@carlosfrodriguez, i've sent signed CLA to cla@otrs.com, is everything fine now?

carlosfrodriguez:
Hi @Fess-AKA-DeadMonk,

Thank you for your CLA, we are processing it.

((enjoy))
cr

carlosfrodriguez:
Hi @Fess-AKA-DeadMonk,

You CLA was processed successfully, thank you so much for sending it to us.

((enjoy))
cr

carlosfrodriguez:
Hi @Fess-AKA-DeadMonk,

Your pull request was merged in https://github.com/OTRS/otrs/commit/54812717dea4a327130d366503eda26873d4b0f9 with some modifications, please stay in touch if we need to discuss anything related to this fix.

((enjoy))
cr

Fess-AKA-DeadMonk:
Great!
But if you have some time, could you explain, why don't use my __for__ loop from[3627dc6](https://github.com/Fess-AKA-DeadMonk/otrs/commit/3627dc6fa24862d69d278ab6a20538004000bd08)?
is there any reason?
or maybe those changes was not visible to you, and i had to make a new pull request or something?

carlosfrodriguez:
Hi @Fess-AKA-DeadMonk,

I saw the change but I revert it back to a while (as in your initial fix), basically because of the code style in OTRS

1 We normally do not write 'for' statements in C/C++ style
2 I feel it much more complex to understand (provably less explicit than a 'last')
3 It requires more lines ;-)

I hope that the solution that I implement is still working for you, please let me know otherwise, you can always re-open the bug if something is wrong.

((enjoy))
cr

PR #535, opened 2015-04-14
SASL authentication for LDAP

lameventanas:
This patch implements (optional) SASL authentication in all LDAP binds:
I tested it with GSSAPI (Kerberos) only, but I have no reason to think it won't work with the other mechanisms.
I also added new optional dependencies in otrs.CheckModules.pl:
- Authen::SASL
- GSSAPI

There is also a typo fix in the language list in preferences, for Japanese.

### DIFF
--- a/Kernel/Config/Defaults.pm
+++ b/Kernel/Config/Defaults.pm
@@ -174,7 +174,7 @@ sub LoadDefaults {
         'hr' => 'Hrvatski',
         'hu' => 'Magyar',
         'it' => 'Italiano',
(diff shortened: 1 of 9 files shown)


mgruner:
Hello @lameventanas! Thank you very much for your contribution. Can I ask you to please sign our CLA as described on http://otrs.github.io? This will be required to merge your changes.
I already applied the ";" fix. I see two other features in your PR, SASL and different image handling. Can you please split them up in two separate PRs?
Regards, mg

PR #536, opened 2015-04-14
[Master - Bug 10961]   Adding email recipients via addressbook does not update customer information

s7design:

Hi @zottto 

There is PR for master for bug with AgentBook. 

Regards
Zoran


zottto:
Merged with 2bac74b.

PR #540, opened 2015-04-16
Customer logout button not translating

Kjellarne6:
Fixed the logout button in the customer portal not translating to other languages.


mgruner:
Hi @Kjellarne6. Thank you very much for your submission. The real issue was a problem with out translation script that I fixed in https://github.com/OTRS/otrs/commit/709c0f450f540b9cf26fb764a6e7f784eee06c30. Now the correct string is available in Transifex and can be translated. Thanks!

PR #543, opened 2015-04-17
Customer's photo in ticket layout.

lameventanas:
I would totally understand if this is not merged because I don't consider this a nice solution, but it does the job for me right now.

The important thing to note is that many people are interested in showing the customer's photo together with the rest of their information. It makes OTRS look much more professional.

I use this with the thumbnailPhoto attribute from LDAP.

Please see this for some background:
http://forums.otterhub.org/viewtopic.php?t=22217#p116410


mgruner:
Hello @lameventanas, thank you very much for your suggestion. Interesting solution! At this point I would not want to merge it. Please also see #662 for a different approach on image handling in OTRS (not sure yet how we will proceed with this one though). Best regards, Martin

PR #544, opened 2015-04-20
Don't display 0 accounted time in articles (IB#1040541)

pboguslawski:
https://dev.ib.pl/ib/otrs-ib/issues/13

Please merge to master if you find it useful (we still use 3.3 for devel now and will try to be more compatible in the future).

Other ideas:
* adding extra parameter in sysconfig might be better for hiding 0 accounted times in articles and in tickets maybe,
* consider adding indexes on article.create_time and time_accounting.create_time columns in db schema to allow optimal searching (in custom queries/stats for example) for accounting data in given period and for articles without accounting time defined by mistake). 

Regards,
Pawel


mgruner:
Merged to master in 1a17d1f, thanks a lot!

PR #546, opened 2015-04-21
AgentTicketBulk: Redirect to Zoom if only one Ticket comes out

moritz:
if the bulk action involves only one ticket, or one of the action is to
merge all tickets into one, redirect to the AgentTicketZoom view of
that one ticket after successfully doing the bulk action.

This makes for a much smoother user experience, since usually people
want to actually work with the ticket that they just merged.

### DIFF
--- a/Kernel/Modules/AgentTicketBulk.pm
+++ b/Kernel/Modules/AgentTicketBulk.pm
@@ -317,6 +317,7 @@ sub Run {
 
     # process tickets
     my @TicketIDSelected;
+    my @ResultTicketIDs;
     my $LockedTickets = '';
     my $ActionFlag    = 0;
     my $Counter       = 1;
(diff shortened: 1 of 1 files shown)


mgruner:
Can you rebase this please, @moritz?

moritz:
@mgruner it's now rebased on latest master.

Kernel/Modules/AgentTicketBulk.pm
```diff
@@ -317,6 +317,7 @@ sub Run {
 
     # process tickets
     my @TicketIDSelected;
+    my @ResultTicketIDs;
```
mgruner: Actually, would it not be possible now to just use the MainTicketID instead that you introduced in the other PR? Or is it not enough?

moritz: Depends on what you want to happen if a bulk action works on only one ticket (for example only one was locked); with my patch, it redirects to AgentTicketZoom; when I just use MainTicketID, it'd go back to the previous screen.

I prefer it my way, though I'm not too attached to it.

moritz: @mgruner After thinking a bit about it, it's probably more intuitive to only redirect after a merge; so I went with your suggestion in 5e35c7c18017f276e345b243bfec4c503b81a831.

mgruner:
Merged in 17f8222, thank you very much!

PR #547, opened 2015-04-21
Improve performance of AgentTicketBulk

moritz:
when the "merge to oldest" option is selected, the old code determined
the oldest ticket for each ticket anew, thus causing quadratic runtime.

Instead, precompute the target ticket ID outside the loop.


mgruner:
Hi @moritz. I merged this in 720c59d with a slight addition: if you input something in the MergeTo field, the option element will be automatically selected. Please let me know if this works for you. Thanks a lot! :+1:

moritz:
Works perfectly fine for me, thanks :-)

PR #549, opened 2015-04-21
Port to OM - Output/Html/Preferences

s7design:
Hi @mgruner 

We stated porting to OM for Output modules. Your follow-up is welcomed. There is also one test I will add the rest two as well. If you have any suggestion let me know.

Regards
Zoran


mgruner:
Hi Zoran,
thank you. I merged this in 1fc163c. Please note that I changed the file names to ServicePreferences/Generic.pm. The duplication is not really needed. Please make it like this also with the other portings.
Regards, mg

PR #550, opened 2015-04-21
Port to OM - AgentTicketProcess

s7design:


mrcbnsls:
The UT is unfortunately failing for me. 

``#57 - Exception in Selenium': An element could not be located on the page using the given search parameters: #PriorityID option[value='5'],css selector at (eval 99) line 135.``

Can you please have a look?

s7design:
Hi @mrcbnsls 

Selenium test is updated, added active sleep just before completing process scenario in line 156 and on other places where it could be necessary. Test is also tested on different machines and working as expected.

Please let me know if this fix solved crash for you.

Kind regards,
S7

mrcbnsls:
Merged in 0346884, thx. Had to change only a small thing to avoid an issue where you would refresh the main window while the popup was still open in the background. Refresh should not be necessary here, because submitting the popup will automatically cause the parent window to reload. So I now just check for the presence of a specific element on the main window with active sleep. As soon as it is there, we know that the main window has been reloaded :+1:

PR #556, opened 2015-04-28
Master selenium test - AgentTicketAttachment

s7design:
Hi,

While we were doing on Selenium test for some modules in Kernel/Output, we saw that selenium test for  AgentTicketAttachment is wrong, because of that we make this PR.

Regards
Zoran


mgruner:
Merged, thanks.

PR #557, opened 2015-04-28
Master port to OM  - Kernel/Output/Article

s7design:
Hi,

There are ported Article modules from Kernel/Output/HTML.
We added only one Selenium test, to be honest I don't know how to test these modules for SMIME/PGP/Sign/Crypt with Selenium, and I think that test for ArticleAttachmentDownload is not needed here, because there is test for AgentTicketAttachment. Here is test for ArticleAttachmentHTMLViewer. There is checked how  MIME-Viwer works, if it is enabled. 

In some modules is used EmailParser.pm (ArticleCheckPGP and ArticleCheckSMIME), and I am not sure if we used it well. We had such dilemma only in this PR https://github.com/OTRS/otrs/pull/521 (line 3053 - AgentTicketZoom.pm). I saw that in  Kernel::System::EmailParser there is  $ObjectManagerDisabled = 1;, maybe because of that we cannot use it with OM.
Also, I delete %{$Self}, from constructor for EmailParser. I think it is not necessary anymore.

And one more thing, I saw that in some modules is used for checking needed prams the flowing
    # check needed stuff
     for (qw(File Article)) { 
... and then in body of for loop is used $_

and in the others also is used 
    # check needed stuff
    for my $Needed (qw(File Article)) {

In my opinion the second one is better and I feel free to fix it in these ported modules. 
As I said it is my opinion, and your follow up is welcome in any of these question and dilemmas.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
+++ b/scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
@@ -31,211 +31,138 @@ $Selenium->RunTest(
         );
         my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
 
-        # get sysconfig object
-        my $SysConfigObject = $Kernel::OM->Get('Kernel::System::SysConfig');
-
-        # do not check RichText
-        $SysConfigObject->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'Frontend::RichText',
-            Value => 0
-        );
-
-        # do not check service and type
-        $SysConfigObject->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'Ticket::Service',
-            Value => 0
-        );
-        $SysConfigObject->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'Ticket::Type',
-            Value => 0
-        );
-
-        $SysConfigObject->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'PDF',
-            Value => 0
-        );
-
-        # enable MIME-Viewer for PDF attachment
-        $SysConfigObject->ConfigItemUpdate(
-            Valid => 1,
-            Key   => 'MIME-Viewer###application/pdf',
-            Value => "pdftohtml -stdout -i",
-        );
-
-        # create test user and login
-        my $TestUserLogin = $Helper->TestUserCreate(
-            Groups => [ 'admin', 'users' ],
-        ) || die "Did not get test user";
-
-        $Selenium->Login(
-            Type     => 'Agent',
-            User     => $TestUserLogin,
-            Password => $TestUserLogin,
-        );
-
-        # get test user ID
-        my $TestUserID = $Kernel::OM->Get('Kernel::System::User')->UserLookup(
-            UserLogin => $TestUserLogin,
-        );
-
-        # create test company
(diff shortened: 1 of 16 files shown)


scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
```diff
@@ -0,0 +1,242 @@
...
+            UserEmail      => "$TestCustomer\@localhost.com",
+            ValidID        => 1,
+            UserID         => $TestUserID,
+        );
+
+        # create test phone ticket with PDF attachment
+        my $ScriptAlias = $Kernel::OM->Get('Kernel::Config')->Get('ScriptAlias');
+        $Selenium->get("${ScriptAlias}index.pl?Action=AgentTicketPhone");
+
+        # create test phone ticket
```
mgruner: In general, here in this test we don't want to test the ticket creation, so it should not be done with Selenium. Create the ticket in the backend, this is much easier. You can skip the code from here...

scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
```diff
@@ -0,0 +1,242 @@
...
+
+        # search for new created ticket on AgentTicketZoom screen
+        my %TicketIDs = $TicketObject->TicketSearch(
+            Result         => 'HASH',
+            Limit          => 1,
+            CustomerUserID => $TestCustomer,
+            UserID         => $TestUserID,
+        );
+        my $TicketNumber = (%TicketIDs)[1];
+        my $TicketID     = (%TicketIDs)[0];
```
mgruner: ... until here.

scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
```diff
@@ -0,0 +1,242 @@
...
+            sleep 1;
+        }
+
+        $Self->True(
+            index( $Selenium->get_page_source(), $TicketNumber ) > -1,
+            "Ticket with ticket id $TicketID is created"
+        );
+
+        # go to ticket zoom page of created test ticket
+        $Selenium->find_element("//a[contains(\@href, \'Action=AgentTicketZoom' )]")->click();
```
mgruner: ... even actually until here. Then you can call up TicketZoom directly.

scripts/test/Selenium/Output/ArticleAttachmentHTMLViewer.t
```diff
@@ -0,0 +1,242 @@
...
+            Valid => 1,
+            Key   => 'PDF',
+            Value => 0
+        );
+
+        # enable MIME-Viewer for PDF attachment
+        $SysConfigObject->ConfigItemUpdate(
+            Valid => 1,
+            Key   => 'MIME-Viewer###application/pdf',
+            Value => "pdftohtml -stdout -i",
```
mgruner: Does the test require this program to be installed?

s7design:
Hi @mgruner,

Selenium test is now updated to check if pdftohtml is installed on start of the test. 
Ticket, article and attachments are added through backend, which results in cleaner and faster test.

Regards,
Sanjin

mgruner:
Merged, thank you very much! I believe you forgot to remove some old files which I did now in the merge commit.

PR #558, opened 2015-04-29
Master port to OM - Kernel/Output/Preferences

s7design:
Hi,

These are ported Preferences modules from Kernel/Output/HTML. Some tests are excluded from PR, Theme, PGP, SMIME and ColumnFilters preferences. For ColumnFilters, when enabled from SysConfig, doesn't work right off the bat, since it missing Data to select filters from. For Theme, couldn't find a way to do it without adding 3rd party theme and loading it up.

If you have any idea how to set remaining ones properly please let us know and we will create test for it as well.

Regards,
S7


s7design:
Hi @mgruner,

I saw some problem here while I tested customer preferences with these porting modules. We will fix it during the day, and then we will inform you that you can follow-up it. 
I write this only in case you don't waste your time to follow-up completed PR. But if you have any idea about question that Sanjin mentioned in the previous comment it will be welcome.

Regards
Zoran

s7design:
Hi,

PR is now complete. Added selenium tests for customer preferences. It is tested internally and they perform as expected.
Additionally for modules that are both for agent and customer preference, $Self->{UserObject} is left as it is since it's called for both preferences. Also added in constructor to check for UserObject as a safe trigger. Looking forward to your follow-up on this. In case it is necessary to change and write separate calls depending on which preference is called, please let me know.
Lastly only missing tests at this moment are for agent ColumnFilter and Theme preference. Any input on these two is welcome.

Regards,
Sanjin

mgruner:
Merged, thank you very much! I will add some inline comments.

scripts/test/Selenium/Output/Preferences/Customer/SMIME.t
```diff
@@ -0,0 +1,105 @@
...
+        my $ScriptAlias = $ConfigObject->Get('ScriptAlias');
+
+        # go to customer preferences
+        $Selenium->get("${ScriptAlias}customer.pl?Action=CustomerPreferences");
+
+        # change customer SMIME certificate preference
+        my $Location = $ConfigObject->Get('Home')
+            . "/scripts/test/sample/SMIME/SMIMECertificate-1.asc";
+        $Selenium->find_element( "#UserSMIMEKey", 'css' )->send_keys($Location);
+        $Selenium->find_element(".//*/div/div[5]/form/fieldset/button")->click();
```
mgruner: This is not good as you reference the div only by its position. I replaced this with 
`$Selenium->execute_script("\$('#UserSMIMEKey').parents('form').submit();");`. Can you please adapt it for the other tests too where the positional indexing was used?

scripts/test/Selenium/Output/Preferences/Customer/TimeZone.t
```diff
@@ -0,0 +1,80 @@
...
+# --
+
+use strict;
+use warnings;
+use utf8;
+
+use vars (qw($Self));
+
+# get selenium object
+$Kernel::OM->ObjectParamAdd(
```
mgruner: Is this really needed? If not I recommend that you remove it from all tests.

scripts/test/Selenium/Output/Preferences/Customer/SMIME.t
s7design: Thanks @mgruner it is very nice, I was looking for something like this.  It will be helpful in some selenium test where we used only complex xpath. We will fix it for the others tests.
Zoran

PR #559, opened 2015-04-30
Master - fixed issue with Dynamic Fields in AgentTicketActionCommon

s7design:


s7design:
Hi @mgruner 

I hope that I fixed this issue. There was our mistake in porting to OM.
I tested AgentNotification but it works for me. It would be nice if I had step by step test case in witch there is a problem with send notification. 
I used the following one:
* Log in as User1
* Create New ticket
* Change Owner to User2
* Look in History: Notification QwnerUpdate is sent

I tested this with otrs branch before marge for AgentTicketActionCommon ( https://github.com/OTRS/otrs/commit/0d171dddbd71f58288f84de7f53398aa78ca4fe1 ) and after that as well. I have the same behavior in the both state of OTRS.

We are here to solve if there is something else, please let me know if it is ok now. 

Regards
Zoran

mgruner:
Merged, thanks! I'll ask the colleague who reported this to also test again.

PR #560, opened 2015-04-30
selenium tests

s7design:
There are two test for Preferences SLA/Service that I have owed since earlier.
Regards
Zoran


mgruner:
Merged, thanks!

PR #561, opened 2015-04-30
Improvment in Selenium testing

s7design:
Hi,

While we tested some tests, we saw that there is a issue if it is used Firefox.
In command setWindowSize, there is the following message
<pre><code>
[Thu Apr 30 15:52:12 2015] otrs.Console.pl: Use of uninitialized value $ret in pattern match (m//) at /opt/otrs/Kernel/cpan-lib/Selenium/Remote/Driver.pm line 1303.
</pre></code>

It seems that there is a bug, and I felt free to fix it. Maybe there is better solution if I can solve this on another way please let me now.

P.S. You can check it with AdminType.t but only with Firefox the bug occurs.

Regards
Zoran


mgruner:
Merged, thanks a lot! Usually we try to avoid patching external modules, but in this case I'll make an exception as we use an outdated version of Selenium::Remote::Driver, but I want to avoid an update at this point due to the external requirements the module has.

PR #562, opened 2015-04-30
Selenium tests improvements

s7design:


s7design:
Hi,

We run all Selenium tests and found out few that didn't work completely. 
Added active sleep for AdminPGP.t and AdminSystemAddress.t which should solve issue running tests in Firefox.
In Admin.t tests if S/MIME is not enabled test was not prepared for it. Added workaround for both scenarios now.

Regards,
Sanjin

mgruner:
Merged, thanks a lot! I removed the special handling in Admin.t where you produce a successful test for AdminSMIME when the object cannot be loaded. IMHO there should be a failure -- or maybe I misunderstood your intention.

s7design:
Hi @mgruner ,

I added checking if  SMIME is enabled. I used a little unusual way for this, because in constructor for SMIME and PGP there are checks if it is possible to be created these modules.  Maybe it is not the best solution. 
I used return value of creating an instance of  SMIME object. If there is '0' some sysconfig is not set, and  there will be error message on AdminSMIME screen. There is needed different test case because there is different behavior from the others Admin screens.
Without part of test that you removed there will be error if run Admin.t in case SMIME::CertPath and SMIME::PrivatePath files are not writable.

I have spoken to Marc before that I am planing to change AdminSMIME.pm, and it will be the best if we have the same behavior as well in AdminPGP (if PGP is not enabled, there is hint, and link to PGP SysConfig).  

As I mentioned I can change AdminSMIME.pm, and then this check in Selenium test will not be necessary. 

Please, let me know what do you think about this.

Regards
Zoran

PR #563, opened 2015-05-03
Deny customer access to tickets in some queues

albthe:
Hello,

in our company we would like to hide some tickets from the customer.
The best solution seems to be hiding the tickets in some queues.
This way the agent will be able to hide or unhide the tickets simply by moving them from one queue to another one.
The queue names for hiding should be changeable via sysconfig.
This pull requeust contains the needed changes.

Could you please check if it makes sense to include this function and could you please review the code changes.

Thanks.
Albrecht

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -1714,6 +1714,15 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::Frontend::CustomerDenyQueuesTicketAccess" Required="0" Valid="1">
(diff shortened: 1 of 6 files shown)


mgruner:
Hello @albthe! Thank you very much for your proposal.

Some technical comments: Such PRs with new features should be done against the master branch, please. Also, please consider signing the CLA as described on https://otrs.github.io.

For the current PR I believe there is a feature in OTRS that does the job already. Please see the setting  CustomerGroupSupport in SysConfig. Here you can define group permissions for your customers so that you can exactly control which queues they have access to, even with a dedicated GUI (AdminCustomerUserGroup).

Therefore I'll close this PR now. Thanks anyway, and feel free to respond if I misunderstood you.

Best regards, mg

PR #564, opened 2015-05-04
Use `CustomerID` param when creating new CustomerUser.

eliasp:
When creating a new user by clicking on `+ Add Customer` in the
CustomerInformationCenter, the param `CustomerID` isn't re-used which
requires the user to fill this field manually which is cumbersome and
error-prone.

In the case of `CustomerCompanySupport` being enabled, the param
`CustomerID` is re-used to set the default selection of the
`CustomerID` drop down.

When `CustomerCompanySupport` is not enabled, the `CustomerID` field
remained empty so far. This patch improves this behaviour and
pre-populates the `CustomerID` field based on the `CustomerID` param
value.

### DIFF
--- a/Kernel/Modules/AdminCustomerUser.pm
+++ b/Kernel/Modules/AdminCustomerUser.pm
@@ -936,6 +936,9 @@ sub _Edit {
                 Class      => $Param{RequiredClass} . ' ' . $Param{Errors}->{ $Entry->[0] . 'Invalid' },
             );
         }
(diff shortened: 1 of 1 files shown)


eliasp:
Probably also worth a backport/cherry-pick to `rel-4_0` as bugfix.

mgruner:
Merged in master in 0074d87 with a small fix, and backported to OTRS 4. With your code the field would not be correctly stored any more after reloading the add mask due to validation errors.

PR #565, opened 2015-05-04
Master selenium xpaths improvement

s7design:
Hi @mgruner 

There are changed Selenium tests and we used execute_script() instead of complicated xpath for finding elements.
I hope that we fixed that for all Selenium tests.

Regards
Zoran


mgruner:
:+1:

PR #568, opened 2015-05-05
Master - Removed unnecessary param Verbose for Selenium objects

s7design:


s7design:
Hi @mgruner,

As you suggested I checked if Verbose pram is needed in Selenium object. You are right, it is not necessary any more. I looked into history of Selenium.pm, and I think that Verbose was needed before, when you use WWW::Selenium driver.
I went trough all selenium tests and changed creating Selenium object. I also ported to OM Helper and Selenium object where it was needed (for tests that  have done before.).

Regard
Zoran

mgruner:
Merged, thank you! I removed all instances where RestoreSystemConfiguration is 0 because it is  the default. I also added a comma for correct indentation of the argument hash. Can you please check if you have a recent Perl::Tidy installed? It seems to indent a little different for you.

s7design:
Hi Martin,

     Yes you are right, we had all version of Perl::Tidy. We have 
updated it now.
Thanks for suggestion.

Regards
Zoran


> Merged, thank you! I removed all instances where 
> RestoreSystemConfiguration is 0 because it is the default. I also 
> added a comma for correct indentation of the argument hash. Can you 
> please check if you have a recent Perl::Tidy installed? It seems to 
> indent a little different for you.
>
> —
> <https://github.com/OTRS/otrs/pull/568#issuecomment-99575693>.
>


-- 
Zoran Ilibašić

S7 Design
Narodnog Fronta 23D/109

w. www.s7designcreative.com
e. zilibasic@s7designcreative.com

PR #569, opened 2015-05-06
Master - Selenium tests replaced go_back function

s7design:
Hi @mgruner,

We replaced go_back function in the Selenium tests with exact URL of screens that we expected next in a test case.
I saw that you did this in AgentTicketMerge, maybe it could be better that we use something that we suggest in this PR because in this way test scripts are more readable, but of course behavior and test case is the same.
If you mean that it is good, you can merge it.

Regards
Zoran


mgruner:
Merged, thanks!

PR #570, opened 2015-05-07
Removed VariableWidth CSS class from Postmaster filter table.

nilei:
The variable postmaster filter table width is more or less unique in the admin interface. All other tables are width 100%, so should the postmaster filter table :-)


mgruner:
Hi Nils, thanks for your suggestion. I tried it, but now the table still only has 2 columns and that looks quite broken. We might consider adding more columns to show at some point, then it could make sense to make the table wider.

PR #571, opened 2015-05-07
Master - Selenium sleep improvement

s7design:
Hi @mgruner 

here is fixed and changed sleep with WaitFor function. We tested all Selenium test again, and there were some unsuccessful test that is cussed with WaitFor, we also fix that.
For example
 $Selenium->WaitFor( JavaScript => 'return $("li.ui-menu-item:visible").length' );
is changed to
 $Selenium->WaitFor( JavaScript => 'return $("li.ui-menu-item:visible")' );

And then all works fine. Please followup this, and tell us what do you think about this.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminCustomerUserGroup.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminCustomerUserGroup.t
@@ -116,7 +116,6 @@ $Selenium->RunTest(
 
         # test Filter for Groups
         $Selenium->find_element( "#FilterGroups", 'css' )->send_keys($GroupRandomID);
-        sleep 1;
 
         $Self->True(
             $Selenium->find_element( "$GroupRandomID", 'link_text' )->is_displayed(),
@@ -180,7 +179,7 @@ $Selenium->RunTest(
             $Kernel::OM->Get('Kernel::System::Cache')->CleanUp( Type => $Cache );
         }
 
-    }
+        }
 
 );
 
--- a/scripts/test/Selenium/Agent/Admin/AdminCustomerUserGroup.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminCustomerUserGroup.t
@@ -179,7 +179,7 @@ $Selenium->RunTest(
             $Kernel::OM->Get('Kernel::System::Cache')->CleanUp( Type => $Cache );
         }
 
-        }
+    }
 
 );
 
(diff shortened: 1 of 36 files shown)


mgruner:
Hello S7, the .length is needed because $() will always return a valid jQuery object and thus be a true value. .length  indicates if it actually contains DOM elements. Please check this, otherwise the WaitFor call seems useless.

s7design:
Hi @mgruner 

I changed test in Selenium/Agent/Admin/ProcessManagement/*, there were the problems with toggling sidebar for elements of process  (Transition, TransitionAction...)
I had to add sleep 1, because I could n0t find way to use WaitFor() in this case. 
<pre>
        # click on ActivityDialog dropdown, wait to load it and "Create New Activity Dialog"
        $Selenium->WaitFor( JavaScript => 'return $("a.PopupType_ActivityDialog:visible").length' );
        $Selenium->find_element( "Activity Dialogs", 'link_text' )->click();
        # wait to toggle element
         sleep 1;
       $Selenium->find_element("//a[contains(\@href, \'Subaction=ActivityDialogNew' )]")->click();
</pre>
And of course the problem was in Firefox again :)
Please follow-up it.

I will test it more also, but your opinion is welcomed.

Regards
Zoran

s7design:
Hi @mgruner 

I hope that I fixed all wait for issues, please follow-up this. I think it works better now.
 If you have any suggestion, let me know.

Regards
Zoran

s7design:
Hi @mgruner 

I have tested all these in this PR in Firefox and PhantomJS. All of them works fine in Firefox now, but some of them is not passed all test steps in PhantomJS. 
As you can see I use AdminDynamicField.t with $Selenium->get_alert_text(), it was before your commit ((
https://github.com/OTRS/otrs/commit/2d0eb7cb4fd336fd49fb03ac8a0933f4d8333184 ) and  it is nice only in Firefox, but if I use AdminDynamicField.t from your comity it works only  in PhantomJS.
Please check this and let me know what I can do with this.

Regards
Zoran

mgruner:
Hello S7, thanks very much for your work on this so far. This PR is getting too big to QA properly. Can you please split this up in several PRs where each has around 5 changed files (does not have to be all at the same time, can be one by one).
Regarding accept_alert, we cannot use this as it is not supported by PhantomJS. For me it worked ok in Firefox with the JS based solution IIRC. What fails for you?
Regards, mg

mgruner:
I will close this PR for now, I hope this is ok.

PR #572, opened 2015-05-08
Master port to OM - Output overview

s7design:
Hi @mgruner.

Here  are ported modules for ticket overview. We made thee separated Selenium test for each view ( Small, Medium and Large). Maybe it was posible put al of them in one, but there are some different and for us it was better to test in separated ones.

It is our proposal a test case, actually there are many way that we could test this modules. In the test are created 15 tickets and then on each view it is  tested a filters and ordering. We use ticket_number for sorting. If you have any suggestion for better test case for these modulles please let us know.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Ticket/OverviewMedium.pm
+++ b/Kernel/Output/HTML/Ticket/OverviewMedium.pm
@@ -6,18 +6,24 @@
 # did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
 # --
 
-package Kernel::Output::HTML::TicketOverviewMedium;
+package Kernel::Output::HTML::Ticket::OverviewMedium;
 
 use strict;
 use warnings;
 
-use Kernel::System::CustomerUser;
-use Kernel::System::DynamicField;
-use Kernel::System::DynamicField::Backend;
 use Kernel::System::VariableCheck qw(:all);
 
 our @ObjectDependencies = (
-    'Kernel::System::JSON',
+    'Kernel::System::CustomerUser',
+    'Kernel::System::DynamicField',
+    'Kernel::System::DynamicField::Backend',
+    'Kernel::Config',
+    'Kernel::System::Log',
+    'Kernel::Output::HTML::Layout',
+    'Kernel::System::User',
+    'Kernel::System::Ticket',
+    'Kernel::System::Main',
+    'Kernel::System::Queue'
 );
 
 sub new {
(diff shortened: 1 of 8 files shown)


Kernel/Output/HTML/Ticket/OverviewMedium.pm
```diff
@@ -27,40 +33,8 @@ sub new {
     my $Self = \%Param;
     bless( $Self, $Type );
 
-    # get needed objects
-    for (
-        qw(ConfigObject LogObject DBObject EncodeObject LayoutObject UserID UserObject GroupObject TicketObject MainObject QueueObject)
```
mgruner: Here everything that is not an object (UserID etc.) should be kept. Even if it is not used directly in the code there could be code still relying on it in submodules etc.
Please also keep this for the other overviews.

Kernel/Output/HTML/Ticket/OverviewMenuSort.pm
```diff
@@ -54,6 +51,17 @@ sub Run {
         return;
     }
 
+    # get layout object
+    my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
+
+    # create language object
+    $Kernel::OM->ObjectParamAdd(
```
mgruner: I believe this is not needed. The LanguageObject is already there because the LayoutObject creates it. You can just run $Kernel::OM->Get() to fetch it, or use $LayoutObject->{LanguageObject} like in the other overviews. (the first option is better IMHO).

mgruner:
Looks nice. Can you please make the small adjustments I suggested? Then I'll merge this one.

Kernel/Output/HTML/Ticket/OverviewMedium.pm
s7design: Hi @mgruner ,

Please if you could clarify a bit on this, since UserID (line 37) is kept and remaining params that are objects are deleted. Not sure if that's what you meant or it's something else.

Regards,
Sanjin
S7

Kernel/Output/HTML/Ticket/OverviewMenuSort.pm
```diff
@@ -22,16 +27,8 @@ sub new {
...
-        $Self->{$_} = $Param{$_} || die "Got no $_!";
-    }
-
-    $Self->{LanguageObject} = $Self->{LayoutObject}->{LanguageObject}
-        || Kernel::Language->new(
-        %{$Self},
-        UserLanguage => $Self->{LayoutObject}->{UserLanguage},
-        );
+    # get Action param
+    $Self->{Action} = $Param{Action} || die "Got no Action!";
```
mgruner: Hi Sanjin,
you are right, sorry for my mistake about the UserID. But in this place here I believe it is missing?

s7design: Hi @mgruner ,

Yes, you are right, that one slipped away, my bad. It's corrected now and pushed for merge.

mgruner:
Merged, thank you very much! I made some slight improvements to OverviewSmall.t, maybe they are helpful for you.

PR #573, opened 2015-05-11
Master port to OM - Output - Notificaation_and_Heder

s7design:
Hi @mgruner 

There are ported modules for kernel/Output/Notification and Kernel/Output/Heder.

As you said recently when we started porting for Kernel/Output, we did not make Selenium tests for Notification modules, but we tested them manually.
If you think that it is necessary we will be able to make this.

Regards
Zoran


mgruner:
Merged, thank you! I also ported the CustomerHeaderMetaTicketSearch.pm and reorganized the HeaderMeta modules for a better overview in a subdirectory.

PR #574, opened 2015-05-12
Master - port to OM output ticket menu process

s7design:
Hi @mgruner 

There are ported Ticket/MenuProcess.pm to OM and test for it.

Please check this in constructor:

<pre>
   for (qw( LayoutObject UserID )) {
         $Self->{$_} = $Param{$_} || die "Got no $_!";
     }

    $Kernel::OM->ObjectParamAdd(
        'Kernel::Output::HTML::NavBarAgentTicketProcess' => {
            %{$Self},
        },
    );
    $Self->{NavBarModule} = $Kernel::OM->Get('Kernel::Output::HTML::NavBarAgentTicketProcess');
</pre>

As you can see we add $Self param for NavBarAgentTicketProcess.pm, we will port to OM also NavBarAgentTicketProcess.pm next days, and maybe then it will not be needed.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Ticket/MenuProcess.pm
+++ b/Kernel/Output/HTML/Ticket/MenuProcess.pm
@@ -11,7 +11,13 @@ package Kernel::Output::HTML::Ticket::MenuProcess;
 use strict;
 use warnings;
 
-use Kernel::Output::HTML::NavBarAgentTicketProcess;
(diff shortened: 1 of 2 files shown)


mgruner:
Hi Zoran. Please include these changes when you port the navbar modules, this is an incomplete state (as you said) that I would not like to merge.

PR #575, opened 2015-05-13
improved selenium test for service_customeruser

s7design:
Hi @mgruner,

There is fix for CustomerUserService Selenium test. I saw that there is issue in the test because we created test customer and service from fronted and the problem is with Customer, now there is dropdown for Company. Because of that this test was not good. I changed this test and now CustomerUser and Service are created in backed, I believe that it is much better in this case. 

If you think that it is relevant you can merge in master.

Regards
Zoran


mgruner:
Merged, thank you very much!

PR #576, opened 2015-05-13
Add support for Gentoo/emerge in bin/otrs.CheckModules.pl

eliasp:


mgruner:
Merged in 0231e2f, thanks a lot Elias! I removed the additional '' around the emerge package name as it seemed unneeded and causing confusing output. Please let me know if you have any issues with this.

eliasp:
Hi Martin,
I added those additional quotes intentionally, as Gentoo allows to specify package names like `>pkg-category/packagename-1.2.3` which then requires quoting to prevent the shell from interpreting allowed operators like `>`, `>=`, `~` or a suffix for a USE flag requirements like `dev-perl/GD[jpeg,png,truetype]` when copy`n pasting the output into a shell.

Although the current PR doesn't contain any of those special operators for now, I might add a few ones soon to make some requirements (e.g. `truetype` support in `dev-perl/GD`) explicit.

An alternative could be to print the install command colored and/or bold to highlight it instead of putting it into quotes to prevent this "double quoting".

For now the PR is fine, I might implement the suggested changes with another PR.

PR #577, opened 2015-05-15
Modularize Postmaster CheckFollowUp

moritz:
Pluggable architecture for deciding whether an email belongs to an existing ticket.

This ports two existing checks (Ticket number in subject, and `In-Reply-To`/`References` header) to the new API.


mgruner:
Merged in 5610fd4, thank you very much. Very nice!

PR #578, opened 2015-05-15
 Master port to OM - Output/NavBar

s7design:
Hello,

Hereby are ported NavBar modules from Output with there Selenium tests. There was one issue that we found out during our work. Module NavBar/CustomerCompany.pm did not perform as it is expected. Module was not removing NavBar feature when CustomerCompany support is disabled. We tried couple of different solutions and found out one that was working and is most appropriate to the rest of modules.

Furthermore we had to change Framework.xml for same module (description and NavBarName). It must be leftover from previous changes since it was placing this module in Admin NavBar, which was wrong since it should be shown in Customers. Please during QA, verify these changes.

Lastly, selenium test for NavBar/ModuleAdmin is excluded since it is same to already existing Admin.t test. 

Kind regards,
Sanjin
S7


Kernel/Output/HTML/NavBar/CustomerTicketProcess.pm
```diff
@@ -95,26 +80,24 @@ sub Run {
...
-        $Self->{ActivityDialogObject}   = Kernel::System::ProcessManagement::ActivityDialog->new( %{$Self} );
-        $Self->{TransitionActionObject} = Kernel::System::ProcessManagement::TransitionAction->new( %{$Self} );
-        $Self->{TransitionObject}       = Kernel::System::ProcessManagement::Transition->new( %{$Self} );
-        $Self->{ProcessObject}          = Kernel::System::ProcessManagement::Process->new(
-            %{$Self},
-            ActivityObject         => $Self->{ActivityObject},
-            ActivityDialogObject   => $Self->{ActivityDialogObject},
-            TransitionObject       => $Self->{TransitionObject},
-            TransitionActionObject => $Self->{TransitionActionObject},
+        $Kernel::OM->ObjectParamAdd(
```
mgruner: This seems to be wrong/unneeded.

Kernel/Output/HTML/NavBar/AgentTicketProcess.pm
```diff
@@ -93,26 +78,24 @@ sub Run {
...
-        $Self->{ActivityDialogObject}   = Kernel::System::ProcessManagement::ActivityDialog->new( %{$Self} );
-        $Self->{TransitionActionObject} = Kernel::System::ProcessManagement::TransitionAction->new( %{$Self} );
-        $Self->{TransitionObject}       = Kernel::System::ProcessManagement::Transition->new( %{$Self} );
-        $Self->{ProcessObject}          = Kernel::System::ProcessManagement::Process->new(
-            %{$Self},
-            ActivityObject         => $Self->{ActivityObject},
-            ActivityDialogObject   => $Self->{ActivityDialogObject},
-            TransitionObject       => $Self->{TransitionObject},
-            TransitionActionObject => $Self->{TransitionActionObject},
+        $Kernel::OM->ObjectParamAdd(
```
mgruner: Wrong/unneeded?

Kernel/Output/HTML/Ticket/MenuProcess.pm
```diff
@@ -129,8 +137,16 @@ sub Run {
         return if ( !$ACLLookup{ $Param{Config}->{Action} } );
     }
 
+    # use the nav bar module to check for process ACLs to prevent code duplication
+    $Kernel::OM->ObjectParamAdd(
```
mgruner: Wrong/unneeded?

mgruner:
Hi Sanjin,
this looks good. The changes to AdminCustomerCompany also seem fine. I had a UT failure if a process already existed in the DB before the test was run.

Please also add code to DBUpdate-to-5.pl to migrate the configuration settings if they were locally changed. In this case they are stored in ZZZAuto.pm and NOT updated when a new XML files is placed. That is why the DBUpdate script is needed.

Could you please look at the inline comments and follow up?
Best regards, mg

mgruner:
Merged in 4d30cb1, thank you, very nice!

PR #579, opened 2015-05-19
Master - improved selenium tests with processes

s7design:
Hi @mgruner 

As we discussed I have made this PR with some improvement for Selenium test from QutputNavBar.
I also changed  MenuProcess.pm, as you suggested in the previous PR, %{$Self} is not needed in ObjectParamAdd()

Please follow-up it, I hope that it is better with the changes now. 

Regards
Zoran


mgruner:
Merged, very nice, thank you!

PR #580, opened 2015-05-20
Master port to OM - Output/Dashboard

s7design:
Hi @mgruner,

We ported some modules from Output/Dashboard. We are still doing on others Dashoard modules, but the rest we are planning to add in another PRs (maybe two more). It will be easier for follow-up in my opinion.
Because of that we did not changed DBUpdate-to-5.pl, but at the end of porting all Output/Deshboard modules we will change it of course.

Regards
Zoran


mgruner:
Merged, thank you!

PR #581, opened 2015-05-20
Master port to OM - Output/Dashboard - part2

s7design:


scripts/test/Selenium/Output/Dashboard/EventsTicketCalendar.t
```diff
@@ -0,0 +1,240 @@
...
+            Valid => 1,
+            Key   => 'Ticket::Frontend::AgentTicketPhone###DynamicField',
+            Value => \%SysConfigDynamicField,
+        );
+
+        # get customer user object
+        my $CustomerUserObject = $Kernel::OM->Get('Kernel::System::CustomerUser');
+
+        # create customer user
+        my $TestCustomerUser   = "SeleniumCustomer" . rand int(1000);
```
mgruner: Please check in the Helper.pm, there is a method that gives you a test customer user and also cleans it up at the end, just like TestUserCreate().

scripts/test/Selenium/Output/Dashboard/EventsTicketCalendar.t
```diff
@@ -0,0 +1,240 @@
...
+            $TestCustomerUserID,
+            "Customer user $TestCustomerUser - created"
+        );
+
+        # create test ticket that will be shown on EventsTicketCalendar
+        my $ScriptAlias = $Kernel::OM->Get('Kernel::Config')->Get('ScriptAlias');
+        $Selenium->get("${ScriptAlias}index.pl?Action=AgentTicketPhone");
+
+        my $AutoCompleteString
+            = "\"$TestCustomerUser $TestCustomerUser\" <$TestCustomerUser\@localhost.com> ($TestCustomerUser)";
```
mgruner: Please create the ticket directly in the DB. I'm having trouble with this, it cannot find the TicketID if I run selenium.

s7design: Yes we will try to do on this way, to be honest we thought about it, but we had a problem with add Dynamic_filed to ticket in beckend, but I hope we will find how to do that.
all of that will be done as soon as possible.

s7design:
Hi Martin,

I changed test as you suggested. I hope that they are much better now.

I also enabled appropriate  SysConfig item for each of test. Some of them is enabled by default, but maybe it is better to be sure that test will be successfully run. I had some situation on my local system that some of these SysConfig item were disabled and test was failed.

Regars
Zoran

mgruner:
Merged, thank you! I just removed one duplicated my $LayoutObject in UserOnline.pm.

PR #583, opened 2015-05-22
Postmaster: Followup-check by dynamic field header

moritz:
This new module lets you define a header that is connected to a dyanmic field. If an existing ticket (in configurable states) has this dynamic field set to the same value as the header in an incoming mail, the mail is considered a follow-up to this ticket.

Use cases include for example cron jobs that only send mails on errors; with this feature, you can make sure that all mails form a certain cron job end up in the same ticket, until somebody resolves the problem and the ticket.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10121,7 +10121,7 @@ Thanks for your help!
         </Setting>
     </ConfigItem>
     <ConfigItem Name="PostMaster::CheckFollowUpModule###0800-DynamicField" Required="1" Valid="1">
-        <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dynamic field (and in configurable states) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
+        <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dynamic field (and in one of the state types configred in &quot;StateType&quot; below) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
         <Group>Ticket</Group>
         <SubGroup>Core::PostMaster</SubGroup>
         <Setting>
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10121,7 +10121,7 @@ Thanks for your help!
         </Setting>
     </ConfigItem>
     <ConfigItem Name="PostMaster::CheckFollowUpModule###0800-DynamicField" Required="1" Valid="1">
-        <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dyanmic field (and in configurable states) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
+        <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dynamic field (and in configurable states) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
         <Group>Ticket</Group>
         <SubGroup>Core::PostMaster</SubGroup>
         <Setting>
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10120,7 +10120,7 @@ Thanks for your help!
             </Hash>
         </Setting>
     </ConfigItem>
-    <ConfigItem Name="PostMaster::CheckFollowUpModule###0800-DynamicField" Required="1" Valid="1">
+    <ConfigItem Name="PostMaster::CheckFollowUpModule###0800-DynamicField" Required="0" Valid="0">
         <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dynamic field (and in one of the state types configred in &quot;StateType&quot; below) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
         <Group>Ticket</Group>
         <SubGroup>Core::PostMaster</SubGroup>
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10120,6 +10120,26 @@ Thanks for your help!
             </Hash>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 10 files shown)


mgruner:
Nice! I'll add some inline comments.

Kernel/Config/Files/Ticket.xml
```diff
@@ -10120,6 +10120,26 @@ Thanks for your help!
             </Hash>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="PostMaster::CheckFollowUpModule###0800-DynamicField" Required="1" Valid="1">
```
mgruner: This should probably be Required="0" Valid="0"?

Kernel/Config/Files/Ticket.xml
```diff
@@ -10120,6 +10120,26 @@ Thanks for your help!
...
+        <Description Translatable="1">Executes follow up checks on a dynamic field  header. If a ticket with the same content of the dyanmic field (and in configurable states) is found, it is considered a follow-up to that ticket. The dynamic field must be configured as a text field.</Description>
```
mgruner: Typo (dyanmic). Also please add a hint about the states.

Kernel/System/PostMaster.pm
```diff
@@ -155,7 +155,7 @@ sub Run {
     my $GetParam = $Self->GetEmailParams();
 
     # check if follow up
-    my ( $Tn, $TicketID ) = $Self->CheckFollowUp( %{$GetParam} );
+    my ( $Tn, $TicketID ) = $Self->CheckFollowUp( GetParam => $GetParam );
```
mgruner: This looks like the existing CheckFollowUp modules might need to be updated, right?

Kernel/System/PostMaster/Filter/CheckFollowUpDynamicField.pm
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    my $DynamicFieldConfig = $Kernel::OM->Get('Kernel::System::DynamicField')->DynamicFieldGet(
+        Name => $Self->{DynamicField},
+    );
+    if ( !IsHashRefWithData($DynamicFieldConfig) ) {
+        return;
+    }
+
+    my $HeaderValue = $Param{GetParam}{ $Self->{Header} };
```
mgruner: Should this not be ```$Param{GetParam}->{ $Self->{Header} }```? Similar in other places below.

Kernel/System/PostMaster.pm
moritz: @mgruner I've updated all CheckFollowUp moudles in core. (Note that some of them don't even use the GetParams, but obtain their information from the ParserObject; those didn't need updating).

Kernel/System/PostMaster/Filter/CheckFollowUpDynamicField.pm
moritz: @mgruner  Perl allows allows you to emit all except the first arrow in a chain of dereferences, so it's equivalent. Unless you have a coding style which says otherwise, of which I'm not aware :-)

moritz:
@mgruner Thanks for the review. I think I've addressed all of your feedback now.

mgruner:
We usually use {}->{}, so I'd appreciate if you could update that as well.

scripts/test/PostMaster/CheckFollowUpDynamicField.t
```diff
@@ -0,0 +1,140 @@
...
+        Config          => { dummy => 1},
+    );
+    $DynamicField = $DynamicFieldObject->DynamicFieldGet(
+        Name => $FieldName,
+    );
+}
+
+
+my %TicketsToDelete;
+sub ProcessMailFile {
```
mgruner: Another minor comment here: we discourage the use of package subs in tests as they will survive past eval {}. If you update your OTRSCodePolicy module it will also complain about this. Please use ```my $ProcessMailFile = sub { ... }``` instead.

moritz:
@mgruner ok, I now use `{}->{}` and  `my $ProcessMailFile = sub { ... }`.

moritz:
Hi there, is there anything still to do for me here?

mgruner:
Hello @moritz. I am very sorry for not getting back to you earlier! Your proposal looks very nice. It came to my attention that OTRS already has a kindof similar feature, the external ticket number recognition (ExternalTicketNumberRecognition.pm and .t), which is kindof a hack to make the follow up detection work by manipulating the subject. See also http://otrs.github.io/doc/manual/admin/stable/en/html/email-settings.html#filter-module-externalticketrecognition-example.
This module has a slightly different feature set, for example it searches in Subject and/or Body to extract the external ticket number that is to be stored in a dynamic field later. Do you think that it would be useful and possible to merge these features into your new module so that we can drop the old one? I think this is also closer to your use case where the information comes in the subject, right?
Best regards, mg

moritz:
The obvious generalization for this branch would be to allow inspection of any header with a regex, and use the result of the regex match.

That way one can configure it to act on the subject to emulate the external ticket number recognition, or an a dyanmic field header with regex `.*` to achieve the current feature set.

mgruner:
Yes, this sounds correct, plus the possibility to search in the body. Would you be able to make these changes, dropping ExternalTicketNumberRecognition.pm and -.t (maybe take over helpful test cases)?

mgruner:
Hello @moritz, will you be able to update this PR with these changes? We are getting close to the release.

moritz:
@mgruner I'll try, but can't promise anything.

mgruner:
Hello @moritz. Today I discovered that the follow up check layer modularization was not complete and fixed that in 78ccbe3. I also included the API change of this PR at hand in 244446d, so that the follow up checks now may modify the GetParams.
This PR here requires some changes now. Therefore I'll close it for now. If you would be able to rebase your branch and merge with ExternalTicketNumberRecognition, this would be awesome. One problem might be that ExternalTicketNumberRecognition supports multiple configurations by registering several instances of the same filter as a pre filter. We might need something similar here, several XML config settings that the follow up check module uses.
Please open another PR then. Depending on the time it will go into OTRS 5 or 6.
Best regards, mg

moritz:
OK; after reviewing the external ticket number recognition module, I think the sane approach is to slowly extend it to do what I want, instead of other way around. Pull requests will be forthcoming :-)

mgruner:
Ok. Currently this is a regular pre filter. Would you want to change it to become a follow up detection module?

moritz:
@mgruner yes, I think so; but I still have to take a closer look before I can be sure.

PR #585, opened 2015-05-22
Master port to OM - Output/Dashboard - Part3

s7design:


s7design:
Hi @mgruner 

As I wrote on an email in my opinion there is wrong URL for image in the following config item 
In SysConfig item for this module (DashboardBackend###0200-Image) there is key URL with wrong content of image
* http://www.otrs.com/uploads/pics/jointhecommunity_02.jpg
we suggest that this URL could be changed to
* https://www.otrs.com/wp-uploads//2013/10/OTRS_Logo-300x170.png
or maybe you are some other idea. What do you think about this?
If you want we will be able to change it in the xml file.

Regards
Zoran

s7design:
Hi @mgruner,

I fixed wrong image for SysConfig Dashboard item as we discussed.
Now Dashboard ported to OM PRs are ready for follow-up. 

If you are any suggestion for improvement we will be able to do that.

Regards
Zoran

mgruner:
Merged in 715d72d, thank you!

PR #587, opened 2015-05-25
Master port to OM - Output/Dashboard - Part4

s7design:
Greetings,

Hereby are last ported Output/Dashboard modules with selenium tests. There is one mark I would like to point out and check for your opinion.

In Selenium test Selenium/Output/Dashboard/TicketGeneric.t in order to test escalation times, had to use FixedTimeSet() function @ line 33. This results in console errors on start of the test e.g. ( "Subroutine new redefined at /opt/otrs/Kernel/System/Time.pm line 46." ). Function works and test is performing as it should. Later in the test @ line 95, set time is pushed forward for 5 minutes, so escalation can take effect.

Last time FixedTimeSet() function was used, there was remark that we should avoid using it. Is there any way escalation can be triggered without manipulating time? What is your opinion on this? If there is something we can change, let us know.

Kind regards,
Sanjin
S7


mgruner:
Hello Sanjin,

this looks good. Using FixedTimeSet is ok if it is enough to change the time in the unit test process. The Selenium tests always run on the web server which has the regular time. But the way you use it (to date things back) works correctly in this case. :+1: 

I'm getting errors in TicketGeneric.t: ```/ws/otrs5/scripts/test/Selenium/Output/Dashboard/TicketGeneric.t #50 - TicketNew dashboard plugin test ticket link - found, 54 - TicketOpen dashboard plugin test ticket link - found```. Probably because I have other tickets in the system from other tests. You can simulate with ```bin/otrs.Console.pl Dev::Tools::Database::RandomDataInsert```. Please find a way to make this more stable. Maybe using permissions (allow user only access to a new queue where the test tickets live).

Best regards, mg

mgruner:
Merged in 4697146, thanks!

PR #588, opened 2015-05-26
Consistent wording on Dashboard

eliasp:
All Dashboard modules show `none` when no matching tickets were found
except of `Ticket Queue Overview` which shows `No data found`.
This makes for an inconsistent appearance of the Dashboard modules and
confuses users which might interpret this as an error message instead.


mgruner:
Thanks!

PR #589, opened 2015-05-26
Fix bug#11307 -  Reference to uninitialized value at change of sort o…

noritnk:
This PR is fix for bug#11307.


mgruner:
Thanks a lot, @noritnk. I applied a modified solution, please see http://bugs.otrs.org/show_bug.cgi?id=11307.

PR #590, opened 2015-05-27
Master port to OM - OutputCustomer

s7design:
Hi @mgruner,

There are ported modules for Output/Customer_.
You can see that we changed some lines in DBUpdate-to-5.pl, I hope you like this. I believe it is better now.

Regards
Zoran


mgruner:
Merged in 019860c6f3eee306bc00142ff64433c6864ca85e, thanks! I made a small change to the namespaces: CustomerUser / CustomerNewTicket instead of just Customer, as these are actually different module layers. Could you please also test this and let me know if everything is ok for you?

s7design:
Hi @mgruner ,

It works good with your changes.

Regards,
Sanjin

PR #591, opened 2015-05-27
Make sure cache is correct for Output/Dasboard modules

s7design:


s7design:
Hi @mgruner,

I made this PR because I saw while I testing these Dashboard Selenium tests that cache is not cleaned up after executing these ones. I believe that it is better now.

Regards
Zoran

mgruner:
Hi @s7design, can you please tell me why you think this would be needed? As long as you use the OTRS API, the cache should always be consistent. Only if you make direct manipulations in the DB, like deleting a Queue via SQL, the cache will not be updated and should be cleared.

s7design:
Hi @mgruner,

I have just tested these modules again, and to be honest I do not sure if I am right :)
I can see that there are some data in cache after executing tests, and because of that I thought that it could make some inconsistent state on the  Dashboard page.
For example if you run ProductNotify.t, in the cache there will be DashboardProductNotify, and as I mentioned, maybe it is not a problem.
Maybe it was only my wrong opinion, sorry. 
If I made mistake, we can close this PR.

Regards
Zoran

mgruner:
No problem! If we would really want to keep things clean we'd have to clear the complete cache after every test. So far this seems not neccessary.

PR #592, opened 2015-05-27
'MaxRequestsPerChild' → 'MaxConnectionsPerChild'

eliasp:
Since Apache httpd 2.3.9, the `MaxRequestsPerChild` directive is named `MaxConnectionsPerChild`.
Defaulting to the old name for now in case `mod_version` isn't available is fine for now, as the old name will be still available for backwards compatibility.

See also: http://httpd.apache.org/docs/2.4/en/mod/mpm_common.html#maxconnectionsperchild


mgruner:
Hi Elias, thank you very much for your proposal. As long as the old name is still supported, I think we can keep it like this in the config, reducing complexity. Whenever we do have to upgrade, we should use your approach.

PR #593, opened 2015-05-27
Updated DBUpdate-to-5.pl for Output/Dashboard

s7design:
HI @mgruner,

We forgotten to update  DBUpdate-to-5.pl for Dashboard modules. I have made the new PR because we changed a little bit in meantime (PR https://github.com/OTRS/otrs/pull/590). I think maybe there could be a conflict during merging, but we will be able to solve it.  

Regards
Zoran


mgruner:
Merged, thanks a lot.

PR #594, opened 2015-05-28
Remove inconsistent check in TicketGet

moritz:
Previously, Kernel::System::Ticket->TicketGet with DynamicFields => 1
would ignore dynamic fields whose Config is an empty hash ref.

This is inconsistent with the creation (and the rest of the handling of)
dynamic fields, which generally allow Config to be an empty hash ref.

This case does not seem to occur when generating dynamic fields through
the admin frontend, but it happened to me when writing tests, causing
a long debugging session in search for the missing dynamic field.


mgruner:
:+1:

PR #595, opened 2015-05-28
 Master port to OM - OutputFilter

s7design:
Hello,

Here's PR with ported OutputFilter's modules. During porting and creating Selenium test for OutputFilterTextAutoLink.pm, we stumble upon problem. This module did not work for us when activated.
After some thorough research and testing of module, we found out, what we think is possible cause of bug. RegExp is correctly loaded from SysConfig, but it seems that at line 76 it is not assigned correctly further to MatchData. We tried to bypass it by hardcoding RegExp there and assigning it to MatchData, which worked partially. What we tried is:  replacing "(m{([\s:]$RegExp)}gxi)" with just "$RegExp". In AgentTicketZoom it did indeed matched and inserted appropriate images and links but in the middle of matched text. This might be totally wrong approach and way, but it was one of many ideas we tried around during testing.

Our field of expertise with RegExp (will work on improving it), limit us further investigating cause of problem, hopefully you can help us there and point in right direction.

It seems that this issue is long known to you, it is reported in bug #3923 and been popping up over the years by various bug reporters. Please just let us know if this module works ok for you, or is it work in progress.

King regards,
Sanjin
S7


mgruner:
Hi Sanjin, I merged your changes, thank you. I didn't look into the bug yet, for now just porting is ok.

mgruner:
Hello, I decided to revert this. As you can see we have different types of output filters: http://otrs.github.io/doc/manual/developer/stable/en/html/otrs-module-layers.html#outputfilter. I'd recommend that you change the name space to Output/HTML/FilterText for these filters. Can you please update the PR accordingly?

mgruner:
Sorry for the misinformation in the first place.

s7design:
Hi @mgruner 

It is not a problem, we will do that soon.

Regards
Zoran

s7design:
Hi @mgruner 

PR is updated with changed name space.

Regards,
Sanjin

Kernel/Output/HTML/FilterText/TextAutoLink.pm
```diff
@@ -6,11 +6,13 @@
 # did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
```
mgruner: I think we don't need Text twice. FilterText/AutoLink would be enough.

s7design:
Hi @mgruner 

Think it's better like this as well, PR is updated.

Regards,
Sanjin

mgruner:
Merged, thank you Sanjin!

PR #596, opened 2015-05-29
 Master port to OM - Output/Layout

s7design:
Hello,

Hereby is first part of Layout port to OM. LayoutLinkObject.pm and LinkObjectTicket.pm will be ported in next PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Layout/Datepicker.pm
+++ b/Kernel/Output/HTML/Layout/Datepicker.pm
@@ -11,9 +11,7 @@ package Kernel::Output::HTML::Layout::Datepicker;
 use strict;
 use warnings;
 
-our @ObjectDependencies = (
-    'Kernel::Config',
-);
+our $ObjectManagerDisabled = 1;
 
 =head1 NAME
 
@@ -58,7 +56,8 @@ sub DatepickerGetVacationDays {
     # translate the vacation description if possible
     for my $Month ( sort keys %{$TimeVacationDays} ) {
         for my $Day ( sort keys %{ $TimeVacationDays->{$Month} } ) {
-            $TimeVacationDays->{$Month}->{$Day} = $Self->{LanguageObject}->Translate( $TimeVacationDays->{$Month}->{$Day} );
+            $TimeVacationDays->{$Month}->{$Day}
+                = $Self->{LanguageObject}->Translate( $TimeVacationDays->{$Month}->{$Day} );
         }
     }
 
--- a/Kernel/Output/HTML/Layout/Datepicker.pm
+++ b/Kernel/Output/HTML/Layout/Datepicker.pm
@@ -6,14 +6,19 @@
 # did not receive this file, see http://www.gnu.org/licenses/agpl.txt.
 # --
 
-package Kernel::Output::HTML::LayoutDatepicker;
+package Kernel::Output::HTML::Layout::Datepicker;
 
 use strict;
 use warnings;
 
+our @ObjectDependencies = (
+    'Kernel::Config',
+    'Kernel::Language',
+);
+
 =head1 NAME
 
-Kernel::Output::HTML::LayoutDatepicker - Datepicker data
+Kernel::Output::HTML::Layout::Datepicker - Datepicker data
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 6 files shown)


Kernel/Output/HTML/Layout/Datepicker.pm
```diff
@@ -36,30 +41,35 @@ Returns a hash of all vacation days defined in the system.
...
+        if ( $ConfigObject->Get( "TimeZone::Calendar" . $Param{Calendar} . "Name" ) ) {
+            $TimeVacationDays        = $ConfigObject->Get( "TimeVacationDays::Calendar" . $Param{Calendar} );
+            $TimeVacationDaysOneTime = $ConfigObject->Get(
                 "TimeVacationDaysOneTime::Calendar" . $Param{Calendar}
             );
         }
     }
 
+    # get language object
+    my $LanguageObject = $Kernel::OM->Get('Kernel::Language');
```
mgruner: Here it would be better to keep using $Self->{LanguageObject}. This will stay for now in the LayoutObject as it is used in many places in OTRS.

Kernel/Output/HTML/Layout/Ticket.pm
```diff
@@ -37,7 +51,7 @@ sub AgentCustomerViewTable {
         $Self->FatalError( Message => 'Need Hash ref in Data param' );
     }
     elsif ( ref $Param{Data} eq 'HASH' && !%{ $Param{Data} } ) {
-        return $Self->{LanguageObject}->Translate('none');
+        return $Kernel::OM->Get('Kernel::Language')->Translate('none');
```
mgruner: Similar here.

Kernel/Output/HTML/Layout/Loader.pm
```diff
@@ -223,11 +234,14 @@ sub LoaderCreateCustomerCSSCalls {
 
     $Self->{LoaderObject} ||= Kernel::System::Loader->new( %{$Self} );
```
s7design: Hi @mgruner ,

Just got a question/confirmation on this. Should we load loader object this way, or use object manager and ObjectParamAdd function to pass whole $Self to it.

Regards,
Sanjin

Kernel/Output/HTML/Layout/Ticket.pm
```diff
@@ -146,7 +163,7 @@ sub AgentCustomerViewTable {
                 my $CompanyValidID = $Param{Data}->{CustomerCompanyValidID};
 
                 if ($CompanyValidID) {
-                    if ( !$Self->{MainObject}->Require('Kernel::System::Valid') ) {
+                    if ( !$MainObject->Require('Kernel::System::Valid') ) {
                         $Self->FatalDie();
                     }
```
s7design: Hi @mgruner

This is same question like in Loader.pm (line 235). To keep calling object like it is now, or to use object manager.

Regards,
Sanjin

s7design:
Hi @mgruner 

PR is updated accordingly to your comments.

Regards,
Sanjin

Kernel/Output/HTML/Layout/Loader.pm
mgruner: Well the loader takes no constructor arguments, right? So you don't have to pass anything. The ||= is no more needed as the OM keeps the loader copy. You can just use my $LoaderObject = $Kernel::OM->Get() instead.

Kernel/Output/HTML/Layout/Ticket.pm
mgruner: Just switch to OM. This extra handling is not really needed.

s7design:
Hi @mgruner 

I updated the PR as you suggested.
I also changed a little the code.
As you can see I deleted from all modules: 
 * our @ObjectDependencies = (
and added
+our $ObjectManagerDisabled = 1.

These Layout_ modules is not real class, and there will not be needed creating instance of them, or creating objects.

I am not sure about this, but this is my proposal for now, if I am wrong, I will be able to change that.
I don't now how I can test if Loader.pm works well, I tested all modules from Selenium/Output and all of them worked fine
If you have any suggestion please let me know. 

Regards
Zoran

mgruner:
Merged, thank you!

PR #597, opened 2015-06-01
jsTree updated to v3.1.1; Ported Core.UI.TreeSelection to new API

dvuckovic:
Latest **jsTree** version has somewhat different API than previous 1.0 pre-release. **TreeSelection** script has been reviewed and updated where necessary. Some logic had to be changed completely (for example, JSON data structure and way that entries are initially selected on tree load), but mostly everything is the same. Tested on both multiple and single select fields.


mrcbnsls:
Hi @dvuckovic,

Thanks for the PR. Currently this can't be merged because the bug we created the patch for in the older version is no longer fixed. I could see that you just removed the patch, but in my tests the wrong behavior is still there in the new version. Please have a look. Thanks!

Marc

dvuckovic:
Hi @mrcbnsls, sorry about this, I misunderstood the bug description and thought it was fixed. It should work as expected now in d9378a1, I fixed it without patching the library. Please take a look and let me know what you think.

Regards,
Dusan

mrcbnsls:
Merged in 270f8eb8800cab5ef496ff4f6875dc2a9371be65, thanks a lot!

PR #598, opened 2015-06-02
Master port to OM - Output/LinkObject

s7design:
Hello,

Hereby is PR for last ported modules from Output directory, covering Layout/LinkObject.pm and  LinkObject/Ticket.pm with AgentLinkObject.t selenium test.

In Layout/LinkObject.pm, line 263-264 and 448-449, we left creation of the new layout object instance way it was before. We tried porting it, using ObjectParamAdd(), but only this way worked. If you could just confirm this is right approach or should we use different way. 

In LinkObject/Ticket.pm, since objects are called in constructor for the creation of new LayoutObject instance, we didn't port them. Should we call them again in every function where they are needed or to leave calling from the constructor. Need confirmation here as well.

Since these are last modules from Output, we updated Layout.pm so it doesn't load old sub modules.

Regards,
Sanjin
S7


mgruner:
Merged and completed in 9f6ffb0, thanks! I believe you forgot to port LinkObject/Ticket.pm, I have done that now.

PR #599, opened 2015-06-02
Master Selenium - improved sleep 1

s7design:
Hi @mgruner 

There is improvement for some Selenium test. As we said recently we will make a couple PR for sleep in the tests.
This is first one, we will make the others soon.

In this PR there is one interesting case with AdminDynamicField.t, I think that I have mentioned this before. It worked fine with firefox, but in case phantomjs is used it did not work. We add our proposal how it could be solved. We check which browser is used duo to $Selenium->accept_alert(); does not work in phantomjs and $Selenium->execute_script($CheckConfirmJS);   does not work in Firefox (see line 118).
I am not sure if you like this hard-coding, but it works now :)

Please let me know what do you mean about it.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
@@ -108,38 +108,15 @@ $Selenium->RunTest(
                     "#ValidID stored value",
                 );
 
-                $Selenium->go_back();
+                $Selenium->get("${ScriptAlias}index.pl?Action=AdminDynamicField");
 
                 # delete DynamicFields, check button for deleting Dynamic Field
                 my $DynamicFieldID = $Kernel::OM->Get('Kernel::System::DynamicField')->DynamicFieldGet(
                     Name => $RandomID
                 )->{ID};
 
-                # depending on browser use dirrefernt delete approach
-                if ( $TestBrowser eq 'firefox' ) {
-
-                    $Selenium->find_element(
-                        "//a[contains(\@data-query-string, \'Subaction=DynamicFieldDelete;ID=$DynamicFieldID' )]"
-                    )->click();
-
-                    # check for opened confirm text
-                    my $LanguageObject = Kernel::Language->new(
-                        UserLanguage => $Language,
-                    );
-
-                    $Self->Is(
-                        $Selenium->get_alert_text(),
-                        $LanguageObject->Get(
-                            'Do you really want to delete this dynamic field? ALL associated data will be LOST!'
-                        ),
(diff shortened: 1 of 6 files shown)


mgruner:
Hi Zoran,
the current JS-Based solution works fine for me on phantomjs and also on latest Firefox with Selenium on MacOS (Mavericks). What fails for you?
Regards, mg

scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
```diff
@@ -42,6 +46,9 @@ $Selenium->RunTest(
         $Selenium->find_element( "table thead tr th", 'css' );
         $Selenium->find_element( "table tbody tr td", 'css' );
 
+        # get which browser is used for testing
+        my $TestBrowser = $ConfigObject->Get('SeleniumTestsConfig')->{browser_name};
```
mgruner: This might not work for me as I always use 'firefox' as browser name. This allows me to use Selenium/Firefox or phantomjs (port 4444) with the same configuration. For phantomjs this setting is ignored.

scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
```diff
@@ -98,13 +108,38 @@ $Selenium->RunTest(
                     "#ValidID stored value",
                 );
 
+                $Selenium->go_back();
```
mgruner: Why should we use go_back instead of get? I remember that this caused problems for me.

scripts/test/Selenium/Agent/Admin/AdminNotification.t
```diff
@@ -43,7 +48,6 @@ $Selenium->RunTest(
 
         # test filter for Notification
         $Selenium->find_element( "#FilterNotification", 'css' )->send_keys("Agent::AddNote");
-        sleep 1;
```
mgruner: This will also not work for me I guess as the filter is only executed after some hundred milliseconds. On a very fast machine this fails.

s7design:
Hi @mgruner ,

I removed in AdminDynamicField.t check for different browsers, as it was before. Yet there is still this error for me when executed with Firefox "not ok 69 - Exception in Selenium': Error while executing command: An unknown server-side error occurred while processing the command.: window.getLastConfirm is not a function". Please if you could just confirm that this is working now for in Firefox.

In AdminNotification.t, i returned one second sleep. This should solve issues on very fast maschines.

Regards,
Sanjin

mgruner:
Merged, thanks! I removed AdminNotification.t as this will be dropped soon.

PR #600, opened 2015-06-02
Master Selenium - improved sleep 2

s7design:
Hi @mgruner 

There are some improvement in Selenium tests for Process Management.
The main intention is improvement with sleep, but I also added synchronization of process at the end of tests.

Regards
Zoran


mgruner:
Hello Zoran,
I am getting some errors with phantomjs on my development machine (which is quite fast).
In the Admin*.t files, I needed to place a sleep after the filter is used:
        $Selenium->find_element( "#ActivityFilter", 'css' )->send_keys($ActivityRandom);
        sleep 1; # wait for filter to kick in
Then it worked there.
But I also have an error in Agent/AgentTicketProcess (67 - Ticket open state found on page). There it checks for closed, but the test name says open - strange, please check. 

Two more recommendations: Please don't use yml files from the development/ directory in tests, they are not stable. Just copy it to the folder where also the test file lives (AgentTicketProcess.yml), as this process is closely associated with the test now.
And if you want to wait for a ticket to (re)load, you can use the DOM in the ticket information box: the values have titles, so you could wait for $('p.Value[title=open]') or something similar (untested). Perhaps that is also better in find_element than using get_page_source.

Regards, mg

s7design:
Hi @mgruner,

I updated PR, I hope it works better now.
To be honest I had a lot of problem with this, and  It make me crazy a little :)
I tested these tests on a couple VM and they sometimes works, but sometimes don't ( very strange)

I hope it is OK now, it is tested with the both browsers. 
I also tested  $('p.Value[title=open]') and some version similar to this one, but I did not succeed in this and I left solution with get_page_source. and simple 'sleep'.
If I manage to solve this on that way I agree that it could be better solution.

Regards
Zoran

scripts/test/ProcessManagement/DB/Process/ProcessImport.t
```diff
@@ -482,7 +482,7 @@ for my $Test (@Tests) {
     my $FileRef;
     if ( $Test->{ProcessFile} ) {
         $FileRef = $MainObject->FileRead(
-            Location => $Home . '/scripts/test/sample/ProcessManagement/' . $Test->{ProcessFile},
+            Location => $Home . '/scripts/test/sample/ProcessManagement/TestProcess.yml/' . $Test->{ProcessFile},
```
mgruner: This looks strange?

s7design: Yes, it is my mistake. I see now, I have not tested this test.
I will fix that immediately. 

Thanks

s7design:
I updated PR and fixed my mistake with ProcessImport.t test.
I did some find-replace when I wanted to change new path for TestProcess.yml file, and I had not seen that I accidentally change that test.

Regards
Zoran

mgruner:
Merged with slight changes, thanks!

PR #603, opened 2015-06-09
Sync custom agent preferences from LDAP

seidler2547:
This allows to add any custom agent preference into AuthSyncModule::LDAP::UserSyncMap. We use it for example to sync phone numbers and department from ActiveDirectory to display it in the agent signature.

### DIFF
--- a/Kernel/System/Auth/Sync/LDAP.pm
+++ b/Kernel/System/Auth/Sync/LDAP.pm
@@ -293,6 +293,20 @@ sub Sync {
                     Message  => "Initial data for '$Param{User}' ($UserDN) created in RDBMS.",
                 );
 
+                # copy attributes
+                NEWATTRIBUTE:
+                for my $Attribute ( sort keys %SyncUser ) {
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @seidler2547, thank you very much for your contribution! This looks interesting. Can I ask you to please sign the OTRS CLA as described on https://otrs.github.io?

Regarding the implementation, I would do this differently. AuthSyncModule::LDAP::UserSyncMap is there to specify the LDAP->DB mapping for the user data. You could add a similar setting AuthSyncModule::LDAP::UserPreferencesSyncMap where you explicitly configure which key is mapped to which entry. This would allow for key renaming and avoid conflicts with UserSyncMap.

What do you think?

mgruner:
Please add the new setting to Defaults.pm also.

mgruner:
Hello @seidler2547. Without a CLA I can unfortunately not accept PRs. Sorry for that! Please feel free to reopen / open a new one about this matter if you were able to sign the CLA.
Best regards, Martin

PR #604, opened 2015-06-12
Master - PDF object issue

s7design:
Hi @mgruner 

While we were porting some modules for ITSM packages, we saw that there is bug with Print output.
I felt free to open a bug. There is link to the bug.
http://bugs.otrs.org/show_bug.cgi?id=11332

I solved it. I also added test for the modules that I fixed. I believe that it is good, but I know that it could be solved on another way. And I had some ideas. 

The problem is because PDF object return nothing if PDF SysConfig is not enabled.

~~~perl
sub new {
    my ( $Type, %Param ) = @_;

    # allocate new hash for object
    my $Self = {};
    bless( $Self, $Type );

    # load PDF::API2
    return if !$Kernel::OM->Get('Kernel::Config')->Get('PDF');
...
~~~

Maybe it could be better if we change it on the following way ( return 0 instead return ):

~~~perl
sub new {
    my ( $Type, %Param ) = @_;

    # allocate new hash for object
    my $Self = {};
    bless( $Self, $Type );

    # load PDF::API2
    return 0 if !$Kernel::OM->Get('Kernel::Config')->Get('PDF');
... 
# the same could be solved with the other 'returns'

~~~

Similar to this there was situation with PGP and SMIME objects.
I have not changed PDF.pm yet. It is only my proposal. With my fix I solved the issue, but with this my proposal in the comment, i belive we could solve it on the better way.

What do you think about all of that, please let me know.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentStats.pm
+++ b/Kernel/Modules/AgentStats.pm
@@ -2387,11 +2387,7 @@ sub Run {
             $MainObject->Require('Kernel::System::PDF');
 
             # get PDF object
-            my $PDFObject;
-
-            if ( $ConfigObject->Get('PDF') ) {
-                $PDFObject = $Kernel::OM->Get('Kernel::System::PDF');
-            }
+            my $PDFObject = ( $ConfigObject->Get('PDF') ) ? $Kernel::OM->Get('Kernel::System::PDF') : undef;
 
             # PDF Output
             if ($PDFObject) {
--- a/Kernel/Modules/AgentTicketPrint.pm
+++ b/Kernel/Modules/AgentTicketPrint.pm
@@ -184,7 +184,8 @@ sub Run {
     }
 
     # get PDF object
-    my $PDFObject = $Kernel::OM->Get('Kernel::System::PDF');
+    my $PDFObject
+        = ( $Kernel::OM->Get('Kernel::Config')->Get('PDF') ) ? $Kernel::OM->Get('Kernel::System::PDF') : undef;
 
     # generate pdf output
     if ($PDFObject) {
--- a/Kernel/Modules/AgentTicketSearch.pm
+++ b/Kernel/Modules/AgentTicketSearch.pm
@@ -783,7 +783,8 @@ sub Run {
         elsif ( $GetParam{ResultForm} eq 'Print' ) {
 
             # get PDF object
-            my $PDFObject = $Kernel::OM->Get('Kernel::System::PDF');
+            my $PDFObject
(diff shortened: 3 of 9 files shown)


s7design:
Hi @mgruner 

* I would like to not that I made print tests with enabled PDF, becouse I have a problem in tests if I disabled PDF, because I got standard print dialog with close/print button, but I don't know how I can handle with such dialog, but I also think that these test cases is better with PDF output.
If you want I will be able to make another PR with the Selenium test only, becouse they are not directly connected to this bug.

* And one more note, we had a problem with waiting while PDF output is loaded. We tried to use WaitFor function but it is not posible we got the following error. 
<pre>
 not ok 30 - Exception in Selenium': Error while executing command: An unknown server-side error occurred while processing the command.: $ is not defined
</pre>
I think that the problem is with executing JavaScript if PDF output is used, or something like this.
We also tried to use ACTIVESLEEP block, but as you know it makes a problem with phantomJS.
And at the end I chose the simplest solution with sleep 3 seconds. I know that it is not the best solution, but I don't know the better one that works :)
I left ACTIVESLEEP block commented. I will be able to delete this.

Regards
Zoran


Regards
Zoran

Kernel/Modules/AgentStats.pm
```diff
@@ -2387,11 +2387,7 @@ sub Run {
             $MainObject->Require('Kernel::System::PDF');
 
             # get PDF object
-            my $PDFObject;
-
-            if ( $ConfigObject->Get('PDF') ) {
-                $PDFObject = $Kernel::OM->Get('Kernel::System::PDF');
-            }
+            my $PDFObject = ( $ConfigObject->Get('PDF') ) ? $Kernel::OM->Get('Kernel::System::PDF') : undef;
```
s7design: Hi @mgruner 

I also changed this, this solution could be better.
Regards
Zoran

mgruner:
Hi Zoran, thanks a lot so far for this. :+1:  I merged it without the new selenium tests. Unfortunately get_page_source does not work in phantomjs so the tests fail.
I just removed the auto print for HTML print view. Could you please check if the Selenium tests would work ok with that now? Of course both HTML and PDF would be nice, but it has to work both for Selenium and phantomjs.
Regards, mg

s7design:
Hi @mgruner 

Thanks, it will be done next days. i hope we will manage to find  good solution for Selenium test.
Yes it could be a problem with print output for the both type in phantomjs. We are planning to look into this more detailed.

Regards
Zoran

PR #606, opened 2015-06-15
Master - Changed output of invalid attachments on overview screen

s7design:
Hi @mgruner 

We changed a little admin attachment overview. It is our proposal in this moment.
To be honest we played with it :) 
However maybe it could be interesting and we can change the other Admin overview screen on the same way. 
As I said it is only proposal, but we did this also in purpose of training, because I would like to see how our new colleagues understand some features which have already  been in OTRS, but they should use a couple of  them for implementing something new or alike.

It look like this:
valid_flags_1

Another proposal is something like this (flag at the beginning of table):
valid_flags

What do you think about that, in any case we will be able to provide one PR for some Admin screen 
( e.g. AdminSalutation and AdminSignature  ) with implemented this features for greying out invalid entity at the end of day. And if  you agree with our proposal we can add the flags as well.
Please give me the feedback about it.

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Default.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Default.css
@@ -559,14 +559,6 @@ dd {
     margin: 0px auto;
 }
 
-.Flag.Valid {
-    width: 17px;
-    height: 8px;
-    float: left;
-    cursor: unset;
-    margin: 5px 10px;
-}
-
 .Flag span {
     display: block;
     height: 100%;
@@ -575,11 +567,6 @@ dd {
     /* set a default color for priorities */
     background-color: #cdcdcd;
 }
-
-.Flag.Valid span {
-    cursor: unset;
-}
-
 
 .Flag span.Escalated {
     background-color:#ff505e;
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Zoran,
I like your proposal, but at this point the flags in OTRS are reserved for ticket priorities. Using colors for validity could cause confusion, and we really only want to see valid or not valid (grey/inactive). That is enough and will improve usability of the admin area a lot. Please change the PR to remove the colors again.
Thanks and best regards, mg

s7design:
Hi @mgruner 

I have updated this PR. If you agree with me, it could look like it in the other Admin screens and colleagues can continue to do on it.

Regards
Zoran

var/httpd/htdocs/skins/Agent/default/css/Core.Default.css
```diff
@@ -967,8 +967,8 @@ textarea.cke_dialog_ui_input_textarea {
     background-color: #ccc;
 }
 
-input.cke_dialog_ui_input_password:focus, 
-input.cke_dialog_ui_input_text:focus, 
+input.cke_dialog_ui_input_password:focus,
+input.cke_dialog_ui_input_text:focus,
 textarea.cke_dialog_ui_input_textarea:focus {
     box-shadow: none !important;
     border-color: #f92 !important;
```
s7design: These changes are from CodePolicy.
Regards
Zoran

mgruner:
Merged, thank you. I cleaned up some unneeded whitespace around the IF condition. If you need to append a class to others you can add the whitespace inside the IF: 
```class="...[% IF Data.ValidID != 1 %] Invalid[% END %]"```

s7design:
Thanks, I have got it now.
We are going to  continue to do on the other screens.

Regards
Zoran

PR #607, opened 2015-06-15
Add UnitTest::Helper methods for starting and rolling back transactions

moritz:
this way, tests that do not need to interact with the outside can
easily discard their own test objects, and avoid polluting the DB.

### DIFF
--- a/Kernel/System/UnitTest/Helper.pm
+++ b/Kernel/System/UnitTest/Helper.pm
@@ -17,6 +17,7 @@ use Kernel::System::SysConfig;
 our @ObjectDependencies = (
     'Kernel::Config',
     'Kernel::System::DB',
+    'Kernel::System::Cache',
     'Kernel::System::CustomerUser',
     'Kernel::System::Group',
     'Kernel::System::Main',
@@ -41,6 +42,8 @@ construct a helper object.
         'Kernel::System::UnitTest::Helper' => {
             RestoreSystemConfiguration => 1,        # optional, save ZZZAuto.pm
                                                     # and restore it in the destructor
+            RestoreDatabase            => 1,        # runs the test in a transaction,
+                                                    # and roll it back in the destructor
         },
     );
     my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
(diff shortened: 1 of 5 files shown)


moritz:
I'm also happy to report that adding the BeginWork and Rollback to scripts/test/Ticket.t reduces the run time of that test for me from about 21s to 8s (mysql, InnoDB); probably because the DB engine isn't force to write anything to disk as long as the transaction is uncommitted.

mgruner:
Hello @moritz. The HelperObject only lives within one unit tests at maximum and can therefore be used as a scope guard. We already have a constructor parameter ```RestoreSystemConfiguration``` that is used to save the ZZZAuto.pm before the test and to restore it afterwards, even after a die(). We should use the same approach with another parameter like ```RollbackDBChanges```, ok?

moritz:
Yes, sounds fine.

moritz:
@mgruner I've Added an option `RestoreDatabase` in my latest commit (making it a bit easier to remeber, since the existing option also starts with `Restore`).

s7design:
Hi @moritz 

As I told you in email, it looks very nice. I believe it could improve testing and especially cleaning up after testing.
We are be able to update all Unit test when it will be merged, of course  in arrangement with @mgruner

:+1: 

Regards
Zoran

mgruner:
Hi @moritz. Thanks a lot, I merged this in d94451c without the changes to the other unit tests. I first want to see what your UT servers report about this change. I'm afraid that some still use MyISAM and then I would not be sure what to do as we cannot rely on the UT servers to rollback the things everywhere.

rolfschmidt:
looks really nice ++

pboguslawski:
Hi,

DB transactions in unit testing may interfere with DB transactions used by OTRS in the future (which seem to be required to avoid races), i.e, after applying mod

https://github.com/OTRS/otrs/pull/1571

unit test for ticket creation fails on MySQL/InnoDB with error

    Transaction isolation level can't be changed while a transaction is in progress[...]

because of trying to create nested transaction for ticket number generation.

Consider not using transactions for UT or at least adding option for otrs.Console.pl to run UT without transactions.

Regards,
Paweł

PR #608, opened 2015-06-15
Master - Changed output for invalid entities (Signature&Salutaion)

s7design:
Hi @mgruner 

As I  mentioned  today, there is PR with a changed output for invalid Signature and Salutation.
We changed a little it, I believe that it is better now. A class is added only if it is needed instead empty class as there is in AdminAttachment. I will reopen PR for AdminAttachment and change like there is in this one. 
Please let me know what do you think about it, we will be able to do as there is in AdminAttachment if you mean that it is necessary.

Regards
Zoran & Vladimir


mgruner:
Thanks. The solution in AdminAttachment is also ok, we don't have to change it.

PR #610, opened 2015-06-15
Master - Reset select in AgentPreferences

s7design:
Hi @mgruner and @ManuelHecht 

There is a task, which I mentioned in mails, that Vladimir and Aleksandar have done on training.
Actually, I gave them to add reset button to select boxes for a couple screen in OTRS. 
They had template how it should do in AdminGenericAgent screen, and on the same way they changed AgentPreferences.tt. You can see it.
Maybe if you think that it is interesting for OTRS, we can make PR for the other screens.

We did this on:

* AgentStats:
  * Subaction=View
  * Subaction=EditSpecification
  * Subaction=EditXaxis
  * Subaction=EditValueSeries
  * Subaction=EditRestrictions

* AdminSLA - Subaction=SLAEdit

Please let me know what do you mean about it.

P.S. We will make a few more PR  from our training tasks next days. Of course we will do on the task that Martin gave us for invalid entities in Admin screens.   

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -4493,6 +4493,9 @@ via the Preferences button after logging in.
                 <Description>Agent Preferences</Description>
                 <Title></Title>
                 <NavBarName>Preferences</NavBarName>
+                <Loader>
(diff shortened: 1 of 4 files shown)


s7design:
.

PR #611, opened 2015-06-15
Update Defaults.pm

crythias:
Update for template code.


mgruner:
Hello @crythias, what's wrong with the code? I believe that the current version should work, while yours should not. Can you provide some more information please?

crythias:
It does work as mentioned, but it works (for me?) as I posted as well. I'm suggesting it for conformity as there is bound to be a question whether to use $Env{variable} vs Env.variable.

Please discard/decline if you disagree. No worries either way.

mgruner:
Hello @crythias. Please have a look in http://otrs.github.io/doc/manual/developer/stable/en/html/TemplatingMechanism.html#env. Env() is actually a function that is supposed to be called in TT.
Best regards, mg

PR #612, opened 2015-06-16
Master - Changed output for invalid entities in Admin Customer Management block

s7design:
Hi @mgruner 

There are changed  Admin screens for Customer user and Customer Company.
I do that with Vladimir.
I also add changes in Selenium tests. I wanted to do this with our new colleagues. 
I am not sure if you want that we also do that in the other AdminModules when we will add Invalid class for overview tables.
Please let me know what do you think about that. For us it will not be a problem, we have a time for it.

Regards
Zoran & Vladimir

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
@@ -97,7 +97,7 @@
                             </tr>
 [% RenderBlockEnd("NoDataFoundMsg") %]
 [% RenderBlockStart("OverviewResultRow") %]
-                            <tr class="MasterAction">
+                            <tr class="MasterAction[% IF Data.ValidID != 1 %] Invalid[% END %]">
                                 <td><a class="AsBlock MasterActionLink" href="[% Env("Baselink") %]Action=[% Env("Action") %];Subaction=Change;CustomerID=[% Data.CustomerID | uri %];Nav=[% Data.Nav | uri %]">[% Data.CustomerID | truncate(20) | html %]</a></td>
                                 <td>[% Data.CustomerCompanyName | html %]</td>
                                 <td title="[% Data.CustomerCompanyComment | html %]">[% Data.CustomerCompanyComment | truncate(26) | html %]</td>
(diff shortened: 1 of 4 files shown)


Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
```diff
@@ -97,7 +97,7 @@
                             </tr>
 [% RenderBlockEnd("NoDataFoundMsg") %]
 [% RenderBlockStart("OverviewResultRow") %]
-                            <tr class="MasterAction">
```
mgruner: I think in these files we also have to check if Data.ValidID actually contains something as it could come from an external database where this field is missing (ForeignDB => 1). Something like Data.ValidID && Data.ValidID != 1?

scripts/test/Selenium/Agent/Admin/AdminCustomerUser.t
```diff
@@ -147,15 +145,11 @@ $Selenium->RunTest(
...
-        $Selenium->find_element( "#Search", 'css' )->clear();
-        $Selenium->find_element( "#Search", 'css' )->send_keys($RandomID);
-        $Selenium->find_element( "#Search", 'css' )->submit();
-
         $Self->True(
             index( $Selenium->get_page_source(), $RandomID ) > -1,
             "$RandomID found on page",
         );
+
         $Self->False(
```
mgruner: What does this mean? You first check that RandomID2 is present, and then check that it is not present?

mgruner: I am also not sure if there could be problems if there are already many customer entries present in the db. That's why the filter was used here previously, I believe.

s7design: Thanks Martin, 
I saw now that we incidentally deleted "test search filter", with this part of code it have sense.
Because at first is checked if RandomID2 is present, and then after using filter by RandomID only this one is present, and RandomID2  is not.
We are doing on it, it will be fixed and committed soon.
Regards
Zoran

s7design:
Hi @mgruner 
I updated PR and also slightly improved tests.
Regards
Zoran

mgruner:
Merged, thanks a lot!

PR #613, opened 2015-06-16
Changed output for invalid entities in Admin Agent Management block

s7design:
Hi @mgruner,

We completed adding Invalid class Admin Agent screens.
As I said in previous PR, we will be able to update Selenium test for them as well, if you want.

Regards
Zoran & Aleksandar


mgruner:
Nice, thanks! Having a check in the selenium tests would be nice, too.

PR #617, opened 2015-06-17
Master - Selenum test, check Invalid class for overview screens

s7design:
Hi @mgruner 

As I had said before, I have updated Selenium test here. I added checking for Invalid class. There are some slight improvement in these tests as well.

Regards
Zoran


mgruner:
Merged, thank you very much.

PR #620, opened 2015-06-18
Master - Menu NavBar and MainMenu fixed for CIC

s7design:
Hi @mgruner 

I fixed CIC NavBar. I saw that there was a mistake in SysConfig module registration for CIC, and I solved that on this way. I hope it is better. 

As you can see I also fixed link for MainMenuItem ( AgentCustomerInformationCenter::MainMenu###010-EditCustomerID ) in CIC, doing on this I saw that link for this menu item is not good. There should be interpolated value, and now it works fine.

Regards
Zoran


mgruner:
Merged, thanks a lot!

PR #619, opened 2015-06-18
Master - Changed output for invalid entities in Admin Queue Settings block

s7design:

PR #621, opened 2015-06-18
Master - Added some info to make link 'AddAction' more specific for the Selenium tests

s7design:
Hi @mgruner 

We fixed the links for AddAction here. We mentioned this in conversation with Carlos, I believe it is better now, because if Agent had admin permissions there would be Stats/Add action, and the these tests would be failed.

If you like it you can marge it. I know that we make some PRs with updated some of them tests ( e.g. https://github.com/OTRS/otrs/pull/619, and others for "invalid" display ), and you could have conflicts during the merging. In that case I will be able to solve this depending witch PR is merged first.

If you have any suggestions, please let me know.

Regards
Zoran


mgruner:
Merged in dcc3e7ca5be06ec9d440f289abc82bcab1864a2a, thanks!

PR #623, opened 2015-06-19
Master - Improved selenium tests (AdminRole/Group display invalid entities)

s7design:
Hi @mgruner 

There is improved Selenium test for checking Invalid class
Doing on task for displaying invalid entities we saw that it could be better if we found more specific Invalid class. If there is filter it is OK, because it is checked for Invalid class after filtering. However if filter did not used, there could be found any other entity with Invalid class.

Regards
Zoran


mgruner:
Merged, thank you very much!

PR #626, opened 2015-06-22
Master - Changed output for invalid entities in System Administration  block 

s7design:
Hi @mgruner 

In this PR we changed display for invalid entities in System Administration admin block. As you can see we added invalid class for Process overview in ProcesManagment screen for inactive process. 
What do you think about it?
We also added Selenium test which test this feature but only in AdminProcessManagementActivity.t, if you want we can add it in other admin process tests.

And one more thing, there is not ValidID value for WebService in OverviewResultRow block of
AdminGenericInterfaceWebservice screen, there is only string for appropriate ValidID. We chacked valid string. The same are in AdminACL.tt (this one was merged to master) and AdminSystemMaintenance.tt.
And of course slight different check is for AdminProcessManagement.tt. I ask you this, because I like if there is the same realization for one feature and behavior in whole screen. It is not vital, but if you want we will be able to adapt all screens, but it could be stay like it is in this PR.

Please let me know what do you mean about all of that.

Regards
Zoran


mgruner:
Merged, thanks! I just changed AdminGenericAgent to not show the "Run now" link if the job is invalid.

s7design:
I like your change :+1: 
To be honest I commented with my colleagues about this, how it was enabled Run feature in case job is invalid. It is better UX now.

Regards
Zoran

PR #628, opened 2015-06-24
[Master - Bug 11112] Number of affected tickets of GenericAgent job via web interface is limited hardcoded

s7design:
Hi @mgruner 

http://bugs.otrs.org/show_bug.cgi?id=11112
Here is fixed bug for affected and executed tickets of GenericAgent job. We also added selenium test for this.

Regards
Zoran & Aleksandar


mgruner:
Merged with text improveements, thank you very much!

PR #629, opened 2015-06-25
[ Master - Bug 11325]  Dashboard TicketList: Queuefilter in an empty list shows inactive queues, too

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11325

Hi @carlosgarciac 

There is our proposal for fixing. I mentioned you in email another possible solution.
Maybe this one actually is the best one. We saw looking into code that there are two case for filling out filter data. The first one is "$ColumnName.FilterValuesGet()" in case there are TicketIDs and the second one is _GeneralDataGet in TicketIDs is not provided ( for empty table)

We saw that there is not unit test for ColumnFilter.pm, we will be able to do that if you want.

Regards
Zoran & Aleksandar

### DIFF
--- a/Kernel/System/Ticket/ColumnFilter.pm
+++ b/Kernel/System/Ticket/ColumnFilter.pm
@@ -971,9 +971,9 @@ sub _GeneralDataGet {
         return;
     }
 
-    # get data list
+    # get data list just for valid
     my %DataList = $BackendObject->$FuctionName(
-        Valid  => 0,
+        Valid  => 1,
         UserID => $Param{UserID},
     );
 
--- a/Kernel/System/Ticket/ColumnFilter.pm
+++ b/Kernel/System/Ticket/ColumnFilter.pm
@@ -149,7 +149,6 @@ sub QueueFilterValuesGet {
         return $Self->_GeneralDataGet(
             ModuleName   => 'Kernel::System::Queue',
             FunctionName => 'QueueList',
-            Valid        => 1,
             UserID       => $Param{UserID},
         );
     }
@@ -934,7 +933,6 @@ get data list
     my $Values = $ColumnFilterObject->_GeneralDataGet(
             ModuleName   => 'Kernel::System::Object',
             FunctionName => 'FunctionNameList',
-            Valid        => 0,     # optional, defaults to 0
             UserID       => $Param{UserID},
     );
 
@@ -995,11 +993,9 @@ sub _GeneralDataGet {
         return;
     }
 
-    my $Valid = $Param{Valid} // 0;
-
     # get data list
     my %DataList = $BackendObject->$FuctionName(
-        Valid  => $Valid,
+        Valid  => 1,
         UserID => $Param{UserID},
     );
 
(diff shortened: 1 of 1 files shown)


carlosgarciac:
Hi s7design,

Your proposal is fine for me, but I would like to handle a bit different, since the change on _GeneralDataGet will affect several places, I would like to include a new param (Valid) for this function, with a default value as '0', then we could use it on '1' from QueueFilterValuesGet in order to avoid invalid queues. I hope this is fine for you.

Best regards,
Carlos García

s7design:
Hi @carlosgarciac 

We updated the PR. We changed the function as you suggested.
Maybe it could be useful if a Valid parameter was used in ServiceFilterValuesGet and SLAFilterValuesGet functions as well. It is my conclusion after analysing possible cases.

Please let me know if you have any suggestions.

Regards
Zoran

carlosgarciac:
Hi S7design

I think we could give a bit more of flexibility here, including a sysconfig setting we could use on each call to '_GeneralDataGet',something like:

    <ConfigItem Name="TicketOverviewColumnFilter###ValidValues" Required="0" Valid="1">
        <Description Translatable="1">Defines 'Valid' values used for getting general data for column filters.</Description>
        <Group>Ticket</Group>
        <SubGroup>Frontend::Agent::Ticket::ColumnFilter</SubGroup>
        <Setting>
            <Hash>
                <Item Key="Queue">1</Item>
                <Item Key="Service">1</Item>
                <Item Key="SLA">1</Item>
                <Item Key="Others">0</Item>
            </Hash>
        </Setting>
    </ConfigItem>

(Please polish wording)

Then the administrator could decide the better approach for him. Sorry for too much changes with something look so simple, but I think this could be the best option here.

Please let me know your comments about this proposal.


Best regards,
Carlos García

mgruner:
Hello @s7design, @carlosgarciac! Just for my understanding, why should we include any invalid entities in these lists, so that we'd need a new config setting?

s7design:
Hi @carlosgarciac  and @mgruner 

I understand the both of you and we will be able to do some update if it will be necessary. 
I think that you have more experience and I accept your advice. Regarding to Martin's suggestion, current solution could be acceptable.  Please let me know if we should do some changes.

Regards
Zoran

carlosgarciac:
Hi Martin,

The reason I ask @s7design for including a new setting is because I have on mind that due some business logic on a company ( an invalid state for a ticket that company needs to move it out from that state but needs the review of an agent, and put this state as valid is not an option because could be selected on other places in the system, or if administrator doesn't want to receive new emails temporary on a queue but show ticket with this invalid value is required, and some other cases like these ) the administrator have the need to show invalid values, but just for some fields, then with this configuration administrator could take this decision, and is not anchored to this Valid => 0 value. @mgruner if you don't agree on this new config setting please just let us know and we could just change to Valid => 1 on _GeneralDataGet, thanks in advance for your answer.

mgruner:
Hello @carlosgarciac, thank you for the explanation. I see your point now. However, I feel that the filtering in the ticket overviews is not the right place to handle invalid entities, and this bug is also only about when the ticket list is empty, right? Besides that my goal for OTRS is not to have more and more config options, but less, and more simple and intelligent code. We can still implement your proposal if the future shows a real need for this.

s7design:
Hi all,

I have just updated PR. There is case where only valid entities are displayed in filter select boxes for empty lists for Dashboard screens. I hope this case is acceptable.
Please let me now if I can do something else.

Regards
Zoran

carlosgarciac:
Merged, thanks a lot @s7design.

Best regards,
Carlos García

PR #630, opened 2015-06-26
Linkchange

MBRG:
I replaced several occurances of the old "doc.otrs.org" with the new "http://otrs.github.io/doc/". 
I also replaced the no longer working "forums.otrs.org" with "forums.otterhub.org".


mgruner:
Merged in 60a77a9, thanks a lot!

PR #631, opened 2015-06-30
[Master - Bug 10984] Issues when add customer user by "Customer user" option

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10984

Hi,

There is fixed bug in master branch. We will provide the selenium tests for this as soon as possible. 

Regards,
Aleksandar


mrcbnsls:
Merged in 85bc38e, thanks. I decided to leave the check in the code to see if the radio button is actually selected but exchanged attr() with prop() which is working more consistant than attr().

PR #633, opened 2015-06-30
[Rel-4_0 Bug 11334] - If you don't have permission in a queue, you get an eMail Notification

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11334

Hi @carlosgarciac,

We fixed this bug, this is our proposal. We solve this in the SendAgentNotification function of Article.pm.
At first I tried to solve it in ArticleCreate but I believe that it is better to fix here. In this case all possible notification is covered.
We used your advice and we checked a permission  using “GetQueueGroupID” and GroupMemberList(), thanks for that it was useful.

Regards
Zoran

### DIFF
--- a/Kernel/System/Ticket/Article.pm
+++ b/Kernel/System/Ticket/Article.pm
@@ -2743,6 +2743,17 @@ sub SendAgentNotification {
         DynamicFields => 0,
     );
 
+    # get create permission groups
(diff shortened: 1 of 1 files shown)


carlosgarciac:
Merged, after review a bit more, I modify the solution, using 'TicketPermission' function that is better for this case,  I already confirm that with @mgruner , thanks a lot @s7design.

Best regards,
Carlos García

PR #634, opened 2015-07-01
Bugfix for Bug 10878: Bulk action does not lock tickets on owner-change

moritz:
if RequiredLock is set to False, setting the owner should lock the ticket


mgruner:
Merged. Thank you very much @moritz! :+1:

PR #637, opened 2015-07-06
Master - new statistics test

s7design:
Hi @mgruner 

I looked into new statistics module and Run test a little bit. I have advice for slight changing this test. 
I added test for Delete button, I believe it is better fronted test if there is checking this feature as well.

Aldo I changed names for tests in /Selenium/Agent/AgentStats/ directory. It is better now, like you add AgentStatistics and in them Run test.
I am planing to add more test for new statistics module (maybe in another PR or in this one).

What do you think about this, do you like it?

Regards
Zoran


mgruner:
Merged, thank you. I reactivated the return 1 in Run.t for phantom. You don't need to invest more time in AgentStats/ because the old GUI and tests will be dropped soon.

PR #638, opened 2015-07-07
Transliteration support for language translations update (sr_Latn)

dvuckovic:
Serbian language is a little bit special :) There are two alphabets (**Cyrillic** and **Latin**), and both are supported in OTRS (**sr_Cyrl** and **sr_Latn**). The problem is that they are treated as two different languages in OTRS, and translations have diverged over time (they are not in sync).

Luckily, there is a process that can automatically convert strings in one alphabet to another: it's called transliteration and it works best in Cyr2Lat direction. There is even an [ISO standard](https://en.wikipedia.org/wiki/ISO_9) for it, although slightly older version (from 1954) is being used. There is a convenient Perl library for translation ([Lingua::Translit](http://search.cpan.org/~alinke/Lingua-Translit-0.21/lib/Lingua/Translit.pm)), that lacked support for this specific standard, but it's modular architecture allowed for easy inclusion of additional conversion table especially for Serbian language ([ISO/R 9:1954](https://en.wikipedia.org/wiki/ISO_9#ISO.2FR_9)).

Pull request includes:
* patch for Console command **Dev::Tools::TranslationsUpdate** in order to automatically transliterate strings for supported languages (currently, only **sr_Cyrl** to **sr_Latn** conversion defined)
* modified CPAN library **Lingua::Translit** with *ISO/R 9:1954* transliteration table

If patch is accepted, recommendation is to remove the **sr_Latn** (*sr@latin*) language from **Transifex**, since this language variant can now be auto-generated from **sr_Cyrl** when it's updated.

### DIFF
--- a/COPYING-Third-Party
+++ b/COPYING-Third-Party
@@ -311,6 +311,15 @@ o Name:    CPAN Font::TTF
   License: # GNU Lesser v2.1 - http://www.gnu.org/licenses/lgpl.html
   Note:    # Copyright (c) 1998-2015, SIL International (http://www.sil.org).
 
+o Name:    CPAN Lingua::Translit
+  Web:     http://lingua-systems.com/translit/
+  License: # This module is free software. It may be used, redistributed and/or
+           # modified under the terms of either the GPL v2 or the Artistic license.
+           # GNU v2 - http://www.opensource.org/licenses/gpl-2.0.php
+           # Artistic v2 - http://www.opensource.org/licenses/artistic-license-2.0.php
+  Note:    # Copyright (C) 2007-2008 Alex Linke and Rona Linke.
+           # Copyright (C) 2009-2015 Lingua-Systems Software GmbH
+
 # --
 # JavaScript
 # --
--- a/Kernel/System/Console/Command/Dev/Tools/TranslationsUpdate.pm
+++ b/Kernel/System/Console/Command/Dev/Tools/TranslationsUpdate.pm
@@ -13,6 +13,7 @@ use warnings;
 
 use base qw(Kernel::System::Console::BaseCommand);
 
+use Lingua::Translit;
 use Pod::Strip;
 
 use Kernel::Language;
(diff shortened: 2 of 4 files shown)


s7design:
Nice, good job :+1: 

S7Design
Zoran

mgruner:
This looks very nice already. Please add the new dependency to Copying-ThirdParty. I'm also kindof uncomfortable with the patching, but at this point we can probably do nothing better. Let's hope you can get the changes upstream too.

dvuckovic:
Good news: Alex from Lingua Systems already accepted my change (just conversion table), and published [new version of the library](http://lingua-systems.com/translit/). No patching will be necessary, I will update the library files to new ones, add the dependency and push new commit.

mgruner:
Amazing!

Am 07.07.15 um 14:13 schrieb Dusan Vuckovic:
> Good news: Alex from Lingua Systems already accepted my change (just
> conversion table), and published new version of the library
> <http://lingua-systems.com/translit/>. No patching will be necessary, I
> will update the library files to new ones, add the dependency and push
> new commit.
> 
> —
> <https://github.com/OTRS/otrs/pull/638#issuecomment-119185830>.
> 

-- 
Martin Gruner
Senior Developer R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

Geschäftssitz: Bad Homburg, Amtsgericht: Bad Homburg, HRB 10751,
USt-Nr.: DE256610065
Aufsichtsratsvorsitzender: Burchard Steinbild, Vorstand: André
Mindermann (Vorsitzender), Christopher Kuhn, Sabine Riedel

Grenzenlose Kommunikation mit der OTRS Business Solution™! Bereichern
Sie Ihren IT Support und Kundenservice mit smarten Business Features!
https://www.otrs.com/solutions/otrs-business-solution-on-premise/?lang=de/#OTRS-Business-Solution-Basic-Features

mgruner:
Thanks a lot! I'm keeping the language in Transifex for now, we should delete it after the switch to OTRS 5.

PR #640, opened 2015-07-07
Master -  new statistic tests

s7design:
Hi @mgruner,

There are two Selenium test for Import and Overview feature for new statistics.
I am planning to add more test ( e.g. Add, Export ...)

I have a problem in edit screen ( Action=AgentStatistics;Subaction=Edit; ). You can see in Import.t, I wanted to edit imported statistics. When I tested X-axis, Y-axis and Data Restrictions dialog for edit, I had a problem if I wanted to add some element. I can select only the firsts elements.
dialog_issue_in_tests

If I want to select for example Priority in case from image above, test will be failed. There is behavior like it is clicked outside of dialog. Because of that I selected the first element in the select box (Service, SLA) and test is successful. Of course, the problem is only with Selenium tests, you know that this dialog work property if it is tested manually.
I am informing you about this due to I wanted to make better test case for testing Add feature. I am hope that I managed to explain what the problem is. If you have any idea please tell me and I will be able to improve Import.t test.

Regards
Zoran


s7design:
Hi @mgruner 

The problem I mentioned in the previous comment occur only in firebox. I can select all options in select box with phantomjs.

Regards
Zoran

s7design:
I added in the last commit version of Import.t that works well only in phnatomjs If it would work in Firefox it could be able to test much better these dialogs for edit statistics.

Regards
Zoran

s7design:
Hi @mgruner 

I complied all test for new statistics module. I also cleaned code in Run.t test.
If you have some idea for more Selenium tests or if you have suggestions for improvement in these tests, please let me know.

Regards
Zoran

mgruner:
Hello Zoran, awesome! Merged. Works just fine for me both in phantomjs and also Selenium. Do you still have issues in selenium? If so, you could try clicking with jquery ( ```$('option[value=...]).click()``` ), but after that you have to do manual waiting. Clicking with webdriver is better because it waits for the form submit. Thanks a lot!
Regards, mg

s7design:
Thanks, now with this test all works fine. I had a problem with another test case that I wanted to test. 
I wanted to test more details, but I believe that I managed to find enough good test case in this moment
But in this one all is ok. 

Regards
Zoran

PR #641, opened 2015-07-09
Fixed #11295: 'No such user' notification while otrs automatically responses…

acran:
This should fix [#11295](http://bugs.otrs.org/show_bug.cgi?id=11295) which was introduced in 203084b771be5d98ef8ea4aa87427e12a04edb58 (release 4.0.8)

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -581,7 +581,7 @@ sub AutoResponse {
     }
 
     # get recipient
-    my %User = $Kernel::OM->Get('Kernel::System::CustomerUser')->GetPreferences(
+    my %User = $Kernel::OM->Get('Kernel::System::CustomerUser')->CustomerUserDataGet(
         UserID => $Ticket{CustomerUserID},
     );
 


carlosfrodriguez:
Hi @acran,

Thank you for your proposal, can you please print, sign and send us back the CLA as stated in http://otrs.github.io ?

I will appreciate it a lot

acran:
@carlosfrodriguez of course, just did so and you should have received it by now

Kernel/System/TemplateGenerator.pm
```diff
@@ -581,7 +581,7 @@ sub AutoResponse {
     }
 
     # get recipient
-    my %User = $Kernel::OM->Get('Kernel::System::CustomerUser')->GetPreferences(
+    my %User = $Kernel::OM->Get('Kernel::System::CustomerUser')->CustomerUserDataGet(
         UserID => $Ticket{CustomerUserID},
```
carlosfrodriguez: User instead of UserID is needed, otherwise it return false silently and language is never seem then english is always used

((enjoy))
cr

carlosfrodriguez:
Hi @acran,

This pull request has been merged in https://github.com/OTRS/otrs/commit/28338c0b4dcc6693f7cd2a1460dd7a5f718b5471, with an small change in the function parameter, additionally I create a unit test to be sure that this issue an also the language issue are fixed. This fix was also back ported to OTRS 4.

Usually if a bug is detected in a release branch (e.g. rel-4_0) the pull request should be done for that release, (additionally a pull request for master can be done and it is very welcome, specially if the fix is different due to the changes in the code).

Thank you for all your help.

((enjoy))
cr

PR #642, opened 2015-07-10
Genericinterface pendingtime diff

fredericve:
With this change you can perform a request towards GenericInterface with the PendingTime->{Diff} attribute set to a numeric value.

This allows for easier pending time setting for pending states. Kernel/System/Ticket.pm already supports this. I'm just adding it for GenericInterface where it always required the <year><month><day><hour><minute> attributes before.

### DIFF
--- a/Kernel/GenericInterface/Operation/Ticket/Common.pm
+++ b/Kernel/GenericInterface/Operation/Ticket/Common.pm
@@ -735,7 +735,7 @@ sub ValidatePendingTime {
 
     # if only the Diff attribute is present, check if it's a valid number and return.
     # Nothing else needs to be checked in that case.
-    if (keys %{ $Param{PendingTime} } == 1 && exists('Diff', $Param{PendingTime})) {
+    if ( keys %{ $Param{PendingTime} } == 1 && exists( $Param{PendingTime}->{Diff} ) ) {
         return if $Param{PendingTime}->{Diff} !~ /^[0-9]+$/;
         return 1;
     }
--- a/Kernel/GenericInterface/Operation/Ticket/Common.pm
+++ b/Kernel/GenericInterface/Operation/Ticket/Common.pm
@@ -715,6 +715,12 @@ checks if the given pending time is valid.
         },
     );
 
+    my $Success = $CommonObject->ValidatePendingTime(
+        PendingTime => {
+            Diff => 10080,
+        },
+    );
+
     returns
     $Success = 1            # or 0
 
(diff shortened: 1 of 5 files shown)


carlosfrodriguez:
Hi @fredericve,

Thank you for your proposal, this is in deed interesting.

Could it be possible for you to also update TicketCreate and TicketUpdate operations POD, also the WSDL has to be updated.

((enjoy))
cr

Kernel/GenericInterface/Operation/Ticket/Common.pm
```diff
@@ -727,6 +733,13 @@ sub ValidatePendingTime {
     return if !$Param{PendingTime};
     return if !IsHashRefWithData( $Param{PendingTime} );
 
+    # if only the Diff attribute is present, check if it's a valid number and return.
+    # Nothing else needs to be checked in that case.
+    if ( keys %{ $Param{PendingTime} } == 1 && exists( $Param{PendingTime}->{Diff} ) ) {
```
carlosfrodriguez: Why use exists and not defined?

Provably you can use defined and you can remove the ()

:-)

((enjoy))
cr

Kernel/GenericInterface/Operation/Ticket/Common.pm
```diff
@@ -727,6 +733,13 @@ sub ValidatePendingTime {
...
+        return if $Param{PendingTime}->{Diff} !~ /^[0-9]+$/;
```
carlosfrodriguez: Might be better:
```perl
 return if $Param{PendingTime}->{Diff} !~ m{\A \d+ \z}msx;
```

((enjoy))
cr

Kernel/GenericInterface/Operation/Ticket/Common.pm
fredericve: Hello @carlosfrodriguez,

Thanks for the feedback!

From the perl defined() documentation:
When used on a hash element, it tells you whether the value is defined, not whether the key exists in the hash. Use exists for the latter purpose.

Hence why I used exists.

fredericve:
Hello @carlosfrodriguez,

I updated the files you suggested as well.

Tested all the changes in our UAT environment.

Any other suggestions are welcomed!

Kernel/GenericInterface/Operation/Ticket/Common.pm
```diff
@@ -727,6 +733,13 @@ sub ValidatePendingTime {
     return if !$Param{PendingTime};
     return if !IsHashRefWithData( $Param{PendingTime} );
 
+    # if only the Diff attribute is present, check if it's a valid number and return.
+    # Nothing else needs to be checked in that case.
+    if ( keys %{ $Param{PendingTime} } == 1 && exists $Param{PendingTime}->{Diff} ) {
```
carlosfrodriguez: Hi @fredericve,

The problem with exists is that it return success even if you have Diff => undef, as the key exists in the hash, no matter if the value is undefined, then that is why I believe that what you really really what is to make sure that you have a value in the key even if it is 0, then the best option is to use defined.

Does this makes sense?

((enjoy))
cr

fredericve: Hello @carlosfrodriguez 

Agreed! I have updated the pull request.

Cheers,

Frederic

carlosfrodriguez:
Hi @fredericve,

I've merged your feature in https://github.com/OTRS/otrs/commit/273d6f3bf735095b6148373afd722821ffd40a67, I just add an extra check to forbid the use of Diff together with other keys and I add a couple of unit tests.

Thanks again for your contribution!

((enjoy))
cr

PR #645, opened 2015-07-13
Bugfix in Bulk actions: preserve Unlock option

moritz:
`$Param{Unlock}` can be 1 or 0, so `$Param{Unlock} || 1` is always 1.

This leads to the "Unlock Tickets" choice always being set to "Yes".


mgruner:
Merged in c070453 with a slight addition: LinkTogether suffered from the same problem, just a few lines above. Thanks! :+1:

PR #647, opened 2015-07-16
Convert rpm spec description to UTF-8

vinzent:
The Pulp repository manager (used in Foreman/Katello/Satellite6) rejects RPM's if fields have non-utf8 characters.


vinzent:
seems the spec generation was reworked in master. so its only needed on 4_0.

mgruner:
Applied, thanks a lot @vinzent! :+1: 
For future submissions, please consider signing the CLA as described at https://otrs.github.io.

PR #648, opened 2015-07-17
Master - fixed print output for search screens

s7design:
Hi @mgruner ,

Here is fixed print output for CustomerTicketSearch and cleaned code for some other modules and tests.
I went trough all modules. I believe that all modules is cleaned now and there is the same print output for all modules.

Regards
Zoran


mgruner:
Hello Zoran,
thanks a lot for your proposal. Right now I see a problem that the PDF outputs more information than the previous HTML print output. This is a problem because customers may not see all information. The logic should work in the same way as before to collect the data for the PDF. Could you amend that please?
Best regards, mg

s7design:
Hi,

Thanks @mgruner, yes of course I will be able to update this soon

Regards
Zoran

s7design:
Hi @mgruner 

I updated this PR previous days. I deleted owner info for customer print view. Please let me know if I can change anything else.

Regards
Zoran

mgruner:
Merged, thanks! I added some code to customer search print to open in a new tab if Print is selected for better usability, please let me know if you see any problems with this.

PR #649, opened 2015-07-17
Handle /var/lock/subsys/otrs in redhat init script

vinzent:
RedHat <= 6 init scripts need to create a lockfile in /var/lock/subsys or the
init script won't be run on shutdown.

Another requirement is, that one should get more appropriate return codes when
calling the status command. For example puppet only starts the service if it's
not already started (very sane) but right now status returns 0 ("program is
running or service is OK") if start was called or not and in turn does not
enable the cron jobs.


mgruner:
Thanks! Please consider signing the OTRS CLA as described on https://otrs.github.io if you plan to make more contributions. :+1:

PR #650, opened 2015-07-20
Allow selecting children of matched nodes during tree search

dvuckovic:
Same as previous behavior in OTRS4. Solved by just activating parameter in jsTree search plugin.


mgruner:
Thanks!

PR #651, opened 2015-07-21
Micro-optimize FilenameCleanup in Kernel::System::Main

moritz:
The file system backend of Kernel::System::Cache uses it,
so it is a pretty hot path. In a profiling of a ticket
listing, this patch reduces the total run time of
FilenameCleanup from 34ms to 2.4ms, and the overall run
time of the request by 3%


mgruner:
:+1

PR #652, opened 2015-07-21
Input fields update

dvuckovic:
Following issues have been addressed:
* prevent removing selections in disabled forms
* prevent removing last selection in PossibleNone=0 single select fields
* modernized fields in CustomerTicketMessage
* RTL tweaks
* Agent Preferences fixes
* Owner field in AgentTicketMove screen (filter)

Please note that select options with empty values are automatically filtered by jsTree class. Because of this, we must make sure to always have defined values for option tags and use empty values only for undetermined selection (-). If server-side checks later depend on false values, simple 0 (zero) will be enough to keep them working. I already fixed CreateNextMask in Agent preferences, since it was using empty value as default.


mgruner:
bildschirmfoto 2015-07-21 um 13 47 17

RTL seems still not correct for me.

mgruner:
If I disable the preferences form via Core.Form.DisableForm(), I can still change the language field.

mgruner:
After activating the previous owner filter in AgentTicketMove which returns an empty set for my ticket, the field is emptied and I cannot click into it to deactivate the filter any more. That should be fixed; but empty filters should also perhaps be ignored / unselectable?

mgruner:
@dvuckovic, thanks a lot for your changes so far. Could you look at the issues mentioned above please? I'd merge your PR after they are addressed.

mgruner:
CustomerTicketMessage looks really nice. I have a strange effect though: when I remove the Queue, validation kicks in to tell me that this field is required (even though I did not try to submit yet). After a short time (<1s) the error is removed again.

dvuckovic:
Most things were fixed:
* RTL in preferences screen: since widths were resolution based, I had to make them smaller. This is not an ideal solution, but I don't have a better ATM.
* Disabling fields should be solved now. The reason is that there is no event to hook on this, I still have to do the manual check on focus. It might still leave focus ring, but will prevent from changing the selection.
* Empty filter selection is disabled, but Perl was sending two empty selections for some reason. Should be fixed now.

Not fixed:
* This auto validation in forms is active for other fields too (try entering subject, then clearing it). Since there is no difference between this validation and that on submit, I'm not sure how to check for it? Maybe some change in Validate class? Please advise.

mgruner:
Merged, thank you very much, @dvuckovic.
- Disabled filters are nice.
- Regarding validation in customer interface we should talk to Marc.
- Strange effects with disabled forms that you might want to check:
  - I set some queues in my queues and reload.
  - I ran Core.Form.DisableForm(). Looks ok. If I click "MyQueues" the appearance changes (elements narrower.
  - I ran Core.Form.EnableForm() again. Now all fields that I clicked during disabled state are not functional any more, the others still work. :smile: 

Thanks!

mgruner:
Hello @dvuckovic, I rolled out the new fields for the rest of the customer interface. If you look at the ticket search screen there, you'll see that the multiple selection fields are broken (button bar). Could you review this please?

PR #653, opened 2015-07-22
Input fields update2

dvuckovic:
Addressed issues:
* Increased minimum width of fields back to 200px
* Additional RTL tweaks
* CustomerTicketSearch tweaks
* Reworked disabling form logic (Publish/Subscribe)
* Tweaked disabled styles for selection boxes


mgruner:
I still have the problem that after DisableForm() and EnableForm() all fields which I clicked in the meantime are dysfunctional (Chrome).

dvuckovic:
Hm, this should be fixed, completely new approach has been implemented. I just tried in Firefox/Chrome/IE, please check if there might be a caching problem with some of the scripts (Loader?).

mgruner:
Yep, works fine now, some caching problem. Thanks a lot! RTL is also better, customer ticket search fixed too. :+1:

mgruner:
(Merged)

PR #654, opened 2015-07-22
Fix action highlight behavior in certain situations

dvuckovic:
Currently, if user hovers the cursor right on the edge of a ticket action (where orange highlight is supposed to show), browser will enter a perpetual state of activating and deactivating hover mode, like in the animation below:

Weird

Behavior has been confirmed in all major browsers (Firefox/Chrome/IE), and will continue as long as user does not move the mouse from position that activated the highlight.

This is because border is applied only to hover state, and does not exist in normal state (padding is applied/removed instead). It's much better practice to define transparent border in normal state, and just color it on hover (exactly what this fix does).

### DIFF
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Agent.TicketMenuModuleCluster.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Agent.TicketMenuModuleCluster.css
@@ -77,7 +77,7 @@
 }
 
 .Cluster ul ul li a:hover {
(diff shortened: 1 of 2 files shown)


mgruner:
@mrcbnsls could you please review this?

mgruner:
@mrcbnsls: ping, could you check this small change please?

mrcbnsls:
Merged in aa9ed70, thanks.

PR #655, opened 2015-07-23
Fix left offset handling in Agent and Customer interface; Turn off Dialog overflow hiding

dvuckovic:
Left offset for the drop downs should be fixed now. I also had to turn off *overflow: hidden;* directive in Dialog CSS, since it hides our drop downs if dialog is too narrow. Apparently, this rule didn't have any impact on the look of the dialogs across the pages, but I could be wrong.

@mgruner Do you have a list of all places where you implemented Input Fields? I would really like to check every use case personally, since there might be some hidden problems in border line cases. Thanks!

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -817,8 +817,12 @@ Core.UI.InputFields = (function (TargetNS) {
                     // Create main container
                     $ContainerObj = $('<div />').insertAfter($InputContainerObj);
                     $ContainerObj.addClass('InputField_Container')
-                        .attr('tabindex', '-1')
-                        .css('left', $SearchObj.offset().left + 'px');
+                        .attr('tabindex', '-1');
+
+                    // Set left offset if in Customer interface
+                    if (Core.Debug.CheckDependency('Core.UI.InputFields', 'Core.Customer', 'Core.Customer', true)) {
+                        $ContainerObj.css('left', $SearchObj.offset().left + 'px');
+                    }
 
                     // Create container for jsTree code
                     $TreeContainerObj = $('<div />').appendTo($ContainerObj);
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -820,7 +820,7 @@ Core.UI.InputFields = (function (TargetNS) {
                         .attr('tabindex', '-1');
 
                     // Set left offset if in Customer interface
-                    if (Core.Debug.CheckDependency('Core.UI.InputFields', 'Core.Customer', 'Core.Customer', true)) {
+                    if (Core.Customer) {
                         $ContainerObj.css('left', $SearchObj.offset().left + 'px');
                     }
 
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Dialog.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Dialog.css
@@ -73,7 +73,6 @@
 }
 
 .Dialog > .Content {
-    overflow: hidden;
     position: relative;
 }
 
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css
@@ -207,13 +207,13 @@ div.InputField_ToolbarContainer ul li a:hover {
  * @subsection  jsTree theme
  */
 .jstree-InputField {
-    font-size: 11px !important;
-    overflow-x: hidden !important;
-    border: none !important;
-    box-shadow: none !important;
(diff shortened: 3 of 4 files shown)


var/httpd/htdocs/js/Core.UI.InputFields.js
```diff
@@ -817,8 +817,12 @@ Core.UI.InputFields = (function (TargetNS) {
                     // Create main container
                     $ContainerObj = $('<div />').insertAfter($InputContainerObj);
                     $ContainerObj.addClass('InputField_Container')
-                        .attr('tabindex', '-1')
-                        .css('left', $SearchObj.offset().left + 'px');
+                        .attr('tabindex', '-1');
+
+                    // Set left offset if in Customer interface
+                    if (Core.Debug.CheckDependency('Core.UI.InputFields', 'Core.Customer', 'Core.Customer', true)) {
```
mgruner: I guess that ```if (Core.Customer) {} ``` would be enough, as this isn't a real dependency check, what do you think?

var/httpd/htdocs/js/Core.UI.InputFields.js
```diff
@@ -817,8 +817,12 @@ Core.UI.InputFields = (function (TargetNS) {
...
+                        $ContainerObj.css('left', $SearchObj.offset().left + 'px');
```
mgruner: What about RTL here, would it be different? Maybe check for class RTL on body?

var/httpd/htdocs/skins/Agent/default/css/Core.Dialog.css
```diff
@@ -73,7 +73,6 @@
 }
 
 .Dialog > .Content {
-    overflow: hidden;
```
mgruner: @mrcbnsls: Can you say something about this one? I think that it is there on purpose.

mgruner:
@dvuckovic I rolled out the fields in the customer interface (all screens), the agent interface (all screens, hopefully) and parts of the admin interface so far. Other admin screens might follow in the beta period, maybe we can get everything.

var/httpd/htdocs/js/Core.UI.InputFields.js
dvuckovic: You are right, how stupid of me. Why make it simple when it can be complicated? :)

var/httpd/htdocs/js/Core.UI.InputFields.js
dvuckovic: No need for RTL check, since positioning is absolute in this case.

var/httpd/htdocs/js/Core.UI.InputFields.js
mgruner: No problem. In German we say sometimes you "don't see the wood because
of all the trees". ;)

Am 23.07.15 um 14:10 schrieb Dusan Vuckovic:
> In var/httpd/htdocs/js/Core.UI.InputFields.js
> <https://github.com/OTRS/otrs/pull/655#discussion_r35314387>:
> 
>> @@ -817,8 +817,12 @@ Core.UI.InputFields = (function (TargetNS) {
>>                      // Create main container
>>                      $ContainerObj = $('<div />').insertAfter($InputContainerObj);
>>                      $ContainerObj.addClass('InputField_Container')
>> -                        .attr('tabindex', '-1')
>> -                        .css('left', $SearchObj.offset().left + 'px');
>> +                        .attr('tabindex', '-1');
>> +
>> +                    // Set left offset if in Customer interface
>> +                    if (Core.Debug.CheckDependency('Core.UI.InputFields', 'Core.Customer', 'Core.Customer', true)) {
> 
> You are right, how stupid of me. Why make it simple when it can be
> complicated? :)
> 
> —
> <https://github.com/OTRS/otrs/pull/655/files#r35314387>.
> 

-- 
Martin Gruner
Senior Developer R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

Geschäftssitz: Bad Homburg, Amtsgericht: Bad Homburg, HRB 10751,
USt-Nr.: DE256610065
Aufsichtsratsvorsitzender: Burchard Steinbild, Vorstand: André
Mindermann (Vorsitzender), Christopher Kuhn, Sabine Riedel

Grenzenlose Kommunikation mit der OTRS Business Solution™! Bereichern
Sie Ihren IT Support und Kundenservice mit smarten Business Features!
https://www.otrs.com/solutions/otrs-business-solution-on-premise/?lang=de/#OTRS-Business-Solution-Basic-Features

var/httpd/htdocs/skins/Agent/default/css/Core.Dialog.css
mrcbnsls: Hm, not offhand... @zottto, any objection on this change?

zottto:
Merged with 3d1d6fc.

PR #657, opened 2015-07-30
Update de.pm

Retiebralanak:
It only deletes one Process, not all
- delete inactive process gives both choices
- Inaktive Prozesse löschen -> alle
 Inactiven Process löschen -> nur den einen


mgruner:
Hello @Retiebralanak, thanks for your suggestion. Translations are managed on transifex, can I ask you to register and fix the issue there please? https://www.transifex.com/otrs/OTRS/
Some more detail can be found at http://otrs.github.io/doc/manual/developer/5.0/en/html/localization-translation-mechanism.html.

PR #658, opened 2015-07-30
memory optimized 'split(...)' using look behind and only create Email…

krumholt:
If SMIME is enabled every message (SMIME or not) is split in lines and every line copied in memory.
This takes a long time and consumes a lot of memory for huge messages with attachments.
Also EmailParser is created for every message. Even if not necessary.
This patch improves the performance by using look behind regexp and creating the parser just when needed

### DIFF
--- a/Kernel/Output/HTML/ArticleCheckSMIME.pm
+++ b/Kernel/Output/HTML/ArticleCheckSMIME.pm
@@ -101,17 +101,6 @@ sub Check {
             UserID    => $Self->{UserID},
         );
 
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @krumholt, thanks a lot for your suggestion. Could you please sign our CLA as described on https://otrs.github.io please?

mgruner:
Hello @krumholt, so far your CLA was not provided, therefore I have to close this PR for now. I'd be happy to revisit it with your signed CLA. Regards, Martin

PR #660, opened 2015-07-31
Updated description of SysConfig option 'PostMasterReconnectMessage'

eliasp:
The old description referenced the long gone/renamed
`bin/PostMasterMailAccount.pl` script and was also overly specific, as
this option is used by the different PostMaster modules
(IMAP/IMAPS/POP3/…), but not by the PostMaster scripts itself.


mgruner:
Very nice, thank you!

PR #662, opened 2015-08-03
Image Class & Image Preferences

dvuckovic:
Introduced image class to OTRS framework, for storing and retrieving images from VirtualFS, with additional support for image manipulation. Currently, only image resizing is implemented, along with support for 3 major Perl image library backends:
* **PerlMagick / ImageMagick** (both via Perl wrapper and command line tool)
* **Imager**
* **GD**

Based on the image class, image preference module is introduced for uploading and storing images as user preference. Sample configuration [here](https://gist.github.com/dvuckovic/6d61c20f4c987bed9bce).

image-preferences


jensoliver:
Nice one :-)

s7design:
Good job :+1:

mgruner:
Hello @dvuckovic. At this point I would not like to merge this to the framework (yet), because there is no immediate use case. The PR will stay in the archive for future reference. Thanks a lot!

PR #663, opened 2015-08-03
Add check perl-version to CheckModules.pl script

dshmelev:
Module IO/Interactive.pm required perl-version, but not checked in CheckModules.pl. Added check of perl-version module to CheckModules.pl script.


carlosfrodriguez:
Hi @dshmelev,

Thank you for your pull request, but it seams to be based in master and request to merge en rel-4_0, please rebase and create a new pull request to the correct branch (in this case master)

((enjoy))
cr

PR #666, opened 2015-08-05
BugFix: Add unchecked modules to otrs.CheckModules.pl.

dshmelev:


carlosfrodriguez:
Hi @dshmelev,

thank you for your pull request, now it is more clear, I have some questions about this:

Looks like Sys::Syslog is part of the core modules  [SysLog](http://perldoc.perl.org/Sys/Syslog.html), then it is supposed to be installed with your perl distribution, did you had any problem and had to install manually?

For version I'm not sure, can you tell me if you had any issue with it? do you find any place where it explicitly require it and you don't had it?

((enjoy))
cr

dshmelev:
Hi @carlosfrodriguez 

I'm using CentOS 7. And yum packet perl-5.16.3-285.el7.x86_64 does not contain Syslog.pm:

[root@web avikez]# rpm -ql perl-5.16.3-285.el7.x86_64 | grep Syslog.pm
[root@web avikez]# rpm -ql perl-Sys-Syslog-0.33-3.el7.x86_64 | grep Syslog.pm
/usr/lib64/perl5/vendor_perl/Sys/Syslog.pm

mgruner:
These are part of perlcore. The rpm has a dependency on perl-core which *should* install the modules.

dshmelev:
@mgruner You are right!
[avikez@web ~]$ sudo yum deplist perl-core | grep Syslog
  dependency: perl-Sys-Syslog
   provider: perl-Sys-Syslog.x86_64 0.33-3.el7

But you need to update documentation (https://otrs.github.io/doc/manual/admin/stable/en/html/manual-installation-of-otrs.html or https://otrs.github.io/doc/manual/admin/stable/en/html/otrs.html#software-requirements). There is no word about perl-core.

mgruner:
I'll add some hints about perlcore to the docs for OTRS 5. Thanks!

PR #668, opened 2015-08-06
Set contents of textareas translatable

urbalazs:
The contents of textareas contain English strings, which is shown at the UI. This patch makes them translatable.


mgruner:
Merged in cd88115, thanks a lot! I applied a small change to DaemonCheck.pm to not include the leading ' ' for translation.

PR #672, opened 2015-08-12
TransitionAction->TickeArticleCreate: When a new article is created, …

balazj:
TransitionAction - TicketArticleCreate
Right now there is no possibility to set the "From" field of the article to the user that made change. Process creator can set this field to fixed value.

This patch allows tracking who created this article without going to the ticket history. If "From" field is not assigned, it will use current user data.


mgruner:
Hi @carlosfrodriguez, can you please review and merge this?

carlosfrodriguez:
Hi @balazj,

Thank you for this pull request, it was merged in https://github.com/OTRS/otrs/commit/f3a56ba7ccadd48eea7eff0b2fbe17251a35fac9.

((enjoy))
cr

PR #673, opened 2015-08-13
Master - Fixed build selections for Select and MultiSelect widget elements in Statistics

s7design:
Hi @carlosfrodriguez 

Working on TimeAccounting I saw that there are some issues. I felt free to fix them here.

* There was problem because all attributes were multiselect. It is not good for example for 'OrderBy', 'SortSequence' or 'Limit'. 
* The second problem was with sorting. There was sorting always by Value _ColumnAndRowTranslation() no mater which  type is selected in SortSequence  ('ascending' or 'descending',) 

There maybe should clean up some code. For example I think that it is not needed text "Please select only one element or turn off the button 'Fixed'." in /Statistics/RestrictionsWidget.tt, and some other. We can discus about it.

Please let me know what do you think about it.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Statistics/View.pm
+++ b/Kernel/Output/HTML/Statistics/View.pm
@@ -758,6 +758,7 @@ sub XAxisWidget {
             }
         }
 
+        my $Multiple = ( $ObjectAttribute->{Block} eq 'MultiSelectField' ) ? 1 : 0;
         if ( $ObjectAttribute->{Block} eq 'SelectField' ) {
             $ObjectAttribute->{Block} = 'MultiSelectField';
         }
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @mgruner,

Can you please take a look on this?

((enjoy))
cr

mgruner:
Hello Zoran,
I am afraid that the code is correct here for the select fields. Even single selects are presented as multi select fields because if the field can be changed by the one running the statistic, they can select from the different values that are offered here. So it must be a multiselect field.
I don't exactly understand what problem you see with the sorting. SortSequence is just one regular field in statistic object. It should control the generated output, nothing in the frontend.
I'm closing this for now. Please feel free to provide more information if you want to continue this.
Regards, mg

PR #675, opened 2015-08-17
Handle empty value correctly when removing selection

dvuckovic:

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -264,7 +264,7 @@ Core.UI.InputFields = (function (TargetNS) {
                                 .attr('role', 'button')
                                 .attr('aria-label', Core.Config.Get('InputFieldsRemoveSelection') + ': ' + Text)
                                 .off('click.InputField').on('click.InputField', function () {
-                                    var Empty = $SelectObj.find('option[value=""]:contains(-)').length === 0 ? false : true,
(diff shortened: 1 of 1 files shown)


var/httpd/htdocs/js/Core.UI.InputFields.js
```diff
@@ -264,10 +264,16 @@ Core.UI.InputFields = (function (TargetNS) {
                                 .attr('role', 'button')
                                 .attr('aria-label', Core.Config.Get('InputFieldsRemoveSelection') + ': ' + Text)
                                 .off('click.InputField').on('click.InputField', function () {
-                                    var SelectedValue = $(this).parents('.InputField_Selection')
-                                        .data('value');
+                                    var Empty = $SelectObj.find('option[value=""]:contains(-)').length === 0 ? false : true,
```
mgruner: Hello @dvuckovic, do you think that we really need the ```:contains``` here? IMHO this might be superfluous.

mgruner: Besides that, maybe the variable name could be more speaking. ```HasEmptyElement``` or the like?

dvuckovic: @mgruner No problem, I'll change it. This currently forced check for both key and value (`'' => '-'`), but as you stated, it might be too much.

mgruner:
Looks very nice, thanks a lot @dvuckovic!

PR #676, opened 2015-08-17
Added 'note' permission to RoleLink.pm

MBRG:
Added 'note' permission to RoleLink.pm to allow adding note permissions via Console.pl


mgruner:
Thanks a lot! :+1:

PR #677, opened 2015-08-19
Fixes for Input Fields

dvuckovic:


mgruner:
Nice, thank you very much @dvuckovic.

PR #678, opened 2015-08-20
Fixes for Input Fields

dvuckovic:


var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css
```diff
@@ -45,6 +45,10 @@ div.InputField_InputContainer {
     color: #aaa;
 }
 
+div.InputField_InputContainer .InputField_Search {
+    margin-left: 0;
```
mgruner: What about margin-left, should that be included too (RTL)? Or maybe all margins?

mgruner:
The escaping does not yet work completely for me. In the field "*Response name scheme:" I always see an entry that I cannot remove, it will reappear after saving.

var/httpd/htdocs/skins/Agent/default/css/Core.InputFields.css
dvuckovic: You are right, I completely forgot about RTL. It's easiest to set all margin, I'll do it.

dvuckovic:
@mgruner I don't think this is related to escaping. The problem lies in empty keys, they are being ignored for jsTree control: https://github.com/OTRS/otrs/blob/master/var/httpd/htdocs/js/Core.UI.TreeSelection.js#L113

If you look closely, you won't find values in the list which have empty keys. This is because empty values are mostly used as placeholders for no selection (-).

Maybe it's better approach to revise these key=>value maps to include non-empty keys: https://github.com/OTRS/otrs/blob/master/Kernel/Modules/AdminGenericInterfaceTransportHTTPSOAP.pm#L332

mgruner:
Merged, thank you! We'll check about the empty value issues.

PR #680, opened 2015-08-24
Test fix: delay next command until page has been loaded

dvuckovic:
\+ code tidying


mgruner:
Merged, thanks. Hm, I thought I had also seen this in other circumstances, but not sure.

PR #682, opened 2015-08-25
Removed EntityIDs from process management canvas tool tips to optimise readability

nilei:
The tool tips are great, but since the entity ids are longer, the tooltip content is most often difficult or impossible to read. Thus the information intended to be served by tooltips is none.

This small fix simply removed the entity ids and displays the names only.

-- Cheers, Nils


mrcbnsls:
Merged in 1b63e1e, thanks!

PR #684, opened 2015-08-26
Change label handling and improve search field accessbility

dvuckovic:
@mgruner Associated `LABEL` element is now redirected to the search field, if it exists. Upon deactivation, old behavior is restored.

If label does not exist, additional `ARIA-LABEL` attribute is added to the search field, based on the values of `ARIA-LABEL` or `TITLE` attribute of the respective `SELECT` field. As a fallback, sanitized search field `ID` is used.

Adding `TITLE` attribute to `INPUT` field via jQuery didn't work for me, it was just silently ignored. Not sure why :(

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -33,7 +33,9 @@ Core.UI.InputFields = (function (TargetNS) {
         SelectionBoxOffsetLeft: 5,
         SelectionBoxOffsetRight: 5,
         ErrorClass: 'Error',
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @dvuckovic, thanks a lot for your help with this. I found that redirecting the label for attribute does not help, at least on MacOS's screen reader (VoiceOver). Therefore I simplified the solution in f34b00a. Please let me know if this is acceptable for you.

dvuckovic:
Hi @mgruner, no problem, sorry to hear it didn't work. I think your solution is fine, it's definitely better option not to redirect label if there is no need for it. Let me know if you need anything else.

PR #685, opened 2015-08-27
Code polish for CustomerTicketZoom.pm

igor0511:
Changes:
* Removed one badly formed if
* Additional check for an array


mgruner:
Thanks!

PR #691, opened 2015-09-04
Integrate TicketZoomDynamicFieldDisplay from OPAR: Show complete cont…

reneeb:
…ent of DynamicFields in AgentTicketZoom when the admin configured it...

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -1429,13 +1429,14 @@ sub MaskAgentZoom {
             };
         }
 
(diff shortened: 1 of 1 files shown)


mgruner:
Seems like you're introducing a new undocumented value 2 in Ticket::Frontend::AgentTicketZoom###DynamicField, right? This should at least be documented on the config setting description. But I'm not fully convinced right now about it. Feels like an abuse of the setting and inconsistent to other places (2 = mandatory).

reneeb:
If you want, I can use an other value...

reneeb:
@mgruner Any idea what value should be used?

mgruner:
Not really. But the general configurability is already there, it just does not distinguish by field. I'm closing this PR for now, please feel free to reopen with more information. Thanks a lot @reneeb.

PR #692, opened 2015-09-08
[Master - bug #11470] - AgentPreferences shows "My Services" even when Ticket::Service is disabled

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11470


s7design:
HI Martin,

This is my proposal for fixing this bug. Please let me know what do you think about that.

Regards
Zoran

mgruner:
Hi Zoran,
would it perhaps be better to perform the check in PreferencesCustomService.pm?
Regards, mg

s7design:
Hi Martin,

Thanks for suggestion.  I updated code. I believe that it is better now.

Regards
Zoran

mgruner:
Merged, thanks Zoran!

PR #693, opened 2015-09-09
Input fields tweaks

dvuckovic:


var/httpd/htdocs/js/Core.UI.InputFields.js
```diff
@@ -1777,14 +1777,19 @@ Core.UI.InputFields = (function (TargetNS) {
...
      *      This function check if Input Field is initialized for the supplied element,
-     *      and returns corresponding ID of the Search field.
+     *      and returns appropriate boolean value.
      */
     TargetNS.IsEnabled = function ($Element) {
-        return $Element.data('modernized');
+        if ($Element.data('modernized') && $Element.data('modernized') !== '') {
+            return true;
+        }
+        else {
```
mgruner: Very minor suggestion: the else { } is not needed around the return false (linear coding).

mgruner:
Hello Dusan,
thanks a lot, I merged this.
There is another issue that was also there before this patch: If I fill out all fields in AgentTicketPhone except for the Queue, then it is focused (because value is missing but mandatory) on the first try to submit, but not subsequently any more. Just nothing happens. Could you check that please?
Regards, Martin

PR #694, opened 2015-09-09
Fix BUG#11475

eliasp:
Don't delete mails, just because the IMAP server didn't return any
payload at the time of asking for it.
Delete mails only from the server, once payload was provided and the
mail was successfully processed by PostMaster.

### DIFF
--- a/Kernel/System/MailAccount/IMAP.pm
+++ b/Kernel/System/MailAccount/IMAP.pm
@@ -229,11 +229,11 @@ sub _Fetch {
                                 . "$File, report it on http://bugs.otrs.org/)!",
                         );
                     }
+                    # mark email to delete once it was processed
+                    $IMAPObject->delete($Messageno);
                     undef $PostMasterObject;
-                }
 
-                # mark email to delete if it got processed
-                $IMAPObject->delete($Messageno);
+                }
 
                 # check limit
                 $Self->{Limit}++;
--- a/Kernel/System/MailAccount/IMAPTLS.pm
+++ b/Kernel/System/MailAccount/IMAPTLS.pm
@@ -233,11 +233,10 @@ sub _Fetch {
                                 . "$File, report it on http://bugs.otrs.org/)!",
                         );
                     }
+                    # mark email to delete once it was processed
+                    $IMAPObject->delete_message($Messageno);
                     undef $PostMasterObject;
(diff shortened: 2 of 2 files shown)


eliasp:
See also: http://bugs.otrs.org/show_bug.cgi?id=11475

Copied from the bug description:
While debugging IMAP issues, I've stumbled upon a possible fatal issue in Kernel::System::MailAccount::IMAP->Fetch() where mails could be deleted from the server without having been processed.

`sub _Fetch()` iterates over all messages provided by `$IMAPObject->select()`.
Within this iteration, the mail:
- is fetched:
```
# get message (header and body)
my @Lines = $IMAPObject->get($Messageno);
```
- gets checked for content (no content: error is logged, content: processed through `PostMasterObject->Run()`):
```
if ( !@Lines ) { … } else { … }
```
- gets deleted from the server

The problem here which could lead to deletion of unprocessed mails:
not every IMAP server is consistent in the advertised mail (U)IDs and their actual payload (I'm looking at you, MS Exchange 2013)
When a new mail arrives in a mailbox/folder, the server might already announce its (U)ID (here: retrieved through `$IMAPObject->select()`, but not yet return it's payload (here: retrieved through `$IMAPObject->get($Messageno)`).

This can lead to a situation, where a mail, whose payload isn't synced yet with its (U)ID is deleted by `$IMAPObject->delete($Messageno)`, as this deletion is based on the condition whether payload was returned or not, instead of being bound to a mail being processed successfully through `$PostMasterObject->Run()` or not.

mgruner:
Hello Elias, thanks very much for the bug report and your PR. I would request a change here. The mail should be deleted always if it was fetched ok and fed into the PostMaster, no matter if the ticket could be created or not. Otherwise there could be "hanging mails". So the delete call should not be in a new ```else``` branch, but before the ```undef $PostMasterObject```. Ok? Can you change this please? Please also provide this fix for ```IMAPTLS```. 

Regards, Martin

eliasp:
@mgruner You're right - this makes sense, as failing mails will be handled by PostMaster anyways (moved to the spool directory).

I've updated the PR to reflect this and also applied the same fix to the `IMAPTLS` module.

mgruner:
Merged, thank you so much @eliasp!

PR #698, opened 2015-09-10
 [Master - Bug11152] - Ending of daylight saving time caused system collapse

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11152

Hi @mgruner ,

This is my proposal for fixing. I tested it a lot, and this is one solution which could be acceptable. 
Maybe it should be tested a little bit more (e.g. different time zone...  )
You can see in the comments that there is the issue  only in the time zone which have ending time from 00:00:00 to 23:00:00.
I tried to fix with another solution where I compare new StartTime and  $CTime00 + 24 * 60 * 60. This is something similar to solution in DestinationTime() function.

<pre>
        # Compensate for switching to/from daylight saving time
        # (day is shorter or longer than 24h)
        if ( $NewCTime != $CTime00 + 24 * 60 * 60 ) {
            my $Diff = $NewCTime - $CTime00 - 24 * 60 * 60;
            $DestinationTime += $Diff;
        }
</pre>

Regards
Zoran

### DIFF
--- a/scripts/test/Time/WorkingTime.t
+++ b/scripts/test/Time/WorkingTime.t
@@ -22,6 +22,7 @@ my @Tests = (
         TimeStampUTCStop  => '2015-05-19 12:00:00',
         ServerTZ          => 'UTC',
         Result            => '7776000',               # 90 days
+        ResultTime        => '90 days',
     },
     {
         Name              => 'Europe/Berlin ( Daylight Saving Time UTC+1 => UTC+2 )',
@@ -29,6 +30,7 @@ my @Tests = (
         TimeStampUTCStop  => '2015-05-19 12:00:00',
         ServerTZ          => 'Europe/Berlin',
         Result            => '7779600',                                                 # 90 days and 1h
+        ResultTime        => '90 days and 1h',
     },
     {
         Name              => 'UTC',
@@ -36,6 +38,7 @@ my @Tests = (
         TimeStampUTCStop  => '2015-02-22 04:00:00',
         ServerTZ          => 'UTC',
         Result            => '21600',                                                   # 6h
+        ResultTime        => '6h',
     },
     {
         Name              => 'America/Sao_Paulo - end DST from 00 to 23  ( UTC-2 => UTC-3 )',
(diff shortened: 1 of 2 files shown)


mgruner:
Hello Zoran,
can you come up with a unit test that reproduces the original problem? Perhaps with the help of $Env{TZ} and FixedTimeSet() if needed, as in MailTimeStamp.t? I didn't look at the details yet.
Regards, mg

s7design:
Hi Martin,

Yes of course I can. It will be done tomorrow. 

Regards
Zoran

s7design:
Hi @mgruner 

There is test for WorkingTime function. I added some test cases with time from different time zone, please let me know if you have any suggestion.

Regards
Zoran

scripts/test/Time/WorkingTime.t
```diff
@@ -0,0 +1,234 @@
...
+
+use vars (qw($Self));
+
+my @Tests = (
+    {
+        Name              => 'UTC',
+        TimeStampUTCStart => '2015-02-17 12:00:00',
+        TimeStampUTCStop  => '2015-05-19 12:00:00',
+        ServerTZ          => 'UTC',
+        Result            => '7776000',               # 90 days
```
mgruner: Please write these numbers in a readable form like 60 * 60 * 24 * 90.

mgruner:
Hello Zoran,
thanks for the fix so far. Test looks very nice. I'm not deeply in this matter right now, but here in Germany the DST switch is 2-3 am in spring and 3-2 am in autumn. Therefore I wonder why you use 0-23 always. And does your code perhaps only cover one of the directions of change?
Best regards, mg

scripts/test/Time/WorkingTime.t
s7design: Yes of course I will update this. Thanks
Zoran

s7design:
Hi Martin,

I use 23 to 00 and 00 to 01 test cases only for time zone where there are such DST (e.g. America/Sao_Paulo, America/Asuncion -Paraguay, Asia/Tehran, Australia/Sydney)and in for this test cases I added checking for UTC and in some of them I added Berlin to show that my fix still count good in different zone which don't have DST in this moment.
You can see that I tested  'Europe/Berlin ( Daylight Saving Time UTC+1 => UTC+2 )' in first test case to show DST in one of time zone in Europe ( I know about DST in Berlin, because here is the same time zone as in Berlin.)

I will update result number if you want, but I see now that this is result of WorkingTime function. In my opinion if we check result of function in unit test we should test if result right. If I change this in different format for output it will not result of this function. I can change output of test message but this param Result should be number of second that represent working time.

I added comment for each result only to clarify.  

Please let me know if you have different opinion.

Regards
Zoran

s7design:
Hi Martin,

I updated test a little. I hope now it is better. You can see I added ResultTime in test data that is displayed in output after testing.

I can convert seconds of WorkingTime to days, hours, minutes, seconds  by gmtime() but in this moment I think that it will not make test better. But if you would like it I will be able to update code.

Regards
Zoran

mgruner:
Exellent, merged, thank you!

mgruner:
One more comment about readable code. If you have something like 60*60*24*90 in the code, this will be calculated at compile time, and thus not slow down Perl at runtime. So this is more readable than the result IMO. But I decided to keep it in the test because the calculations get quite complex and then it is not more readable than now.

PR #699, opened 2015-09-10
Fixed bug#4661 - SearchProfiles updated on UserLogin

mbeijen:
This is a bug fix for bug#4661.
ref: http://bugs.otrs.org/show_bug.cgi?id=4661

Previously, when the login name of an agent or customer user changed,
their search profiles were lost.

I introduced a function in Kernel::System::SearchProfile to update
the search profiles for a given login, plus I modified UserUpdate
and wrote an event module for CustomerUsers.
I also wrote unit tests to cover these scenarios.

### DIFF
--- a/Kernel/System/User.pm
+++ b/Kernel/System/User.pm
@@ -22,6 +22,7 @@ our @ObjectDependencies = (
     'Kernel::System::Encode',
     'Kernel::System::Log',
     'Kernel::System::Main',
+    'Kernel::System::SearchProfile',
     'Kernel::System::Time',
     'Kernel::System::Valid',
 );
@@ -489,7 +490,12 @@ sub UserUpdate {
         }
     }
 
-    # check if a user with this login (username) already exits
+    # store old user login for later use
+    my $OldUserLogin = $Self->UserLookup(
+        UserID => $Param{UserID},
+    );
+
+    # check if a user with this login (username) already exists
     if (
         $Self->UserLoginExistsCheck(
             UserLogin => $Param{UserLogin},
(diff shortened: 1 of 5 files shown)


Kernel/System/User.pm
```diff
@@ -552,6 +558,15 @@ sub UserUpdate {
         Value  => $Param{UserMobile} || '',
     );
 
+    # update search profiles if the UserLogin changed
```
mgruner: Shouldn't this happen in the event handler instead?

mgruner:
Hi @mbeijen, thank you very much so far for this. Please check the issue I mentioned.

mbeijen:
Hi Martin, I could not use the event handler because there is none for `Kernel::System::User`! Only for `Kernel::System::CustomerUser` and there I made use of it.

If needed I can of course add an event handler to Kernel::System::User as well, but this was not my goal, I just wanted to have ths bug fixed...

mgruner:
Hello Mike, my bad, sorry. This looks very nice; could you add a test case for CustomerUser too please? Then I'd like to merge it for OTRS 5. Seems this would fix http://bugs.otrs.org/show_bug.cgi?id=4661.

mbeijen:
Hi Martin, I'll add a test for CustomerUser as well.

Indeed this would fix http://bugs.otrs.org/show_bug.cgi?id=4661 - that's what I indicated in the title of this pull request!

mgruner:
Ok, an added UT would be nice, then we can merge this.

mbeijen:
I just added some tests for the CustomerUser event module; I'd think this makes the pull request complete. Please let me know if you'd still need anything.

mgruner:
Excellent, thank you very much Mike! Merged in OTRS 5 and master.

PR #700, opened 2015-09-11
Changed behavior when creating phone ticket out of a chat

igor0511:
Changed behavior when creating a new phone ticket from a chat:
If Customer is in a chat, chat won't be deleted immediately (as it's on all other forms).

Best Regards,
Igor


mgruner:
Hello @mrcbnsls, can you please take a look at this after your return?

mrcbnsls:
Thanks, works fine, merged in 39d1a1d.

PR #702, opened 2015-09-14
Master - Fixed test MailTimeStamp

s7design:
Hi @mgruner 

Working on a bug with daylight time zone, I saw that there is a mistake in test MailTimeStamp. There was used wrong data which is uninitialized value in concatenation. I suppose it was type error.

I added also FixedTimeUnset to restore the regular system time behaviour.

Regards
Zoran


mgruner:
:+1: merged, thanks

PR #703, opened 2015-09-15
[Master bug11441 ] - It is not possible to sort by Owner

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11441

Hi @mgruner 

Here is added sort by Owner in customer interface.

Regards
Zoran


mgruner:
Perfect, thanks!

PR #707, opened 2015-09-15
Master - AdminGenericAgent - Unset ticket attributes

s7design:
Hello,

During bug testing, noticed issue with AdminGenericAgent. When selecting "Update/Add Ticket Attributes" there was no option to deselect value. This raised a problem, since there was no other option then to create new generic agent job if ticket attribute is inputted by mistake.
Added 'PossibleNone' for fields that are selection based. Think this will provide much nicer user experience.

Additionally fields 'Pending date' and 'Time units' didn't have validation to be numbers only, which could result in input that's not suppose to be. Now only numeral values can be inputted as supposed.
 
Please, review these small life improvement changes and let me know if there are any issued or questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericAgent.pm
+++ b/Kernel/Modules/AdminGenericAgent.pm
@@ -437,13 +437,14 @@ sub _MaskUpdate {
         Class       => 'Modernize',
     );
     $JobData{NewOwnerStrg} = $LayoutObject->BuildSelection(
-        Data        => \%ShownUsers,
-        Name        => 'NewOwnerID',
-        Size        => 5,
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Sanjin,
please see http://bugs.otrs.org/show_bug.cgi?id=11471. This will not do what we want. It must also work with the new input fields being disabled, and in this case it will both not remove the other selections and also not work from a database perspective. I believe we have to fix this in the input fields, but you don't have to do that. Dusan or one of the Marcs will handle this one.
Regards, Martin

PR #708, opened 2015-09-16
TicketUpdate web service: error message, docs

mbeijen:
I found some references in the TicketUpdate web service to
TicketCreate which are there because of copy and paste, I guess.
This commit change these to TicketUpdate.

Also I corrected the documentation on the returned data structure.

As a side note: I wanted to add a test to catch the error message about the attachments which I changed, but this was not so easy to trigger. the closest I could get was by passing `Attachment => [[]],` but this returned locally 'Error in SOAPOutputRecursion'  and I did not want to test against an error message brought up by SOAP::Lite. If you'd be interested you can find this here: https://github.com/mbeijen/otrs-1/commit/8b8cccebe664103cc7d0a105090d6053382fe0a7?diff=unified

### DIFF
--- a/Kernel/GenericInterface/Operation/Ticket/TicketUpdate.pm
+++ b/Kernel/GenericInterface/Operation/Ticket/TicketUpdate.pm
@@ -22,7 +22,7 @@ our $ObjectManagerDisabled = 1;
 
 =head1 NAME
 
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @mbeijen, very nice to hear from you! Thanks for your fix, I merged this in master and rel-5_0. :+1:

PR #709, opened 2015-09-17
[Master - bug11330]  LogModule::LogFile::Date creates log files ending in single digit

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11330

Hi Martin,

Here is my proposal for fixing this bug. Please let me know if you have any suggestion about that.

Regards
Zoran


mgruner:
Merged, thanks!

PR #710, opened 2015-09-17
[Master bug 11481] AutoResponse will be send altough it is set to invalid

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11481

Hello @mgruner ,

Hereby is proposal for bug fix. Now it should just use auto responses that are valid. There is also unit test for auto response sent and testing behavior if they are valid-invalid.

If you want, i can make PRs for rel4-0 and rel5-0.

If there are any suggestion or modification, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/AutoResponse.pm
+++ b/Kernel/System/AutoResponse.pm
@@ -297,29 +297,17 @@ sub AutoResponseGetByTypeQueueID {
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
 
-    # build valid id string
-    my $ValidIDsString;
-    if ( $Param{Valid} ) {
-        $ValidIDsString = join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet();
-    }
-    else {
-        my %ValidList = $Kernel::OM->Get('Kernel::System::Valid')->ValidList();
-        my @ValidIDs  = sort keys %ValidList;
-        $ValidIDsString = join ', ', @ValidIDs;
-    }
-
     # SQL query
     return if !$DBObject->Prepare(
-        SQL => '
-            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id, ar.id, ar.valid_id
+        SQL => "
+            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id
             FROM auto_response_type art, auto_response ar, queue_auto_response qar
-            WHERE ar.valid_id IN (?)
+            WHERE ar.valid_id IN ( ${\(join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet())} )
                 AND qar.queue_id = ?
                 AND art.id = ar.type_id
                 AND qar.auto_response_id = ar.id
-                AND art.name = ?',
+                AND art.name = ?",
         Bind => [
-            \$ValidIDsString,
             \$Param{QueueID},
             \$Param{Type},
         ],
--- a/Kernel/System/AutoResponse.pm
+++ b/Kernel/System/AutoResponse.pm
@@ -15,6 +15,7 @@ our @ObjectDependencies = (
     'Kernel::System::DB',
     'Kernel::System::Log',
     'Kernel::System::SystemAddress',
+    'Kernel::System::Valid',
 );
 
 =head1 NAME
@@ -253,6 +254,7 @@ get a hash with data from Auto Response and it's corresponding System Address
     my %QueueAddressData = $AutoResponseObject->AutoResponseGetByTypeQueueID(
         QueueID => 3,
         Type    => 'auto reply/new ticket',
+        Valid   => 1    # optional
     );
 
 Return example:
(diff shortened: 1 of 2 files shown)


mgruner:
Hello Sanjin,
thanks for your proposal. I have some issues currently:

```
[Fri Sep 18 09:44:34 2015] otrs.Console.pl: DBD::Pg::st execute failed: ERROR:  invalid input syntax for integer: "1, 2, 3" at /ws/otrs6/Kernel/System/DB.pm line 666.
ERROR: OTRS-otrs.UnitTest-13 Perl: 5.16.2 OS: darwin Time: Fri Sep 18 09:44:34 2015

 Message: ERROR:  invalid input syntax for integer: "1, 2, 3", SQL: '
            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id, ar.id, ar.valid_id
            FROM auto_response_type art, auto_response ar, queue_auto_response qar
            WHERE ar.valid_id IN (?)
                AND qar.queue_id = ?
                AND art.id = ar.type_id
                AND qar.auto_response_id = ar.id
                AND art.name = ? LIMIT 1'
```
I recommend to also test on a PostgreSQL DB as it is stricter. Other comments inline.
Regards, mg

Kernel/System/AutoResponse.pm
```diff
@@ -295,16 +297,29 @@ sub AutoResponseGetByTypeQueueID {
...
+    }
+
     # SQL query
     return if !$DBObject->Prepare(
         SQL => '
-            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id
+            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id, ar.id, ar.valid_id
             FROM auto_response_type art, auto_response ar, queue_auto_response qar
-            WHERE qar.queue_id = ?
+            WHERE ar.valid_id IN (?)
```
mgruner: Binds do not work with lists until now. You might have to insert the SQL directly as in other files.

Kernel/System/AutoResponse.pm
```diff
@@ -295,16 +297,29 @@ sub AutoResponseGetByTypeQueueID {
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
 
+    # build valid id string
+    my $ValidIDsString;
+    if ( $Param{Valid} ) {
```
mgruner: Ok, I understand that by making it this way the API is not really changed, just extended. But I don't understand how it should work that without a change in the calling code now only the valid auto responses are supposed to be fetched and sent.

Kernel/System/AutoResponse.pm
```diff
@@ -295,16 +297,29 @@ sub AutoResponseGetByTypeQueueID {
...
+        my %ValidList = $Kernel::OM->Get('Kernel::System::Valid')->ValidList();
+        my @ValidIDs  = sort keys %ValidList;
+        $ValidIDsString = join ', ', @ValidIDs;
+    }
+
     # SQL query
     return if !$DBObject->Prepare(
         SQL => '
-            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id
+            SELECT ar.text0, ar.text1, ar.content_type, ar.system_address_id, ar.id, ar.valid_id
```
mgruner: You add two new fields to the SELECT list but don't fetch them in the result. They are probably also not needed.

mgruner:
Please ask Zoran to have a look on your changes after addressing the issue. Thanks!

Kernel/System/AutoResponse.pm
s7design: Yes it is our mistake, we were trying something with  ar.id but we forgotten to delete in finale version.
It will be deleted

Thanks 
Zoran

Kernel/System/AutoResponse.pm
s7design: Thanks, I hope it could solve error that there is in PostgreSQL maybe 

Zoran

Kernel/System/AutoResponse.pm
s7design: Martin, I have a question about your comment. 

I suggested to add valid param here but in this moment I see that it is not necessary. Did you wanted to suggest such solution? We will select only valid auto response here without adding Valid param. Am I right?
Please let me know if you have another idea about it.

Thanks
Zoran

mgruner: I don't think this will work as you want to be able to edit also invalid auto responses in the admin interface, right? So a new parameter might be needed.

s7design:
Hi Martin,

One another question, I would like to clarify what is your suggestion in this moment.
Do you want that we do something that the reporter suggested in the bug:
* workaround: remove assignment of AutoResponse to queue before, setting AutoResponse to invalid
In this case it would be solved this issue but also you suggest that we use Valid parameter which we add AutoResponseGetByTypeQueueID.
Sorry if I ask many question, but it is better to clarify now and do not do something wrong.

Regards
Zoran

s7design:
We can do something that it is used in Queue - Template assignment ??
Is it good example that we can follow.
Zoran

s7design:
Hello @mgruner ,

First want to thank you for your feedback and advices. AutoResponse.pm is now updated and cleaned of unnecessary code. Did testing in PostgreSQL, both manual and Unittest showed expected results. 

This way of handling now invalid auto-responses is somewhat similar to that of queue-templates. If you would prefer option that reporter suggested to delete queue-autoresponse relation upon switching to invalid, i can update this PR. Think current approach is better since it gives user option to switch back on valid and still keep there other configurations.

Please let me know if there any further issues regarding this PR.

Regards,
Sanjin

mgruner:
Excellent, merged, thank you Sanjin!

PR #711, opened 2015-09-18
[ Master - 11346 ] - Naming different when adding columns to CIC vs Dashboard

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11346

Hi Martin,

Here I my fix for this issue. There was a problem in place where the template  is included.
I tested it for all screens where there is ticket overview. I think it is solved now, and there is the same behavior everywhere. 

Regards
Zoran


mrcbnsls:
Merged in a3a0538df91d003599880b2bf4e11e0b68eeeeda, thanks!

PR #712, opened 2015-09-18
Agent selection on top of text editor

nilei:
Moved agent selection for text recipients in note and other screen to on top of the text editor. This is more userfriendly and known to users from email clients similar programs. Having the agent selection below the text editor feels misplaced and it often happens that users forget to select additional recipients.


mrcbnsls:
Nice idea! Merged in 56e636d, thanks a lot (will also be in OTRS 5 rc1, 5e76c28).

PR #713, opened 2015-09-21
[ Master bug 11196 ] - "Created by" detail missing in PDF print 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11196

Hi @mgruner 

This is our proposal for solution. I added  "Created by" on the same way as it is in Ticket Zoom screen.

Regards
Zoran


mgruner:
Merged, thanks!

PR #715, opened 2015-09-22
 [ Master - bug 11317 ] - When service is mandatory in customerfrontend - error occurred!

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11317

Hello,

Hereby is bug fix proposed by reporter. Tested and it work as expected now.

Regards.
Sanjin


mgruner:
Merged, thanks a lot!

PR #718, opened 2015-09-23
Provide SolutionInMin also without Solution Dest.

capeit-tto:
The ticket has a solution time and thus also a solution time in min, even if there's no solution-destination-time. There's no reason to make the calculation of SolutionInMin depending from Escalation(SolutionTime).


mgruner:
Hello @capeit-tto, thanks a lot. We just discussed this yesterday too, funny. ;) We'll consider this change for OTRS 5 and master, but not for OTRS 4.

PR #720, opened 2015-09-23
Provide SolutionInMin also without Solution Dest.

capeit-tto:
The ticket has a solution time and thus also a solution time in min, even if there's no solution-destination-time. There's no reason to make the calculation of SolutionInMin depending from Escalation(SolutionTime).

see also PR#718, PR#719


dominikklein:
Hello Torsten,

thanks a lot!

Best regards,
Dominik

PR #721, opened 2015-09-23
[ Master - Bug 11315 ] - subservice not apparent in overview when many services are in the list 

s7design:
Hi @mgruner 

Here is PR with our proposal for fixing. We used something similar to Ticket Queue View. As you can see an indentation is not used for RTL. If it is needed, we will be able to provide new PR and solve this for service and queue view as well.
I suppose you know that there is possible only 5 level of services in TicketServiceView. I didn't want to fix it here but again if you consider that it should be fixed, we will be ready to solve it.

Regards
Zoran


mrcbnsls:
Merged in 05f4e99, thx. I also added some .RTL css (for queue view, too). It's not necessary to fix the max levels issue now, because the database does only accept 200 chars anyway. We can do that later.

PR #722, opened 2015-09-23
[ Master - Bug 11496 ] - Search field shown even if type and service disabled

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11496

Hello @mgruner ,

Hereby is our proposal for fixing this bug. Set condition in CustomerTicketSearch.tt to check which part should be shown depending on config. There was other option to solve this with blocks, if you don't prefer this way i will revisit this bug and solve it differently

Additionally while working on it, found out that there was left $Self->{LayoutObject} in CustomerTicketSearch.pm. So corrected that as well

If you have any questions regarding this PR, please let me know.

Regards,
Sanjin


mrcbnsls:
Merged in 6df9702, thx.

PR #723, opened 2015-09-24
[ Master Bug 11457 ] - Agent language preference select field is not translated

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11457

Hi @mgruner 

As we had spoken I provided this PR with proposal for fixing the bug and making some improvement for preferences languages. I believe it is much better now. There are not native names for some languages.
I left it empty and the same for English languages. We could omit the English languages in DefaultUsedLanguagesNative config item.  However, I hope it is better that we avoid the wrong configuration.

Maybe we can change descriptions of these config items, if you have any suggestion about that please let me know.

What do you think about all of that?

Regards
Zoran


mgruner:
HI @s7design, looks very nice! Right now I only have one request: please convert the entities to utf8 characters.

s7design:
Hi @mgruner 

I converted the entities to utf8. I hope it is good now. If I can change anything else let me know.

Regards
Zoran

mgruner:
Hello Zoran,
this was not precisely what I meant, I want real UTF8 in the file. I changed this and merged with other changes in https://github.com/OTRS/otrs/commit/1cd7d76eb796dade838ff7f0c5cb8dc3724afe44. I added the missing Translatable flag in the English name strings and switched the order of display. Please let me know if you agree.
One more thing: please provide future PRs for the lowest branch where they shall be applied, in this case it would be rel-5_0. Directly to master only if it is an OTRS 6 feature.
Regards, mg

s7design:
Thanks Martin. 

I misunderstood, sorry. I meant that you wanted encoded UTF8 characters as it had been with some languages ( e.g Russian, Chinese ... ) Now, I know what you wanted. I believe it is better, because of course it is easier for administrators to update in real character instead of encoded characters. 

We will provide PR in appropriate branch in further work.

Regards
Zoran

PR #724, opened 2015-09-25
[ Rel-5_0 - Bug 8220 ] - OutOfOffice does not check if end date is before start date

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8220

Hello,

Hereby is fix proposal for this bug. Added error message and prevented preference update if user try to set out of office start time after end time. This check is for AdminUser and AgentPreference. Modified selenium test to check for this fix.

If there are any question regarding this PR, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Preferences/OutOfOffice.pm
+++ b/Kernel/Output/HTML/Preferences/OutOfOffice.pm
@@ -11,6 +11,8 @@ package Kernel::Output::HTML::Preferences::OutOfOffice;
 use strict;
 use warnings;
 
+use Kernel::Language qw(Translatable);
+
 our @ObjectDependencies = (
     'Kernel::Output::HTML::Layout',
     'Kernel::System::Web::Request',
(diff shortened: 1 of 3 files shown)


Kernel/Output/HTML/Preferences/OutOfOffice.pm
```diff
@@ -90,48 +91,68 @@ sub Run {
...
 
-        $SessionObject->UpdateSessionID(
-            SessionID => $Self->{SessionID},
-            Key       => 'UserLastname',
-            Value     => $User{UserLastname},
-        );
+        $Self->{Message} = 'Preferences updated successfully!';
+    }
+    else {
+        $Self->{Error} = "Start date shouldn't be defined after End date!";
```
mgruner: These strings must be translated. If that happens later already, they should at least be marked with ```Translatable()``` here. Besides, the English is not fully correct (end lowercase); you could use "Please specify an end date that is after the start date.". This also avoids the negation.

s7design:
Hi @mgruner ,

Thank you for your suggestion, PR is updated accordingly.

Regards,
Sanjin

mgruner:
Excellent, thank you, also for the test!

PR #725, opened 2015-09-25
 [ Master - Bug 8220 ] - OutOfOffice does not check if end date is before start date

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8220

Hello @mgruner ,

Hereby is fix for this bug as well for master branch since i had it already prepared.

Regards,
Sanjin


mgruner:
Merged.

PR #726, opened 2015-09-25
[ Master - Bug 11234 ] - Attachments lost due to filename collision in ArticleStorageFS

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11234

Hello @mgruner ,

Hereby is our bug fix proposal. As you suggested there was part that is responsible for detecting filename collision, but it was not working in console command since Force was set to 1. 
Added unit test for console command StorageSwitch. Used fixed time to create safe environment so it doesn't get in conflict with other possible tickets during testing.

If there are questions or issues regarding this PR, please let me know.

Regards,
Sanjin


dvuckovic:
Hi S7,

thanks a lot for the fix & test. It will be included in 5.0.8.

Regards,
Dusan

PR #729, opened 2015-09-28
Fix mouse scrolling issues with new positioning scheme (Input Fields)

dvuckovic:
The reason why issues were popping up only in FF is because `Event.srcElement` property was Chrome-only and not defined in FF. I substituted it for `Event.target`, and also introduced additional check if drop down list is scrolled.


mrcbnsls:
Very nice, big thx, works perfectly fine. Merged in 2912271.

PR #730, opened 2015-09-28
Fix: always check SysConfig before modernizing fields (bug#11506)

dvuckovic:


mrcbnsls:
Merged in c6ae2e7, thanks a lot.

PR #732, opened 2015-09-29
Master - cleaned up code

s7design:
Hi @mgruner 

Working on bug-fixing we saw that there is this log entries. I suppose it is unneeded code that was used in developing and debugging.

Regards
Zoran


mgruner:
Indeed! Thanks, merged.

PR #733, opened 2015-09-29
Master - port to OM some objects

s7design:
Hi @mgruner 

There are some fixes for porting to OM Log and Cache objects in some modules.

Regards


mgruner:
Merged, thank you!

PR #735, opened 2015-09-29
[ Rel-5 Bug 11510 ] - AgentTicketMove empty by sysconfig changes

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11510

Hi @mgruner,
 
There is proposal for fixing. I think that the reporter is right. There is always necessary WidgetTicketActions because Queue tields are always in this screen.

Regards
Zoran


s7design:
Hi @mgruner 

 There is fixed spelling mistakes in AgentTicketMove.pm as well.

Regards
Zoran

s7design:
There was some problem with rebasing, I close this one and I will make another one

PR #737, opened 2015-09-29
Fix bug#11518 - Error in empty user id and password in login screen

noritnk:
This PR is fix for bug#11518.  See http://bugs.otrs.org/show_bug.cgi?id=11518 .

### DIFF
--- a/Kernel/System/CustomerAuth/DB.pm
+++ b/Kernel/System/CustomerAuth/DB.pm
@@ -93,15 +93,6 @@ sub GetOption {
 sub Auth {
     my ( $Self, %Param ) = @_;
 
-    # check needed stuff
-    if ( !$Param{User} ) {
-        $Kernel::OM->Get('Kernel::System::Log')->Log(
-            Priority => 'error',
-            Message  => "Need User!"
-        );
-        return;
-    }
-
     # get params
     my $User       = $Param{User}      || '';
     my $Pw         = $Param{Pw}        || '';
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @noritnk,

Thank you for the pull request, but it can't be merged cleanly, please update you branch and send the pull request again.

((enjoy))
cr

noritnk:
Hi @carlosfrodriguez,
Sorry, this PR was conflicting with fix for bug#11517.
I updated this PR, and rebased.
Thanks,

carlosfrodriguez:
Hi @noritnk,

I review this pull request, the problem is that is makes it inconsistent with other methods (except for BasicAuth) all other methods included 2FactorAuthentication, logs the lack of User param.

((enjoy))
cr

mgruner:
I also wonder if this really solves a problem. Filling a web server log is always possible, for example with 404 attacks, or calling up OTRS pages without permission, etc. The attacker could also provide a wrong user name and/or a wrong password. This cannot be avoided, I guess.

noritnk:
Hi carlos and mgrunner,
Thank for reviewing.  This bug may not have to fix.
However, Fixed in customer interface in Bug#11517,
so I think it's good that it's also so in agent interface.

noritnk:
However my thought isn't strong.  So if you desagree, please close this PR without fix.
Thans,

noritnk:
Hi carlos and mgrunner,
Thank for reviewing.
I that that an error is written into web server log by normal operation is wrong.
Empty password is occurrable by missing operation even if not 404 attack.
Thanks,

carlosfrodriguez:
Hello @noritnk,

We have to reject this pull request this time, as the log is intended in this cases. Please feel tree to keep sending us more pull requests, we really appreciate that.

((enjoy))
cr

PR #739, opened 2015-09-30
[ Rel-5_0 Bug 11513 ] - Out of Office is missing in owner dropdown

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11513

Hi @mgruner,

Here is PR with a fix for this issue. Actually, it worked fine, but deleting cache date in SetPreferences was wrong.  @CacheKeys  is not configured well and it had made a problem. We conclude it when we deleted cache after update OutOfOfice in AgentPreferences. In this case all worked fine. Updating   OutOfOfice in AdminUser has worked fine before because in UpdateAgent delete entire cache.  

Regards
Zoran

### DIFF
--- a/Kernel/System/User.pm
+++ b/Kernel/System/User.pm
@@ -560,11 +560,11 @@ sub UserUpdate {
 
     # update search profiles if the UserLogin changed
     if ( lc $OldUserLogin ne lc $Param{UserLogin} ) {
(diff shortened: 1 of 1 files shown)


carlosgarciac:
Hi @s7design,

Your PR is now merge on 86d65fa.

Thanks a lot for your effort.

Best regards,
Carlos García

PR #740, opened 2015-09-30
 [Rel-5_0 - Bug 11511] - Text length for title is cut off after 50 characters

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11511

Hello @mgruner ,

Hereby is proposal to fix this bug. It was hard-coded on 50 chars for ticket subject in complex view linked tickets table and in LinkDelete screen. Changed it to use SysConfig Ticket::SubjectSize so users can change it to there needs. 

Updated selenium test to check for this change as well.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/LinkObject/Ticket.pm
+++ b/Kernel/Output/HTML/LinkObject/Ticket.pm
@@ -20,6 +20,7 @@ our @ObjectDependencies = (
     'Kernel::System::State',
     'Kernel::System::Type',
     'Kernel::System::Web::Request',
+    'Kernel::System::Ticket',
 );
 
 =head1 NAME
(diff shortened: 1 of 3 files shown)


Kernel/Output/HTML/LinkObject/Ticket.pm
```diff
@@ -197,9 +203,8 @@ sub TableCreateComplex {
                 CssClass => $CssClass,
             },
             {
-                Type      => 'Text',
-                Content   => $Ticket->{Title},
-                MaxLength => 50,
```
mrcbnsls: Why not just do something like this here?
```
MaxLength => $Kernel::OM->Get('Kernel::Config')->Get('Ticket::SubjectSize') || 50,
```

s7design: Thanks @mrcbnsls I thought that it is the as well as it is in AgentTicketPhone.pm, AgentTicketZoom.pm or similar screens where it is stripped/cleaned up a ticket subject. I did not consider it that you mentioned about 'title'. I will update this PR. 
Regards
Zoran

s7design:
Hi @mrcbnsls 

I updated PR as you suggested. There is some changes in the test as well. Please let me know if I can improve it.

Regards
Zoran

mrcbnsls:
Thanks, merged in 40b53ef.

PR #743, opened 2015-10-01
[Rel-5_0 - Bug 11467] - ACL does not work with Dest Link in Customer::ModuleRegistration

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11467

Hello @mgruner ,

Hereby is proposal for fixing this bug. Apart from reported issue, added so ACL now control fields when default queue is used. Matched ACL will control now other fields as well on load. Corrected few typos.

Please let me know, if there are any questions or issues regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/CustomerTicketMessage.pm
+++ b/Kernel/Modules/CustomerTicketMessage.pm
@@ -54,7 +54,7 @@ sub Run {
     my $LayoutObject  = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
     my $BackendObject = $Kernel::OM->Get('Kernel::System::DynamicField::Backend');
 
-    # reduce the dynamic fields to only the ones that are desinged for customer interface
+    # reduce the dynamic fields to only the ones that are designed for customer interface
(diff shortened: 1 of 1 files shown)


Kernel/Modules/CustomerTicketMessage.pm
```diff
@@ -226,7 +229,7 @@ sub Run {
         $Output    .= $LayoutObject->CustomerNavigationBar();
         $Output    .= $Self->_MaskNew(
             %GetParam,
-            QueueID          => $QueueDefaultID,
+            %ACLCompatGetParam,
```
carlosfrodriguez: Hello @s7design,

What is the advantage here to use the %ACLCompatGetParam instead?

((enjoy))
cr

carlosfrodriguez:
Hi @s7design,

Thank you for your pull request, it was merged in https://github.com/OTRS/otrs/commit/667143853735b15a474b1847167b01b70024f81c.

((enjoy))
cr

PR #744, opened 2015-10-01
Remove superfluous char...

reneeb:
Seems to be an artifact of OTRS 3


mgruner:
Merged in 5 and master, thank you very much @reneeb!

PR #746, opened 2015-10-02
Fix: allow reseting fields with SIZE > 1 (bug#11471)

dvuckovic:


mgruner:
Excellent, thank you @dvuckovic!

PR #747, opened 2015-10-02
[Rel-5_0 - Bug 11393] -  ACL not hiding Process NavBar link when Type is Menu

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11393

Hello @mgruner ,

Hereby is proposal to fix this bug. When you set sysconfig so CustomerTicketProcess is shown in Menu instead of TicketMenu, there was no part to hide it when process is not available to respective customer. 

Added part of code that will remove button in a same fashion as it is done when it's in TicketMenu.

Please let me know if there any questions or issues regarding this proposal.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/NavBar/CustomerTicketProcess.pm
+++ b/Kernel/Output/HTML/NavBar/CustomerTicketProcess.pm
@@ -144,6 +144,9 @@ sub Run {
     # remove CustomerTicketProcess from the TicketMenu
     delete $Return{$NavBarName}->{$Priority};
 
+    # remove CustomerTicketProcess from the Menu if set outside of the TicketMenu, see bug #11393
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @s7design,

I'm not 100% sure that we want this, since moving the Sub-menu to it own group in the way it is done sounds more like a hack than a feature supported by the system.

@mgruner What do you think?

((enjoy))
cr

nilei:
Hi @carlosfrodriguez, moving navigation items to "another place" is 100% a feature of OTRS!

carlosfrodriguez:
Hi @s7design, 

This pull request was merged in: https://github.com/OTRS/otrs/commit/d0e847d14719aefea5ea2eb75c78075a33d04f85.

Thanks for your help.

((enjoy))
cr

PR #750, opened 2015-10-05
Fix: validation highlights wrong field (bug#11527)

dvuckovic:


mgruner:
Oh :frowning:

mgruner:
Thanks a lot for finding and fixing this, @dvuckovic!

PR #751, opened 2015-10-06
[ Rel-5_0 Bug 11535 ] - Listed message for disabled languages in ticket notification management 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11535

Hi @mgruner, 

I believe that reporter of the bug is right. It is not needed that there is notification if the language is disabled.
I have some suggestions for improvement in this screen with languages, but it will be in another PR, where I will explain it.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminNotificationEvent.pm
+++ b/Kernel/Modules/AdminNotificationEvent.pm
@@ -1033,7 +1033,9 @@ sub _Edit {
 
     my $HTMLUtilsObject = $Kernel::OM->Get('Kernel::System::HTMLUtils');
 
+    LANGUAGEID:
     for my $LanguageID (@LanguageIDs) {
+        next LANGUAGEID if !defined $DefaultUsedLanguages{$LanguageID};
 
         # format the content according to the content type
         if ( $Param{RichText} ) {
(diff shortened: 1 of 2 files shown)


carlosgarciac:
Hi S7,

If you avoid showing the content for a notification language and you save the data on that screen, the info for that specific notification will be lost. I think that's not a nice secondary effect. 

Perhaps instead of the language long name, we could add something like " 'de' language is not enabled on the system, you can delete it if is not needed anymore. " ( my English is not good enough, the message have a lot of room for improvement ), then you fix the style issue on the header of the widget and put into the administrator's hands the decision of deleting that language or not.

s7design:
Hi @carlosgarciac 

I updated this PR with solution that you suggested. You are right. It is better that admin have opportunity to check if the notification should be deleted.

This is my proposal for now, if you have any new suggestion please let me know. I add text 
* "language does not exist. This notification could be deleted, if it is not needed anymore" 
but we can change it if you want.

Regards
Zoran

carlosgarciac:
Hi @s7design,

Thanks a lot for your nice PR, it was merged on 7a49521 with some changes.

Best regards,
Carlos García

PR #752, opened 2015-10-06
[ Master - Bug 11260 ] - missing translation escalation

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11260

Hi @mgruner 

As you can see on the bug report, where I have commented, that there was a problem with cash data for the Ticket Escalation result. I have changed cash key and added user language. I believe it is good solution. Please let me know what do you think about that. 

P.S. I made PR in master, I tested something and stayed on this branch when I had fixed. I now saw that it is in maser but if you want I will be able to make another one in rel-5_0.

Regards
Zoran


carlosgarciac:
HI @s7design,

Thanks a lot for your nice pull request, it was merge on af01c04.

Best regards,
Carlos García

PR #754, opened 2015-10-06
Change HTML to Config

crythias:
This fixes a bug where Data.Name is not populated in DynamicField.tt
See https://github.com/OTRS/otrs/blob/master/Kernel/Modules/AgentTicketProcess.pm#L2361 for corresponding entry in AgentTicketProcess.pm


carlosfrodriguez:
Hi @crythias,

Thank you for you pull request, it was merged in  https://github.com/OTRS/otrs/commit/d6dc4ae51e9359e3afe8c764f72c9df20945b622.

((enjoy))
cr

PR #756, opened 2015-10-07
Escape IDs before using them as jQuery selectors

dvuckovic:
In few places we were using HTML IDs directly as jQuery selectors (by prepending `#` in front). This was failing in SysConfig for example, since some IDs had special characters in them (e.g. `:`) and this raises error in jQuery. A function is introduced to escape these strings before use: `escapeId(id)`.

Per https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -194,6 +194,21 @@ Core.UI.InputFields = (function (TargetNS) {
     };
 
     /**
+     * @private
+     * @name escapeId
+     * @memberof Core.UI.InputFields
+     * @param {String} id - String to escape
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @dvuckovic! Thanks a lot. Could you reconsider using Core.App.EscapeSelector() instead?
@carlosfrodriguez: then this will be the fix that you requested.

mgruner:
@mrcbnsls can you please review and apply this before the RC1 build?

mrcbnsls:
Thanks! Applied the PR on master (8d038f8) and rel-5_0 (720a86d). I only replaced the new custom function by Core.App.EscapeSelector as it does the same thing.

PR #757, opened 2015-10-08
Improvement of language selection in NotificationEvent

noritnk:
This PR Improves language selection in NotificationEvent.

Currently, e.g. if UserLanguage is 'ja' and DefaultLanguage in Config is 'cn' and only English is configured in notification event message.

English is required in AdminNotification.pm, but in this case, English won't be selected.

This PR changes into selected English if notification event message for neither UserLanguage nor DefaultLanguage is not found.

It meens that we can always find notification event message.

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -859,10 +859,15 @@ sub NotificationEvent {
         # otherwise use default language
         $Language = $DefaultLanguage;
 
-        # if no message exists in default language, then take the first available language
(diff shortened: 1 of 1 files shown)


carlosgarciac:
Hi @noritnk,

due your pull request I realize about the issue on selecting proper language for notification, I did a different implementation base on the idea from your PR, thanks a lot for it.

Please take a look on http://bugs.otrs.org/show_bug.cgi?id=11632.

Closing this PR.


Best regards,
Carlos García

PR #758, opened 2015-10-08
[ Rel-4_0 Bug 11304 ] - <OTRS_CUSTOMER_Subject> in "move"-notification is replaced with ticket-subject, not with article-subject 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11304

Hi @martin,

We worked on this bug. To be honest there is not enough clear what are <  OTRS_CUSTOMER_SUBJECT> and <  OTRS_CUSTOMER_BODY  >. In master there are ( OTRS_CUSTOMER_SUBJECT > and <  OTRS_CUSTOMER_BODY >) and
 ( <  OTRS_AGENT_SUBJECT > and < OTRS_AGENT_BODY > ) 
I suppose that something that is called OTRS_CUSTOMER_SUBJECT should be customer subject and in case agent notification instead of this tag there is replaced subject of last customer article. For follow up notification there are subject and body of follow up article, however for Note, MoveUpdate, OwnerUpdate, ServiceUpdate notifications I expect  on the place of <OTRS_CUSTOMER_SUBJECT> tag subject of last customer article.

Am I right. I am sure that we have different behavior for different type of notification. In this PR I apply example how it is solved for ServiceUpdate notification. I believe we should check all type of notification and make the same behavior for all. As I mentioned above in OTRS5 and master there is AGENTs tag and there is necessary check all type of notification as well. I provide another PR for master soon.

Please let me know what do you think about that. This is only proposal, I will be able to provide final solution after suggesting with you.

Regards
Zoran

### DIFF
--- a/Kernel/System/Ticket.pm
+++ b/Kernel/System/Ticket.pm
@@ -4554,11 +4554,9 @@ sub TicketOwnerSet {
             $Self->SendAgentNotification(
                 Type                  => 'OwnerUpdate',
                 RecipientID           => $Param{NewUserID},
-                CustomerMessageParams => {
-                    %Param,
-                },
-                TicketID => $Param{TicketID},
-                UserID   => $Param{UserID},
+                CustomerMessageParams => \%Param,
+                TicketID              => $Param{TicketID},
+                UserID                => $Param{UserID},
             );
         }
     }
@@ -4772,7 +4770,7 @@ sub TicketResponsibleSet {
             $Self->SendAgentNotification(
                 Type                  => 'ResponsibleUpdate',
                 RecipientID           => $Param{NewUserID},
-                CustomerMessageParams => {},
+                CustomerMessageParams => \%Param,
                 TicketID              => $Param{TicketID},
                 UserID                => $Param{UserID},
             );
--- a/Kernel/System/Ticket.pm
+++ b/Kernel/System/Ticket.pm
@@ -1945,7 +1945,6 @@ sub TicketQueueSet {
                     RecipientID           => $UserID,
                     CustomerMessageParams => {
                         Queue => $Queue,
-                        Body  => $Param{Comment} || '',
                     },
                     TicketID => $Param{TicketID},
                     UserID   => $Param{UserID},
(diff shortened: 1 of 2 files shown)


mgruner:
Carlos, can you review this thoroughly? It looks dangerous.

carlosgarciac:
Hi @s7design,

I could review this issue, AFI test the fix works fine for me, just some comments:

1.- On Ticket.pm ln 4777 seem that an empty CustomerMessageParams is not really needed, some lines before $Param{Comment} is set as an empty string, perhaps other parameters are needed, please check this other removed params are not needed here and for Article.pm.
2.- Just to clarify, OTRS_CUSTOMER_x means from last customer article, OTRS_AGENT_x means las agent article.

I hope this helps.

Best regards,
Carlos García

s7design:
Hi @carlosgarciac 

Thanks for the follow-up.  I have updated the PR. 

As you know, it is solved much better in OTRS5. With this fix it is solved the issue, which reporter of the bug mentioned. IMO, OTRS_CUSTOMER_x means from last customer article and OTRS_AGENT_x from last agent article.

CustomerMessageParams is needed param in Article::SendAgentNotification(), and because of that it have to add even it is empty. It is forwarded to TemplateGenerator::NotificationAgent() where if CustomerMessageParams is not set, it will be set to Article data

<pre>
    for (qw(From To Cc Subject Body)) {
        if ( !$Param{CustomerMessageParams}->{$_} ) {
            $Param{CustomerMessageParams}->{$_} = $Article{$_} || '';
        }
        chomp $Param{CustomerMessageParams}->{$_};
    }
</pre>

CustomerMessageParams is used for some date from Customer.

After testing, I saw that it works fine now. I know that Notification in OTRS could be done better, but it is solved in OTRS5. This is only patch for the bug. If you have any other suggestion please lat me know and I will be able to update again.

Regards
Zoran

carlosgarciac:
Hi @s7design,

Thanks a lot for you nice PR, it was merged on 86141d8.

Best regards,
Carlos García

PR #759, opened 2015-10-08
[ Master Bug 11304 ] - <OTRS_CUSTOMER_Subject> in "move"-notification is replaced with ticket-subject, not with article-subject 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11304

Hi @mgruner 

I mentioned this issue in PR for the branch rel-4 ( https://github.com/OTRS/otrs/pull/758 )

Here is proposal for fixing in master. This is not final version as I said, but in this moment you opinion will be helpful. 

Regards
Zoran

### DIFF
--- a/Kernel/System/Ticket.pm
+++ b/Kernel/System/Ticket.pm
@@ -1928,7 +1928,6 @@ sub TicketQueueSet {
                 TicketID              => $Param{TicketID},
                 CustomerMessageParams => {
                     Queue => $Queue,
-                    Body  => $Param{Comment} || '',
                 },
                 Recipients => \@UserIDs,
             },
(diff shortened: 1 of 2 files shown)


carlosgarciac:
Hi S7,

I just included the patch for rel-4_0, but for me seems that removing CustomerMessageParams for 'NotificationFollowUp' is needed as well, could you please help me reviewing this.

Best regards,
Carlos Garcia

s7design:
Hi Carlos,

I think that there is necessary this param. If you make follow-up from customer interface, this param is needed for customer data. CustomerMessageParams is actually used here with real purpose, without it OTRS_CUSTOMER_SUBJECT doesn't work properly, it replace with ticket subject, but there is needed the last customer subject.

Regards
Zoran

carlosgarciac:
Hi @s7design,

Thanks a lot for you nice PR, it was merged on 564a4db.

Best regards,
Carlos García

PR #761, opened 2015-10-08
Add new feature: define default headers for emails

reneeb:
Sometimes someone might want to send some headers in *each* email sent via OTRS. Currently there is no way to define such headers. With this patch it would be possible...

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -6841,7 +6841,7 @@ via the Preferences button after logging in.
     <ConfigItem Name="Email::DefaultHeaders" Required="0" Valid="0">
         <Description Translatable="1">Default headers for emails.</Description>
         <Group>Framework</Group>
-        <SubGroup>Core::Sendmai<l/SubGroup>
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @reneeb, thanks for your proposal. Can you please explain the use case a bit (this would be generally good for new features btw)? I am unsure if your example is really good for agent -> customer communication via email.
Regards, Martin

reneeb:
Hi @mgruner, many mail servers do not send "out of office" replys for mails where those headers in the sample config are set. Some of my customers do not want any "out of office" replys at all, so they want to set the email headers.

Another use case is when OTRS "talks" to another system and you want to send a static X-Header to identify the system.

mgruner:
Ok, understood, thanks. Can you please add the new config setting close to the existing Core::Sendmail settings? And please also extend a unit test or add a new one to cover this behaviour. Then I'll merge the  feature for OTRS 6. Best regards, Martin

reneeb:
Thanks, I will do the needed changes...

reneeb:
replaced by PR #982

PR #762, opened 2015-10-09
[ Rel-5_0 Bug 11320 ] Invalid customer user can update password

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11320

Hello @mgruner ,

Here is proposal for fixing this bug. Used already existed security feature that get activated if customer try to request password for non existed customer. This way invalid customer will not receive email to reset there password but there will also be no notification that they are invalid for security reasons.

Please let me know what do you think about this suggestion, if you think there should be some additional message when customer try to request password reset for invalid customer, we can add some notification or some other message to follow that action.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/Web/InterfaceAgent.pm
+++ b/Kernel/System/Web/InterfaceAgent.pm
@@ -636,7 +636,11 @@ sub Run {
             User  => $User,
             Valid => 1
         );
-        if ( !$UserData{UserID} ) {
+
+        # verify user is valid when requesting password reset
+        my @ValidIDs = $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet();
+        my $UserIsValid = grep { $UserData{ValidID} == $_ } @ValidIDs;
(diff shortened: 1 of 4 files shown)


mgruner:
Hi Sanjin, could you add new selenium test cases for this? You can even intercept the emails with the test backend so that the complete password reset chain can be tested (also for agents).

s7design:
Hi @mgruner , @mrcbnsls 

Created selenium tests for agent and customer password recovery. Scenario is to request first password recovery when user is valid and later repeat same process for invalid user. Intercepting emails, as you suggested, with test backend.

Please let me know if there are any questions or issues regarding this commit.

Regards,
Sanjin

mrcbnsls:
Merged in 6257e66. Thanks! Please see my comment on the related PR for rel-4_0, @s7design.

mgruner:
Very nice. Thanks for the wonderful new tests @s7design!

PR #772, opened 2015-10-15
[ Rel-5 Bug 11528 ] - Group Name like 'test::Support::1st Line' is not possible to use for group restrictions

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11528

Hi @mgruner ,

Hereby is proposal for fixing this issue. Added check when creating group to deny any names that contain ":" (colon) in it. This way in case that reporter mentioned, there will be no issue if group is used in restriction context.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/TicketMenu/Generic.pm
+++ b/Kernel/Output/HTML/TicketMenu/Generic.pm
@@ -1,5 +1,5 @@
 # --
-# Copyright (C) 2001-2015 OTRS AG, http://otrs.com/
+# Copyright (C) 2001-2016 OTRS AG, http://otrs.com/
 # --
 # This software comes with ABSOLUTELY NO WARRANTY. For details, see
 # the enclosed file COPYING for license information (AGPL). If you
@@ -93,8 +93,17 @@ sub Run {
         ITEM:
         for my $Item (@Items) {
 
-            my ( $Permission, $Name ) = split /:/, $Item;
+            # get index for first colon (:) into the item
+            my $IndexSeparator = index ($Item, ':');
 
+            # take the part for the permission from string
+            my $Permission = substr($Item, 0, $IndexSeparator);
+
+            # increase the index for skipping the seperator
+            $IndexSeparator++;
+
+            # get the Group name
+            my $Name = substr($Item, $IndexSeparator);
             if ( !$Permission || !$Name ) {
                 $LogObject->Log(
                     Priority => 'error',
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -120,6 +120,15 @@ sub GroupAdd {
         }
     }
 
+    # check group name
+    if ( $Param{Name} =~ m{ : }xms ) {
+        $Kernel::OM->Get('Kernel::System::Log')->Log(
+            Priority => 'error',
+            Message  => "Can't add group! Don't use : (colon) in group name!",
+        );
+        return;
+    }
+
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
 
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -251,6 +251,15 @@ sub GroupUpdate {
         }
     }
 
+    # check group name
+    if ( $Param{Name} =~ m{ : }xms ) {
+        $Kernel::OM->Get('Kernel::System::Log')->Log(
+            Priority => 'error',
+            Message  => "Can't update group! Don't use : (colon) in group name!",
+        );
+        return;
+    }
+
     # set default value
     $Param{Comment} ||= '';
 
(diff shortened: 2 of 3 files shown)


mgruner:
Hello Sanjin,
this looks strange. What exactly is the problem at hand, why does the restriction not work correctly?
A change like the one you did would always require unit tests to be added/modified, but right now I am not convinced that this is the best solution to the problem.
Regards, Martin

s7design:
Hi @mgruner  and @carlosfrodriguez 

I updated PR and added test for this check in function  GroupAdd and GroupUpdate. 

As you can see in the bug report. There is problem if you want make restriction per group. It is reported for Ticket::Frontend::MenuModule###470-Spam. However, there is a problem generally. Because, there is format for restriction by group is something like 'Permission:Group;' 
There is the problem similar to one in Queue and SubQueue. Our intention was to forbid names which contains column ':'
Maybe it is not the best solution, but it is our proposal in this moment.

Please let me know what do you think about all of that. If you have another solution we will be able the change this PR.

Regards
Zoran

carlosgarciac:
Hi @s7design,

The fix looks fine for me, but IMO we are not really solving the problem, we are trying to avoid it, we could check why this restriction is not working correctly, then look for the best option here. 

For Queues is a different case due '::' is used to have the 'queue::subqueue::subsubqueue' structure.

Please let me know your comments about it.


Best regards,
Carlos García

nilei:
I think "::" should be a reserved string in OTRS for substructure purposes. Existing setups could be automatically changed to "_" or similar.

s7design:
Hi @carlosgarciac 

Thanks for follow up. As I mentioned in the previous comments, we actually know why this restriction is not working correctly. There is problem in checking permission for action per group. I explained the problem above, but I will try to add more details.

In the bug report there is issue for  Ticket::Frontend::MenuModule###470-Spam, but the problem is for all ticket action where there are permission restriction.

<pre>
Ticket::Frontend::MenuModule###470-Spam
* Shows a link in the menu to set a ticket as spam in every ticket overview of the agent interface. 
* Additional access control to show or not show this link can be done 
* by using Key "Group" and Content like "rw:group1;move_into:group2". 
</pre>

The problem with ':' in group name occurs because of the logic which there is for restriction for permission in the groups. 

You can see in the code:
https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/TicketMenu/Generic.pm#L96

In this moment I believe that this solution is the best one. My intention was that we solve case that is mentioned in the bug report. Of course we could change logic of how to set  group/permission restrictions. If you have any suggestion please  let me know and we will be able to do that.

To be honest, the case with Queue, Service and this one is very similar or the same. I could say that all of that cases are not the best possible solution. Each of them could make a problem by parsing the name/string and after parsing should decide if there is entity or sub-entities. I know that all of you have more experience, and you have known that without my mentioning. It could be topic for thinking in future and maybe possible improvement.

If you prefer solution that we change logic of restriction for ticket action, please suggest me that or any other advise is welcomed. I will agree with any proposals, but in that moment your help is necessary in case I don't waste time with proposals which are not acceptable in current situation.

I believe we will be able to solve this as well as it is possible. It is not so big problem. We solved the biggest ones, it will be piece of cake :)

Regards
Zoran

carlosgarciac:
Hi @s7design, 

Discussing this issue with @mgruner we agree '::' string should be allow on group's name, due it is used just as a separator for sub-items on other entities in the system.

Then I think we could fix the root problem for the referenced bug on this PR. please take a look on next piece of code and let me know your comments.

diff --git a/Kernel/Output/HTML/TicketMenu/Generic.pm b/Kernel/Output/HTML/TicketMenu/Generic.pm
index 400d87b..27408d1 100644
--- a/Kernel/Output/HTML/TicketMenu/Generic.pm
+++ b/Kernel/Output/HTML/TicketMenu/Generic.pm
@@ -93,7 +93,16 @@ sub Run {
         ITEM:
         for my $Item (@Items) {
 
-            my ( $Permission, $Name ) = split /:/, $Item;
+            # my ( $Permission, $Name ) = split /:/, $Item;
+
+            # get index for first colon (:) into the item
+            my $IndexSeparator = index ($Item, ':');
+            # take the part for the permission from string
+            my $Permission = substr($Item, 0, $IndexSeparator);
+            # increase the index for skipping the seperator
+            $IndexSeparator++;
+            # get the Group name
+            my $Name = substr($Item, $IndexSeparator);
 
             if ( !$Permission || !$Name ) {
                 $LogObject->Log(


Best regards,
Carlos García

s7design:
Hi @carlosgarciac 

I tested your solution it works for me, I think it could be fixed on that way. I committed this solution here in this PR, but I suggest closing this PR and make another one with Selenium test which will reproduce test case that the reporter mentioned in the bug report.

Regards
Zoran

carlosgarciac:
Hi @s7design,

create a new PR sound fine for me, could you please create that new PR, then I will close this one with a references to the new one, in order to don't lose the reference to conversations.

Best regards,
Carlos García

carlosgarciac:
Hi @s7design,

I included the fix in d92e58f, thanks a lot for your big effort.

Best regards,
Carlos García

PR #773, opened 2015-10-15
[Master - bug 11553 ] In some languages the buttons to create a statistics have diffrent heights

s7design:
Hi @mgruner,

Here is our proposal for fixing. There was a problem if some of texts are shorter. It could be a problem also if some of translated text is longer. However it is esthetic, not functional issue.
I think that there could be welcomed @mrcbnsls  opinion. 

Because this is proposal I have done in master, but if you want I will be able to provide PR for OTRS5.

Regards
Zoran


mrcbnsls:
Hi @s7design, thanks for your proposal. I must admin that I didn't like the way I implemented this from the very start, so I now resolved my own mistakes and solved it in the right way in 3beb02f7dce45a13f9a2534377f5681c8fed5197. Thanks anyway!

PR #776, opened 2015-10-16
[ Master - Bug 11365 ] - ACLs don't work in the customer interface

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11365

Hi @carlosfrodriguez ,

Hereby is proposal for fixing this issue. Previously there was no ACL restriction for actions in CustomerTicketZoom, so added that. Currently only CustomerTicketPrint action is restricted with ACL, but there is left space, if need be, to restrict other customer action in future. 

Additionally cleaned some typos in code.

Please let me know what do you think about this suggestion and PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/CustomerTicketZoom.pm
+++ b/Kernel/Modules/CustomerTicketZoom.pm
@@ -94,12 +94,12 @@ sub Run {
     }
 
     my $ACL = $TicketObject->TicketAcl(
-        Data          => \%PossibleActions,
-        Action        => $Self->{Action},
-        TicketID      => $Self->{TicketID},
-        ReturnType    => 'Action',
-        ReturnSubType => '-',
-        UserID        => $Self->{UserID},
+        Data           => \%PossibleActions,
+        Action         => $Self->{Action},
+        TicketID       => $Self->{TicketID},
+        ReturnType     => 'Action',
+        ReturnSubType  => '-',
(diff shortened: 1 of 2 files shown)


Kernel/Modules/CustomerTicketZoom.pm
```diff
@@ -75,10 +75,50 @@ sub Run {
...
+            sort keys %Actions;
+    }
+
+    my $ACL = $TicketObject->TicketAcl(
+        Data          => \%PossibleActions,
+        Action        => $Self->{Action},
+        TicketID      => $Self->{TicketID},
+        ReturnType    => 'Action',
+        ReturnSubType => '-',
+        UserID        => $Self->{UserID},
```
carlosfrodriguez: @s7design  Maybe we want to use here CustomerUserID instead, isn't it?

((enjoy))
cr

s7design:
Hi @carlosfrodriguez ,

Thank you for pointing it out. You are right it should be CustomerUserID. Updated PR accordingly.

Regards,
Sanjin

carlosfrodriguez:
Hi @s7design,

This pull request was merged in https://github.com/OTRS/otrs/commit/cd7f376cb5396f3156a5021deb85f00841f4e406. 

It was only implemented for OTRS 6 since it is more like a missing feature rather than a bug. And also because its complexity.

((enjoy))

PR #780, opened 2015-10-17
Add new config options to define hour and minute for pending time

reneeb:
Until now one can define the a time difference for the pending time. But in many cases, one want to set the pending time to the start of the office times to be reminded at the start of the work day.

This pull request adds exactly this. One can define the hour and minute of the pending time.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -11470,4 +11470,20 @@ Thanks for your help!
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::Frontend::PendingTimeHour" Required="0" Valid="0">
+        <Description Translatable="1">Exact hour for pending time.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Frontend::Agent</SubGroup>
+        <Setting>
+            <String Regex="^\d+$">8</String>
+        </Setting>
+    </ConfigItem>
+    <ConfigItem Name="Ticket::Frontend::PendingTimeMinute" Required="0" Valid="0">
+        <Description Translatable="1">Exact hour for pending time.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Frontend::Agent</SubGroup>
+        <Setting>
+            <String Regex="^\d+$">00</String>
+        </Setting>
+    </ConfigItem>
 </otrs_config>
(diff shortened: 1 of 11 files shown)


xhoy:
+1!

mgruner:
@nilei what's your opinion on this?

jensoliver:
From my side it looks useful.

nilei:
@mgruner, I understand the demand, but I think the actual issue is that selecting the right date and time is too cumbersome. I also do not fully understand how this new settings plays together with the old setting.

I would recommend to ease the date and time selection (not only for pending date and time), so one can e.g. jump to the beginning or end of the next office day. Bootstrap has nice date and time picker: https://eonasdan.github.io/bootstrap-datetimepicker/

reneeb:
@nilei the date is calculated with the old settings, only the hour and the minute is set with the new settings. Currently you just say "pending time is in 'now' in two days", with the new settings you say "It's in two days, but at 8 o'clock".

To ease the date/time selection is not the correct solution (but might be helpful anyways), because the agent would still have to set the correct hour/minute him/herself.

nilei:
@reneeb, as said, I understand the demand and yes, it is useful!

Setting a global setting for users will always affect people positive and others negative. Turning this setting into a user preference or adding (custom) buttons (or similar) next to the pending time is more satisfying, even if you still need to do one click (better than 4 or more clicks as today).

@mgruner, I have no issues with merging, we may find a way to turn this into a user preference later :-)

reneeb:
@mgruner Any news about the PR?

nilei:
@reneeb, do you think you could enhance this to be pending state specific?

Example for **PendingTimeHour**:
- pending reminder => 8
- pending auto close+ => 17

reneeb:
That should be possible... @mgruner what do you think about that?

mgruner:
Hello all, and special thanks to @reneeb for the proposal!

Honestly, I am not convinced. The real problem is that the PendingDiffTime selection is too static and does not take into account the configured business hours. The current patch does not really solve this problem. If you are on a friday, you will still be presented with Saturday 8:00 even though this is not a working day.

@nilei I don't agree with making it even more specific, state dependent. There are far more criteria (priority etc.) that one could want to depend on.

To really fix this we'd probably need a different way of configuring the pending time default selection. I'm open for all suggestions.

Best regards, Martin

reneeb:
@mgruner: It will *not* present Saturday 8:00 as the date is still calculated with respect to business hours. It's just the hour and the minute setting that is changed...

mgruner:
@reneeb are you sure? AFAICT currently the PendingDiffTime is just 24h ahead, ignoring any configured business hours.

nilei:
@mgruner, your are absolutely right, making it depending on the state is not a good enhancement, sorry for that :-)

reneeb:
Argh, @mgruner you're right. Sorry for the noise...

PR #783, opened 2015-10-21
[Master Bug 11574] - Dashboard - Queue - Attributes of the Widget are not used in the 'Tickets in my Service' view

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11574

Hi,

Here is our proposal for fixing. We solved it similar to if Queue is added in SySconfig for param Attributes.
Please, let me know what do you mean about that.

P.S. We are planning to provide new PR with our suggestion for solving same casees if in Attributes is added Owner, Responsible, WatchUser or something like that. We will add in another PR because it is only our proposal for now, and this is not subject of this bug.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
@@ -2116,7 +2116,7 @@ sub _SearchParamsGet {
 
     if ( defined $TicketSearch{QueueIDs} || defined $TicketSearch{Queues} ) {
         delete $TicketSearchSummary{MyQueues};
-        delete $TicketSearchSummary{MyServices};
+        delete $TicketSearchSummary{MyServices}{QueueIDs};
     }
 
     if ( !$Self->{UseTicketService} ) {
--- a/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
@@ -2116,6 +2116,7 @@ sub _SearchParamsGet {
 
     if ( defined $TicketSearch{QueueIDs} || defined $TicketSearch{Queues} ) {
         delete $TicketSearchSummary{MyQueues};
+        delete $TicketSearchSummary{MyServices};
     }
 
     if ( !$Self->{UseTicketService} ) {


carlosfrodriguez:
Hi @s7design,

Thank you for your pull request, I believe that the solution is wrong, deleting the "MyServices" attributes will prevent to see only the ticket that are set for the preferred services of the agent, I think that the solution should be some lines above in the code

```perl
        MyServices => {
            QueueIDs   => \@ViewableQueueIDs,
            ServiceIDs => \@MyServiceIDs,
            LockIDs    => $TicketSearch{LockIDs} // undef,
        },
```

Here the my services are calculated using all queues where the agent has ro access and then this is why it is not restricted to only the queue that is set in the attributes, maybe changing the  \@ViewableQueueIDs could help.

((enjoy))
cr

s7design:
Hi @carlosfrodriguez,

Thanks for follow-up. It is so obviously now, with your explanation :) It is definitely better solution and I can say the one right solution. I will update PR tomorrow.

Regards
Zoran

s7design:
Hi @carlosfrodriguez 

I updated and tested it again, I suppose it is fine now.

Regards
Zoran

s7design:
If you have marge conflict with PR https://github.com/OTRS/otrs/pull/786 please let me know I will be able to rebase it.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

I have the fear, that with current solution it will bring all the tickets where they have a service that is in the list of my services for the user regardless the queue, also even list tickets on queues that the user does not ha permissions... can you check this?

s7design:
Hi @carlosfrodriguez 

I tested such case and it works for me well, as it is expexcted. 

My test case is:
1. create some tickets in Posmaster queues
2. change Attributes  of  one dashboard wiget or create new one - Atributtes=> Queues=Postmaster;
3. set Posmaster to admin group and create two agent one with full permmisions and another with user permision

Result is:
* agent with admin permision can see list of tickets
* agent  with user permision don't see any tickets in this widget

To be honest such possibility in SySconfig for dashboard widget could be so tricky in reality. There are and there were some bugs similar to this one, because there is able to add many types of filters and they could make a problems. My opinion is that it is better to fix as many as possible cases.
Feel free to tell me if you have any suggestion how I can improve this.

Regards
Zoran

carlosfrodriguez:
Hi @s7design,

This pull request was merged in https://github.com/OTRS/otrs/commit/037b61df881e48373d91ab114d9a060b97fdb38f

You where right all other params remains, and removing Queues within the MyServices hash fix the issue correctly, even for the required permissions.

((enjoy))
cr

PR #784, opened 2015-10-21
Normalized HTML structure.

thorsteneckel:
The HTML strucutre got somehow away from the OTRS standard HTML layout / structure. This PR contains the following changes to make it OTRS standard conform again:

- inflationary usage of fieldsets
- p class="FieldExplanation" have to be childs of a div class="Field".
- Translated "Text will also be received by:" is followed by a hardcoded "':" 
- p class="Value" should be a div class="Field"

It would be great if you could apply these changes to the rel-5_0 branch, too.

Thanks to @rolfschmidt


rolfschmidt:
++

mrcbnsls:
Hi @thorsteneckel, please see my comment in the referenced commit. Thanks!

thorsteneckel:
Thanks @mrcbnsls for your feedback and the insights. @rolfschmidt and I will provide an updated PR. :+1:

PR #785, opened 2015-10-21
[ Rel-5_0 Bug 11565 ] - New article notification(star) and entry created for agent that creates the ticket

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11565

Hi @carlosfrodriguez ,

Hereby is proposed solution for this reported issue. Added sysconfig that can control if agent want to receive 'flags' for there own action. Now agent who create articles will be flagged as seen upon article creation. Ticket is flagged just for that agent as well then.  

If there are any question or issues regarding this PR, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -88,6 +88,17 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::NewArticleIgnoreCurrentAgent" Required="1" Valid="1">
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @s7design. I don't agree with your solution; this is nothing that needs to be configurable because the logic is clear. The problem was caused by my commit (pointed out by Dennis in the bug) where I tried to make a pre-mature optimization. I don't understand how your proposal should actually fix the problem at hand. Therefore I decided to just roll back my original commit.
Thanks anyway for your efforts on this! Regards, Martin

PR #787, opened 2015-10-22
[ Rel-5_0 Bug 9884 ] Solution Time counts wrong on daylight time changes

s7design:
http://bugs.otrs.org/show_bug.cgi?id=9884

Hi @carlosgarciac ,

Hereby is proposed solution to fix this issue. Destination time now is calculated correctly when there is a switch for daylight saving time for America/Sao_Paulo time zone. Created unit test with various scenarios for testing destination time in different time zones.

If there are any question or issues regarding this PR, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/Time.pm
+++ b/Kernel/System/Time.pm
@@ -682,7 +682,7 @@ sub DestinationTime {
     );
 
     my $LoopCounter;
-    my $NextDay;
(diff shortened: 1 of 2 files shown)


mgruner:
Merged, closing, thanks a lot!

PR #788, opened 2015-10-23
[ Rel-5_0 Bug 11575 ] - Logo overwrites toolbar 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11575

Hi @mrcbnsls 

Here is our proposal for fixing this bug. After testing it we saw that the issue was happened during re-adding #ToolBar. In the Header.tt at first there is #Logo and then #ToolBar. Because of that Logo in slim skins caused such problem. In OTRS4 there is not this problem. 

I saw another problem here. The fix in Core.Agent.Responsive.js 

<pre> $('#Logo').detach().prependTo('#Header'); </pre>
was not enough. On reloading the page, width of navigation bar (z-index=20) is 98% so in slim skins it is above the toolbar menu (z-index=10). In that case it is not possible to click on items of toolbar menu ( actually it is possible on top of any item, if you have enough patience :) )
I saw also that after resizing the screens the menu items are available again. It is solved with Core.Agent.ResizeNavigationBar. However, the problem was with checking the class 'Visible-ScreenXL'. I removed this checking and it works fine now. I don't know is it will cause some bad side effect.

Please, follow up the PR and let me know what is your opinion about that. 

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.Responsive.js
+++ b/var/httpd/htdocs/js/Core.Agent.Responsive.js
@@ -198,6 +198,7 @@ Core.Agent.Responsive = (function (TargetNS) {
 
         // re-add toolbar to header
         $('#ToolBar').detach().prependTo('#Header');
+        $('#Logo').detach().prependTo('#Header');
 
         // reset field widths
         $('.FormScreen select').each(function() {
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Hi Zoran (@s7design), thanks for the PR, but I preferred to solve it in another way. Please have a look at the commit IDs from the bug entry. I'll close this PR.

PR #789, opened 2015-10-23
[ Rel-5_0 Bug 5946 ] - The ticket count for closed tickets in status view is blocked to 10000

s7design:
http://bugs.otrs.org/show_bug.cgi?id=5946

Hello,

Hereby is proposal for solving this issue. It was hard-coded on 10_000 tickets to be shown in view screens. Added sysconfig that now gives users a choice to set up amount of tickets they would like to see in there overview screens.

if there are any questions or issues regarding this PR, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -1410,6 +1410,14 @@
             <String Regex="^[0-9]$">5</String>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::Frontend::Overview::TicketShownLimit" Required="1" Valid="1">
+        <Description Translatable="1">Set the limit of tickets that will be shown in view screens.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Frontend::Agent::TicketOverview</SubGroup>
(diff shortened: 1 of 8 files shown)


carlosfrodriguez:
Hi @s7design,

Does this is really needed and manageable?

s7design:
Hi @carlosfrodriguez,

It is one possible solution this bug. Maybe it could be interesting for some system and somebody can set this SysConfig less or more depend of situation in specific system. It is my opinion.  I could agree that maybe it is not so vital.
If we don't wont fix this in this moment we can close the bug report with some explanation.  Please, let me know what do you mean about that.

Regards
Zoran

carlosgarciac:
Hi @s7design,

I just closed the bug including an explanation, IMO we could think more than a bug, an enhancement that  could show us we pass the limit (or almost) for the visible entries on the different views, perhaps a SupportDataCollector pluging that shows how much closed tickets we have ( we already have one for open tickets) or other that shows if you already pass the count for each ticket view, or a different approach.

Please let me know your opinion on this.


Best regards,
Carlos García

PR #791, opened 2015-10-25
add a module to make it easier to add new dialogs

reneeb:


Kernel/Config/Files/FlexibleModule.xml.dist
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
eliasp: This should probably be `encoding="utf-8"`

reneeb:
@mgruner what do you think about this PR?

mgruner:
Hello @reneeb, and thanks a lot for your proposal. Am I right that you basically copied AgenTicketActionCommon and provided the FlexiTicket.xml.dist for the user to adapt it to their needs?

This is definitely an interesting approach, but seems quite fiddly to get the new configurations working correctly. Besides that, new code redundancy is introduced. Therefore I would not want to see this in the framework at this point.

We are nevertheless interested in making the GUI less static and more flexible / configurable. But preferrably without having to touch source files. This would probably lead to major changes.

Right now we have one more static part (the regular ticket dialogs) and one extremely flexible part (ProcessManagement). The first one is rather easy to use and works out of the box, while the second one is complex to configure but also very powerful. We can talk about making the regular dialogs more flexible, but then we should also refactor and unify the existing code.

Since this is open source software, you are free to publish this as your own extension (in case you haven't done so yet).

Best regards, Martin

PR #792, opened 2015-10-26
Fix bug#11586 - Cannot show collapsed view by default if config Ticket::Frontend::ZoomExpand is true

noritnk:
This PR is fix bug#11586.
See http://bugs.otrs.org/show_bug.cgi?id=11586


mrcbnsls:
Thanks a lot. Had to add some changes because of the Timeline view of the **OTRS Business Solution**™. See bug entry for the commits.

PR #734, opened 2015-09-29
[Rel-5_0 - Bug 10853 ] - State and StateType search attributes in customer info are not shown in Search screen

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10853

Hello @mgruner ,

Hereby is updated PR with fix proposal for this bug. Updated Selenium tests to check for new links and search attributes.
Please let me know if there are any issues or questions regarding this PR.

Regards,
Sanjin

PR #794, opened 2015-10-27
[Rel-5_0 Bug 11593] - Inconsistency in ticketsearch of the customer information center

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11593
Hi @carlosfrodriguez and @mgruner,

Here is our proposal for fixing this bug. There was problem because some widget have attribute StateType=open, but for example, in Company Status widget in CIC is used StateType=Open. We could fix it on two ways: 
* we can change SysConfig param (DashboardBackend###0130-TicketOpen, AgentCustomerInformationCenter::Backend###0130-CIC-TicketOpen) or
* adapt all modules that use 'open' instead 'Open' ( AgentDashboardCustomerIDStatus, AgentDashboardCustomerUserList, CustomerUserList, TicketGeneric ...)
The same is with StateType=close and StateType=Close.

We can agree that the both cases is possible, however I believe that there was intention that there  is behavior of system where there will be listed all open tickets in sens of  explanation in TicketSearch
<pre>
    # NOTE: Open and Closed are not valid state types. It's for compat.
    # Open   -> All states which are grouped as open (new, open, pending, ...)
    # Closed -> All states which are grouped as closed (closed successful, closed unsuccessful) 
</pre>

P.S.
In some hand, this bug and PR is connected to this one: https://github.com/OTRS/otrs/pull/734. It is not a duplicate, it is only connected to this one. 

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -6776,7 +6776,7 @@
                 <Item Key="Module">Kernel::Output::HTML::Dashboard::TicketGeneric</Item>
                 <Item Key="Title" Translatable="1">Open Tickets / Need to be answered</Item>
                 <Item Key="Description" Translatable="1">All open tickets, these tickets have already been worked on, but need a response</Item>
-                <Item Key="Attributes">StateType=open;</Item>
+                <Item Key="Attributes">StateType=Open;</Item>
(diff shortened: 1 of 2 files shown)


mgruner:
Hi Zoran, please see my comment in the bug. I believe this works as intended. Best regards, Martin

PR #795, opened 2015-10-27
[Rel-5_0 Bug 4640] - Wrong variable <OTRS_CUSTOMER_REALNAME> in Auto Responses

s7design:
http://bugs.otrs.org/show_bug.cgi?id=4640
http://bugs.otrs.org/show_bug.cgi?id=11590

Hi @carlosfrodriguez ,

Hereby is proposed solution for this bug. For fetched mails that have ReplyTo, <OTRS_CUSTOMER_REALNAME> tag will now correctly take value. 

Additionally bugs nr: 7521 ( http://bugs.otrs.org/show_bug.cgi?id=7521, http://bugs.otrs.org/show_bug.cgi?id=11590 ) are also fixed. When there are no registered customer in DB that send email, <OTRS_CUSTOMER_REALNAME> tag will take value from 'From' instead of 'To'.

Please let me know what do you think about these proposed solutions.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -1529,9 +1529,14 @@ sub _Replace {
                 # try to get the real name directly from the data
                 $From //= $Recipient{RealName};
 
+                # get real name based on reply-to
+                if ( $Data{ReplyTo} ) {
+                    $From = $Data{ReplyTo} || '';
+                }
+
                 # generate real name based on sender line
                 if ( !$From ) {
-                    $From = $Data{To} || '';
+                    $From = $Data{From} || '';
 
                     # remove email addresses
                     $From =~ s/&lt;.*&gt;|<.*>|\(.*\)|\"|&quot;|;|,//g;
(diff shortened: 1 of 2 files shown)


mgruner:
I would definitely wan to see unit tests for changes of this level.

s7design:
Hello @mgruner ,

Created unit test for this bug where two emails are fetched through console command. One with Reply-To tag and other one without it. In test you can see different behaviour for 
<OTRS_CUSTOMER_REALNAME>  tag in auto response article.

Additionally you can see that tag To in auto response article is reading different when there is no Reply-To tag in email for unknown customers. Before fix it would take tag To that was during initial fetching of email.

Please let me know if there are any questions or issues with unittest.

Regards,
Sanjin

mrcbnsls:
Merged in 660302f and 82d55d6.

The UT looks good, but I could see that currently only two cases are checked (which is correct for this fix): Reply-To not present and Reply-To present. But as I could see from the code, there are actually four different cases which are influencing the value of $From:

(1) Filled from customer data from the ticket
(2) Filled from $Recipient
(3) Filled from Reply-To
(4) Filled from From

Could you extend the UT to also check for the other cases if possible? This could go into a new PR. Thanks!

Marc

PR #796, opened 2015-10-28
In notifications, use HTML body only if RichText is enabled.

otetard:
If `$Data{HTMLBody}` exists, but RichText is not enabled, don’t include the HTML body. This will prevent from displaying HTML tags in text/plain notifications.

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -1418,7 +1418,7 @@ sub _Replace {
 
                 my $Line = $2 || 2500;
                 my $NewOldBody = '';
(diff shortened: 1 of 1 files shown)


mgruner:
Hi @carlosgarciac, can you review please? Please add tests.

otetard:
Unfortunately, my patch is not complete: I still receive notifications for follows up in raw HTML.

carlosgarciac:
Hi @otetard,

HTMLBody is filled if AttachmentIDOfHTMLBody is present for the last customer article, it doesn't depends of RichText.

I hope this helps.

Best regards,
Carlos García

PR #797, opened 2015-10-28
[Rel-5_0 Bug 11327] - A dyn. DropDown field is not sort alphapetical, if you activate the treeview

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11327
Hi @carlosfrodriguez,

Here is our proposal for fixing, it is tested with TreeView and without  it, and without Modernize fields as well. All works fine now.
If you need master or OTRS4 PR please let me know.

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.TreeSelection.js
+++ b/var/httpd/htdocs/js/Core.UI.TreeSelection.js
@@ -524,13 +524,13 @@ Core.UI.TreeSelection = (function (TargetNS) {
 
         SelectData.sort(function(a, b) {
 
-            var KeyA = a.Key.toLowerCase(),
-                KeyB = b.Key.toLowerCase();
+            var ValueA = a.Value.toLowerCase(),
+                ValueB = b.Value.toLowerCase();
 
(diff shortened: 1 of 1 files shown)


zottto:
Merged with 464ab8f (OTRS 4), 33d0029d8be5075f634a823510e5d25a5aa99488 (OTRS 5) and 4bd4caaf364873cea70ee43011e6e707869b5d05 (Master).

zottto:
Hi @s7design,

I had to revert the commits from yesterday. As Carlos pointed out in the QA, if you have multilevel entries it is necessary to sort for keys instead of values to keep the functionality of multiple levels.

Perhaps you can fix it in another way: after sorting for the keys there could be a second sorting, which sorts after values on each level independently (recursively).

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for the suggestion, we will try to implements you suggested.
I reopen my task for this as well, I suppose it is not top priority. We will work on this after tasks which we are doing now.

Regards
Zoran

zottto:
Closing for the moment. Feel free to reopen.

PR #798, opened 2015-10-29
[Rel-5_0 Bug 11604] - AgentTicketForward is still using the "old" select fields style

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11604

Hello @mgruner ,

Hereby are added modernized fields for AgentTicketForward as Nils reported. Additionally found out that in AgentTicketBounce there was left 'old' field for state. Felt free to change that one as well.

Regards,
Sanjin


mrcbnsls:
Thanks, merged in 7ae0a29.

PR #799, opened 2015-10-30
[Rel-5_0 Bug 11592] - NotificationArticleType ignored at notification for customer

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11592

Hello @mgruner ,

Hereby is proposed solution for fixing reporters issue. When recipient for notification is customer and there is no additional email address ( $Recipient{NotificationArticleType} ), article type was hardcoded on 'email-notification-ext'.

With this proposal, article type is now correctly reading from notification data, and resolving for 'email-notification-ext' as last resort. In testing now it is possible to set article type as 'email-notification-int' while sending customer notification without additional email addresses.

Please let me know what do you think about this proposal. Other solution is to not allow internal notification type for customer and change description message for that field.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/Ticket/Event/NotificationEvent/Transport/Email.pm
+++ b/Kernel/System/Ticket/Event/NotificationEvent/Transport/Email.pm
@@ -225,7 +225,12 @@ sub SendNotification {
             %Address = $QueueObject->GetSystemAddress( QueueID => $Ticket{QueueID} );
         }
 
-        my $ArticleType = $Recipient{NotificationArticleType} || 'email-notification-ext';
+        # get notification article type
+        my $NotificationArticleType = $TicketObject->ArticleTypeLookup(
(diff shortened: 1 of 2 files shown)


mgruner:
Please extend the unit tests (Ticket/Event/NotificationEvent.t?) for this.

s7design:
Hi @mgruner ,

Modified unit test to check for bug fix. Created one more notification where customer is recipient and notification article type is set to email-notification-int. Test will check if there are article created with that type in ticket.

Please let me know if there are any issues or questions regarding this unit tests.

Regards,
Sanjin

carlosgarciac:
Hi @s7design,

Your PR was merged with a very small change on 89d69cb.

Thanks a lot for your hard work.


Best regards,
Carlos García

PR #801, opened 2015-11-02
add new search blocks in toolbar

reneeb:


Kernel/Output/HTML/Templates/Standard/Header.tt
```diff
@@ -89,6 +103,28 @@
...
+            <li class="Extended SearchFulltext" style="width: 400px !important">
+                <form action="$Baselink" method="post" name="SearchFulltextStateSearch">
+                    <input type="submit" style="position: absolute; height: 0px; width: 0px; border: none; padding: 0px" tabindex="-1" name="submit" value="submit" hidefocus="true" />
+                    <input type="hidden" name="Action" value="AgentTicketSearch"/>
+                    <input type="hidden" name="Subaction" value="Search"/>
+                    <input type="hidden" name="SearchTemplate" value="$Search"/>
+                    <input type="hidden" name="CheckTicketNumberAndRedirect" value="1"/>
+
+                    <select style="display:none;" name="StateIDs" multiple="multiple" id="StateIDs" size="5">
+                          <option selected="selected" value="1">new</option>
```
mgruner: Hm, hard-coded...

mgruner:
Hello Renée, thanks a lot for your proposal, but I currently fail to see that this would be a generic requirement. The CIC usually is the place for customer specific information. Please feel free to reopen with more information!

reneeb:
Searching a customer ticket via CIC would be an additional step. This helps a lot when you get a call from a customer and want to search for a ticket...

xhoy:
We use the CiC alot but in some cases it's far from ideal. For example we use company tickets a lot so we get this a lot. for example:
* Customer: Calls "i cannot print my email my colleage had the same with his email some days ago
* Agent: who was it? 
* Customer: dunno i heard it from X? maby Y?
* Agent looks at CiC And sees 50 open tickets... on 9 different users, 0 on x and y....
* agents go to search bar, typs in customerID (company id, a 4digit number) and search word "email" 1 open ticket is left with user z.
**agents goes on to fix the issue**

And we could solved this differently but "hear say" is a real issue with customers, they often haven't got a clue what is wrong and why they have issues. So we cannot use services (since a printer OR email service issue. In this case they hadn't got an lan connection... (cable disconnect by cleaning person....)

Good luck finding that...

PR #802, opened 2015-11-02
In AgentTicketBulk, merge first, and do the other actions only on the main ticket

moritz:
Some of our users complained that using a bulk action to merge several tickets
into one and setting the owner (for example) at the same time sends a bunch of
redundant notifications.

Merging a ticket negates some changes (for example setting ticket state),
and make others (such as owner changes) irrelevant. So do it first, and
save the other actions for the ticket into which was merged.

If this isn't desirable in the general case, I could make this new behavior
configurable.

### DIFF
--- a/Kernel/Modules/AgentTicketBulk.pm
+++ b/Kernel/Modules/AgentTicketBulk.pm
@@ -471,6 +471,20 @@ sub Run {
             # challenge token check for write action
             $LayoutObject->ChallengeTokenCheck();
 
+            # merge first, and do all the other actions only
+            # on the main ticket, to avoid a bunch of notifications
+            # for changes which, in the end, would be undone by the merge
+            # anyway
+            if ( $MainTicketID && $MainTicketID ne $TicketID ) {
+                $TicketObject->TicketMerge(
+                    MainTicketID  => $MainTicketID,
+                    MergeTicketID => $TicketID,
+                    UserID        => $Self->{UserID},
+                );
+                next TICKET_ID;
+            }
+
+
             # set owner
             if ( $Config->{Owner} && ( $GetParam{'OwnerID'} || $GetParam{'Owner'} ) ) {
                 $TicketObject->TicketOwnerSet(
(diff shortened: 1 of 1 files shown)


Kernel/Modules/AgentTicketBulk.pm
```diff
@@ -471,6 +471,20 @@ sub Run {
...
+            # on the main ticket, to avoid a bunch of notifications
+            # for changes which, in the end, would be undone by the merge
+            # anyway
+            if ( $MainTicketID && $MainTicketID ne $TicketID ) {
+                $TicketObject->TicketMerge(
+                    MainTicketID  => $MainTicketID,
+                    MergeTicketID => $TicketID,
+                    UserID        => $Self->{UserID},
+                );
+                next TICKET_ID;
```
mgruner: Hello Moritz, how should the code after this if() ever be reached? Doesn't ```@TicketIDs``` contain just the tickets to be merged to the main ticket?

moritz: Hi @mgruner,

the point of this pull request is that, if the ticket is to be merged into another, that the other actions aren't executed.
Say you merge some tickets, and close them. There's no point in first closing the ticket, and then merging it; it just creates notification spam, and then the merge overrides the ticket state to "merged" anyway.

Instead merge first, and only do the other actions (like closing the ticket, owner update etc.) on the main ticket.

mgruner:
Your code might work with MergeToOldest, because then the MainTicketID is part of ```@TicketIDs```, but what if MergeTo is specified with a custom TicketID? Then the code I mentioned would be dead, wouldn't it?

moritz:
Hi,

On 11/20/2015 09:57 AM, Martin Gruner wrote:
> Your code might work with MergeToOldest, because then the MainTicketID
> is part of |@TicketIDs|, but what if MergeTo is specified with a custom
> TicketID? Then the code I mentioned would be dead, wouldn't it?

Yes. What's the point of setting an owner of a merged ticket? The work 
happens in the main ticket. Likewise with the other attributes.

Or are there any attributes that still relevant on a merged ticket?

Cheers,
Moritz

nilei:
I don't think this is a good idea. We have customers who revoke meged tickets and therefore require the owner to be set. If the main reason for this pull request is "SPAM" coming from the later merged tickest, I would recommend to avoid such SPAM with a "Send notification on merge affected tickets" checkbox in order to let the agent decide. The default value for this setting might be configurable.

mgruner:
Thanks a lot @nilei for your input on this! I'll close this PR for now. We can look for another solution if you want, @moritz.

nilei:
@mgruner, I am not sure if you should close this pull request. The concern is valid, but instead of hard code the execution on the MainTicket only I recommend to enhance the pull request with a checkbox for the agents to decide if the actions should also happen on the merged tickets before merging or not.

PR #803, opened 2015-11-03
[Rel-5_0 Bug 10477] - Printing: numbering of article does not equal to the numbering of PDF

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10477

Hi @mgruner ,

Hereby is proposal for fixing this bug. Added 'ArticleNumber' for single article print link, thats taken from Article{Count}. Using this param in AgentTicketPrint, distinguishing when it's single article or whole ticket print and creating correct article numbering in PDF output accordingly.

Reporter also mentioned wrong article numbering when reverse sorting order is in action in AgentTicketZoom, but this was not reproducible in my test environment, it must have been fixed somewhere along the way. 

If there are any questions or issues regarding this PR, please let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketPrint.pm
+++ b/Kernel/Modules/AgentTicketPrint.pm
@@ -30,10 +30,11 @@ sub Run {
 
     # get ticket object
     my $TicketObject = $Kernel::OM->Get('Kernel::System::Ticket');
+    my $ParamObject  = $Kernel::OM->Get('Kernel::System::Web::Request');
 
     my $Output;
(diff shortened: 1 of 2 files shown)


mgruner:
@mrcbnsls please check if this could also affect customer ticket print.

s7design:
Maybe we can help. IMO it does not affect in  customer interface. There is not article menu. There is possible to print only entire ticket (all articles). 

Regrads
Zoran

mrcbnsls:
Thx @s7design, works fine. I added some changes to make sure the user time zone is taken into account for the date.

PR #805, opened 2015-11-04
[Rel-5_0 Bug 10319] - Link screen does not show all matching tickets for the search values

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10319

Hi @mgruner ,

Hereby is proposal for solving this issue. Created sysconfig to limit maximum number of displayed tickets in the result of search in AgentLinkObject. Left default on 50 as it was hard-coded before, leaving users choice to set limit that suits there needs.

Please let me know, what do you think about this proposal.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10115,6 +10115,14 @@ Thanks for your help!
             </Array>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="LinkObject::TicketLinkSearchLimit" Required="0" Valid="1">
+        <Description Translatable="1">Maximum number of tickets to be displayed in the result of a search in AgentLinkObject.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Core::LinkObject</SubGroup>
+        <Setting>
+            <String Regex="^[0-9]{1,5}$">50</String>
+        </Setting>
+    </ConfigItem>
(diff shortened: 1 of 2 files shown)


mgruner:
Hello Sanjin, thanks for your proposal. I have to agree with Jens; this is not a real solution for the bug. We either need a pagination (that would not be ticket specific, but generic) or extend the search mask with more fields for the ticket backend. I guess the second option would be more helpful for the user. With more fields the search result can be improved. We might need to make the ticket link backend configurable then, to decide which dynamic fields should be shown, and maybe even for the other fields too, similar to the search screen. Not sure if you want to attack this problem now, but the current proposal it not a real solution.
Regards, mg

s7design:
Hi @mgruner 

To be honest I am also not sure that we know to solve this problem in this moment, but I can say that we would like to solve it.
If you agree with me, we can try. It could be good task for us to improve our knowledge and learn something new.
In that case we should specify what is the best solution. I prefer extending search mask with more fields but also pagination, and using search limit config item as the case in AgentTicketSearch. 
However, we can discus about that by mail or Skype.

Please let me know if we can get this task. 

Regards
Zoran

nilei:
From a consulting point of view, the only "real" solution is to align the link object search with the regular search by adding all configured and enabled static and dynamic fields. Pagination is a nice benefit too if the search result is long, but with having more search options, it is unlikely that users will have thousands of search results.

PR #807, opened 2015-11-04
Master - moved connection to DB from AdminQueueAutoResponse to backend module

s7design:
Hi @mgruner 

As we spoke about moving SQL from frontend to backend I made this PR, where I solved this for AdminQueueAutoResponse. There are a couple modules with the same problem. I will be able to solve them as well.

Regards
Zoran

### DIFF
--- a/Kernel/System/AutoResponse.pm
+++ b/Kernel/System/AutoResponse.pm
@@ -342,15 +342,16 @@ sub AutoResponseGetByTypeQueueID {
 get a list of the Auto Responses
 
     my %AutoResponse = $AutoResponseObject->AutoResponseList(
-        Valid => 0,     # (optional) default 0
+        Valid   => 1,                 # (optional) default 1
+        TypeID  => 1,                 # (optional) Auto Response type ID
     );
 
 Return example:
 
     %AutoResponse = (
-        '1' => 'default reply (after new ticket has been created) ( 1 )',
-        '2' => 'default reject (after follow up and rejected of a closed ticket) ( 2 )',
-        '3' => 'default follow up (after a ticket follow up has been added) ( 3 )',
+        '1' => 'default reply (after new ticket has been created)',
+        '2' => 'default reject (after follow up and rejected of a closed ticket)',
+        '3' => 'default follow up (after a ticket follow up has been added)',
     );
 
 =cut
@@ -361,19 +362,38 @@ sub AutoResponseList {
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
 
+    my $Valid = $Param{Valid} // 1;
+
     # create sql
-    my $SQL = 'SELECT id, name FROM auto_response ';
-    if ( $Param{Valid} ) {
+    my $SQL = "SELECT ar.id, ar.name FROM auto_response ar, auto_response_type art";
+    if ($Valid) {
         $SQL
-            .= "WHERE valid_id IN ( ${\(join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet())} )";
+            .= " WHERE ar.valid_id IN ( ${\(join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet())} )";
     }
 
-    # select
-    return if !$DBObject->Prepare( SQL => $SQL );
-
     my %Data;
+
+    # if there is TypeID, select only AutoResponses by that AutoResponse type
+    if ( defined $Param{TypeID} ) {
+        $SQL .= " AND art.id = ? AND ar.type_id = art.id";
+
+        # select
+        return if !$DBObject->Prepare(
+            SQL  => $SQL,
+            Bind => [ \$Param{TypeID} ],
+        );
+
+    }
+    else {
+
+        # select
+        return if !$DBObject->Prepare(
+            SQL => $SQL,
+        );
+    }
+
     while ( my @Row = $DBObject->FetchrowArray() ) {
-        $Data{ $Row[0] } = "$Row[1] ( $Row[0] )";
+        $Data{ $Row[0] } = $Row[1];
     }
 
     return %Data;
@@ -402,7 +422,7 @@ Return example:
 sub AutoResponseTypeList {
     my ( $Self, %Param ) = @_;
 
-    my $Valid = $Param{Valid} || 1;
+    my $Valid = $Param{Valid} // 1;
 
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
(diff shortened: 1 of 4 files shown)


s7design:
@mgruner , as you can see I updated test. I think that the previus one was wromg. There was used Queue with ID 1 (Postmaster by in vanilla system), and AutoResponse with ID 1 (auto reply).

After running this test, the state of system was changed. It was happened because of the following line:

<pre>
$AutoResponseQueue = $AutoResponseObject->AutoResponseQueue(
    QueueID         => 1,
    AutoResponseIDs => [],
    UserID          => 1,
);
</pre> 

If there had been AutoResponse for the Queue Posmaster, they were deleted. The second problem was that test Auto Response, which had been created in the test, is left in the system. It was in invalid state, but I prefer that all test data, which is used in test, are deleted after testing. I think that transaction from Helper is very helpful here.

Regards
Zoran

scripts/test/AutoResponse.t
```diff
@@ -12,9 +12,40 @@ use utf8;
 
 use vars (qw($Self));
 
+use Kernel::System::UnitTest::Helper;
+
+my $HelperObject = Kernel::System::UnitTest::Helper->new();
+
+$HelperObject->BeginWork();
```
mgruner: You don't need this call and the Rollback() call at the end. Just pass ```RestoreDatabase``` to the constructor of the Helper.

Kernel/System/AutoResponse.pm
```diff
@@ -385,12 +402,192 @@ Return example:
...
+
+    %AutoResponseListByType = (
+        '1' => 'default follow up (after a ticket follow up has been added)',
+        '2' => 'auto reject',
+        '3' => 'auto follow up',
+    );
+
+=cut
+
+sub AutoResponseListByType {
```
mgruner: I would prefer it if you could add this code to AutoResponseList, with AutoResponseTypeID being a new optional parameter that filters.

Kernel/System/AutoResponse.pm
```diff
@@ -385,12 +402,192 @@ Return example:
...
+    my ( $Self, %Param ) = @_;
+
+    # check needed stuff
+    if ( !$Param{TypeID} ) {
```
mgruner: This is not a Type, but an AutoResponseType.

Kernel/System/AutoResponse.pm
```diff
@@ -385,12 +402,192 @@ Return example:
...
+        QueueID   => 2,
+    );
+
+Return example:
+
+    $AutoResponseID = 1
+
+=cut
+
+sub AutoResponseIDForQueueByType {
```
mgruner: I wonder if this could also be integrated into the AutoResponseList method as a filter (QueueID), since the return data is similar.

Kernel/System/AutoResponse.pm
```diff
@@ -385,12 +402,192 @@ Return example:
...
+            'ID' => '2',
+            'Type' => 'auto reject',
+            'Name' => 'default reject (after follow-up and rejected of a closed ticket)'
+          },
+       # ...
     );
+
+=cut
+
+sub GetAutoResponseData {
```
mgruner: Is this really needed or could the existing ```AutoResponseGet()``` be used as well?

s7design:
Hi @mgruner 

Thanks for all you suggestion, I tried to fix all of that.

* removed start and BeginWork() and Rollback() and there is used RestoreDatabase now. I have not known that before, thanks.
* removed AutoResponseListByType and updated AutoResponseList, with AutoResponseTypeID new optional parameter
* removed AutoResponseIDForQueueByType and updated AutoResponseGetByTypeQueueID. You suggested here that I update AutoResponseList. What do you think about my solution.

> I wonder if this could also be integrated into the AutoResponseList method as a filter (QueueID)...

You can see that I changed some other part of code. Working on this I saw that there is some mistakes/bugs. If you can I will be able to split that in more PRs.

* I removed $Param{AutoResponseOption} from here - https://github.com/OTRS/otrs/pull/807/files#diff-b4dce0d98a7d1b73a0ab6f2441fdc200L255. It left from this commit - https://github.com/OTRS/otrs/commit/77f170eedee9bbfb5c39b76e530b8c231ef1f8f7#diff-ffafbc8d220bf54720c22da7b977eaf6 
It is not used any more.
* I fixed shows header in for AdminAutoResponse change screens (Add/Edit) - https://github.com/OTRS/otrs/pull/807/files#diff-b4dce0d98a7d1b73a0ab6f2441fdc200R300 It has not worked before.
* removed UserID in AutoResponseList function - https://github.com/OTRS/otrs/pull/807/files#diff-b4dce0d98a7d1b73a0ab6f2441fdc200L346 It is not needed/used parameter.

I know that I have written a lot of things here, maybe it would have been easier for you that I had added inline comments. I hope you understand what I meant. 

Regards
Zoran

Kernel/System/AutoResponse.pm
```diff
@@ -356,19 +361,53 @@ Return example:
...
+    }
+
+    my %Data;
+
+    # if there is TypeID, select only AutoResponses by that AutoResponse type
+    if ( defined $Param{TypeID} ) {
+        $SQL .= " AND art.id = ? AND ar.type_id = art.id";
+
+        # select
+        return if !$DBObject->Prepare(
```
mgruner: I would prefer to have just one Prepare() call. You can have a @Bind array that you can push to if you need.

scripts/test/AutoResponse.t
```diff
@@ -31,161 +62,194 @@ $Self->True(
...
+    );
+
+    $Self->True(
+        $AutoResponseUpdate,
+        'AutoResponseUpdate()',
+    );
+
+    %AutoResponse = $AutoResponseObject->AutoResponseGet( ID => $AutoResponseID );
+
+    $Self->Is(
```
mgruner: You can optimize this code by using a loop with the field names (Subject, Name, ...).

mgruner:
Looks very nice in general!

s7design:
Hi @mgruner 

Thanks for all suggestion. I updated PR, I hope that code is better now. Feel free to tell me if I can change anything else.

Regards
Zoran

mgruner:
Merged in f38ca07. Thanks a lot! I believe the changes were worth it. I fixed some SQL issues in AutoResponseList, please see my following inline comments.

Kernel/System/AutoResponse.pm
```diff
@@ -356,19 +361,47 @@ Return example:
...
-        Valid => 0,
-        Clamp => 1,
-        Table => 'auto_response',
+    # get database object
+    my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
+
+    my $Valid = $Param{Valid} // 1;
+
+    # create sql
+    my $SQL = "SELECT ar.id, ar.name FROM auto_response ar, auto_response_type art";
```
mgruner: The second join is not required, because you can use ar.type_id directly. But if you join it, you always need the ar.type_id = art.it WHERE condition. This was not always present, only if TypeID was sent!

Kernel/System/AutoResponse.pm
```diff
@@ -356,19 +361,47 @@ Return example:
...
+    if ($Valid) {
+        $SQL
+            .= " WHERE ar.valid_id IN ( ${\(join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet())} )";
```
mgruner: WHERE was only added if Valid was true. It could be missing if Valid was 0 but TypeID was true.

mgruner: Please have a look into the merge commit to see how I fixed these issues and let me know if this works for you. I also added a UT case with Valid => 0.

s7design:
Hi @mgruner 

Thanks you Martin, your suggestions was very helpful. I appreciate this because I always learn something new:
* about RestoreDatabase with Helper
* about push param in @Bind and call Prepare only once
* and many other things

Thanks for all

I will have a look into the commit.

Regards
Zoran

PR #808, opened 2015-11-04
[Rel-5_0 Bug 7301] - Customer can reply to merged tickets, wrong ticket will be reopened

s7design:
http://bugs.otrs.org/show_bug.cgi?id=7301

Hi @mgruner ,

Hereby is proposed solution by Urlike in bug#11173 for solving this issue. Agree with his solution that reply button should be disabled for tickets that are in merged state. This way we can avoid having issues with customers replying on merged tickets and causing them to get reopened.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/CustomerTicketZoom.pm
+++ b/Kernel/Modules/CustomerTicketZoom.pm
@@ -1732,7 +1732,7 @@ sub _Mask {
         )
         && (
             ( $FollowUpPossible !~ /(new ticket|reject)/i && $State{TypeName} =~ /^close/i )
-            || $State{TypeName} !~ /^close/i
+            || $State{TypeName} !~ /^close|merged/i
         )
         )
     {


nilei:
If customer users reply by email to merged tickets, the merged ticket is not updated but the merge-target parent ticket. I feel this should be the same if the customer user replies in the customer interface.

It may also be that the customer user does not have access to the merge-target parent ticket and so the customer user would be banned from communicating back to the service agents. Removing the button is not a solution for me.

-- Cheers, Nils

mgruner:
@nilei, what would you propose instead (based on the current design)? We cannot just send the article to another ticket, this would be illogical in the GUI. Why are "merged" tickets shown at all, they do not show in the agent interface if I remember correctly? Should they not be hidden in the customer interface as well?

nilei:
@mgruner - Hi Martin, I agree with you, but I do not like the button to be just hidden. Do you think we may hide the button as proposed and display a message to the customer users to inform them that the ticket is "write protected" protected or such ?!

nilei:
@mgruner - The agent interface does not display the merged ticket in the linked obejcts area anymore, but the actual tickets can be searched, opened and being worked on.

mgruner:
This has been fixed in the meantime, see http://bugs.otrs.org/show_bug.cgi?id=7301. Thanks!

PR #812, opened 2015-11-05
Dutch stopwords

xhoy:
Added dutch stopwords


mgruner:
Hello @xhoy, very nice! I wonder though; there are quite a few longer words like Nederlands. Do you really think these have to be filtered?

xhoy:
Dunno, but there is a Highfrequencie list in webselex for dutch, german and english, i just picked the first 300 of that list, so these are the 300 most common words in dutch (i guess?)

xhoy:
Just some clarification, I thought that the stop words list would include words that would be EXCLUDED in the search results en thus clearing them for "the most common" words, like "the" and "is" for example.

My thought then was, these are words you would never search since they would only return crappy results. so I figured that the 300 "most common" words (but i have more if you like) would as well just return useless results since non of these words would be indexed at all and create a faster search engine.

mgruner:
Thanks. Can I ask you to please sign our CLA as described at https://otrs.github.io please? I believe that I don't have a copy from you yet. Then I will proceed to merge your PR.
Best regards, Martin

mgruner:
Merged, thanks! I reduced the list to about 150 words to be like the other lists.

PR #813, opened 2015-11-09
Modify translation files for Japanese.

sipadan2003:
Modified translation files for Japanese.


mgruner:
Hello @sipadan2003. Translation files are managed exclusively via Transifex, could you please join there? https://www.transifex.com/otrs/OTRS/
Regards, Martin

PR #815, opened 2015-11-10
Add GenericInterface TicketSend operation

fredericve:
This adds a "TicketSend" operation which allows you to create a ticket and then send it out to a CustomerUser by e-mail in the same way as it is done by the "new email ticket" action in the OTRS web interface.

It is basically a copy of the TicketCreate operation with some required changes.


nilei:
Why not enhancing the TicketCreate operation instead? A ticket created with article sender type Agent and article type email-external or email-internal should do the same. 

:-)

-- Cheers, Nils

fredericve:
Hello @nilei 

I had thought about that, but the TicketCreate operation uses the ArticleCreate call instead of the ArticleSend call, so in the current state it will never actually send the article by e-mail. On top of that, TicketCreate seems to require the "To" attribute to be the name of a queue.

I ended up implementing it like this due to time constraints and to have a clear distinction between TicketCreate and TicketSend in the same way there is ArticleCreate and ArticleSend.

If required, I may be able to find some time to integrate it in TicketCreate.

xhoy:
i a gree with nilei that it would be more obvious to use the same code but since it isn't the case at the moment it would be wise to not break backwards compatiblity and create a seperate function.  +1 for this commit!

fredericve:
Improved it a bit more. Now the used queue's systemaddress is used when no from address is added in the request. This makes the From address optional.

I still feel the Ticket and TicketSend operations are fundamentally different and personally prefer to keep them separated. For creating tickets and sending an outbound mail the checks are fundamentally different. E.g. for simply creating a ticket (_towards_ a queue) the From should be a customerUser, while for sending a ticket (_from_ a queue) The from should be the queue's address. In order not to introduce regressions I opted to keep them separated.

mgruner:
Looks very nice. One limitation is that this would not work for existing tickets like ArticleSend does in the API. Would it be useful to also have an ArticleSend operation or to combine this into one? @carlosfrodriguez what do you think, too?

xhoy:
super awesome idea! I would love this feature:)

nilei:
I still recommend to include both into TicketCreate and TicketUpdate. If the article type is calculated by mapping tables and logic, there is no way to use the same operation - so the logic must be present outside of the web service infrastructure, which I feel is counterproductive.

3limin4t0r:
when will it be released? really love to use this feature! :)

xhoy:
Could the backwards compatibility be fixed if there would be a parameter detection? or a boolean set to let the system send it? this would make everybody happy and mabe it would come along faster? :)

xhoy:
@fredericve I offer you €250 to complete the feature including the SendArticle /articlecreate part. 
I would suggest incorporate

xhoy:
Sorry for the weird comment, i was on my phone :)

@fredericve if you fix this PR before the end of the month (this includes the sendarticle/createarticle && create/send-ticket) ,ill send you €250 for fixing it.

I would suggest adding it to the current API and not introducing it separately!

fredericve:
@xhoy I'm dealing with some time constraints right now (both professionally and personally) so end of this month may be tight depending on the amount of work.

@nilei what would it take for you to accept this pull request? I suppose moving the logic to TicketCreate.pm and TicketCreate.pm  and use ArticleCreate and ArticleSend depending on the parameters should be sufficient? I'm trying to estimate the amount of work.

xhoy:
@fredericve give or take 2 weeks if fine, asap :smile: but this would seriously make our lives a lot easier so we are happy to support you :+1:

xhoy:
@fredericve how is it going?

fredericve:
@xhoy sorry, I haven't had any time yet. Honestly I don't foresee that I'll have some time in the next 3 to 4 months.

In any case you can already use the pull request as it is right now. We're using it as a patch in our production environment for a few months already.

xhoy:
Ah, thats nice, Is there an issue when somebody else would finish your PR?

fredericve:
Of course not. It's free and open source!

xhoy:
@carlosfrodriguez can we get a bounty tag or something? ill still over €250 for anybody who finishes this PR within the coming 30 day's

carlosfrodriguez:
Hi @xhoy,

For as funny as it could sound I think this will not happen :-(, but you should ask @mgruner.

((enjoy))
cr

mgruner:
@carlosfrodriguez I would first ask for your assessment of the PR please, as I already asked for a while ago. ;)

carlosfrodriguez:
Hello Together,

You had expend some time working on this and the solution looks complete for what is intended, and I appreciate everybody interest in this. However having a new operation to create a ticket and send the article might be too much, provably it should be better to update existing operations to use either ArticleCreate() or ArticleSend() depending on the parameters you add to the Article.

I think we can extend the Article attribute having this new parameters

```perl
        ArticleSend => 1,                                                      # optional
        To          => 'Some Customer A <customer-a@example.com>',             # optional
        Cc          => 'Some Customer B <customer-b@example.com>',             # optional
        ReplyTo     => 'Some Customer B <customer-b@example.com>',             # optional
        InReplyTo   => '<asdasdasd.12@example.com>',                           # optional
        References  => '<asdasdasd.1@example.com> <asdasdasd.12@example.com>', # optional
        Loop        => 0,                                                      # 1|0 used for bulk emails
        Sign => {                                                              # optional
            Type    => 'PGP',
            SubType => 'Inline|Detached',
            Key     => '81877F5E',
            Type    => 'SMIME',
            Key     => '3b630c80',
        },
        Crypt => {                                                             # optional
            Type    => 'PGP',
            SubType => 'Inline|Detached',
            Key     => '81877F5E',
            Type    => 'SMIME',
            Key     => '3b630c80',
        },
```

Then if ArticleSend is not present or its value is 0 we can use the code as it is right now (backwards compatibility), but if its set to 1 or any other value (except for 0) then we add some more validations for the new parameters to make sure if they are present they have the proper format.

For ArticleSend, 'To' parameters should be always defined.

This should be implemented in both TicketCreate and TicketUpdate operations then in this way you can create a new ticket and send the article (kind of AgentTicketEmail) or you can update an existing ticket and send the article (kind of AgentTicketCompose).

What do you think?

((enjoy))
cr

nilei:
:100: That would be awesome !

fredericve:
@carlosfrodriguez Of course I agree with all your comments. I'll have a quick look if I can implement it like that.

fredericve:
Hello @nilei and @carlosfrodriguez 

I implemented the changes inside the TicketCreate and TicketUpdate operations. Can you please review and comment?

xhoy:
Is it ready for merging and release????

fredericve:
@carlosfrodriguez can you still provide feedback?

xhoy:
@carlosfrodriguez @mgruner any news?

carlosfrodriguez:
Sorry about the delay you will get feedback soon, at first sight it does not look that bad at all :-) some unit test should be great (using the Mail/Test.pm backend)

((enjoy))
cr

carlosfrodriguez:
Just wondering... why not use the customer information if there is no "To" as a fall back, then if we don't have the email address we can fail with an error message.

((enjoy))
cr

nilei:
@carlosfrodriguez, using the customers email as TO would only be okay if the article is an external article, like email-external. Otherwise the TO most obvious SHALL NOT be the customers email! I am not sure if this is easy to implement and it should also not be too strict as the GUI also allows email-internal to customers. So I support the current implementation that TO is a mandatory parameter if ArticleSend is used.

carlosfrodriguez:
Hi @nilei yes well, I mean it depends on the sender type I was just thinking in the Agent side, but for sure if its the customer it should be the other way around (we have this kind of code in other parts of OTRS).

ArticleType should be restricted to only email related stuff if article send is set (i think).

((enjoy))
cr

carlosfrodriguez:
Hi @fredericve 

Please add the needed some unit tests.

Thanks

((enjoy))
cr

xhoy:
@fredericve lets move this PR along, i just patched our dev environment and it works like charm :) you kept hours, so i keep mine.
once merged ill transfer you the funds.

xhoy:
Ok, we started to use this patch in dev. We would it very handy that there is a separate "ArticleSend" property. For development we set this to 0 :)

flexd:
We would be very happy if this was merged! I tried looking at creating the missing unit tests myself, but I have never touched Perl and the files are gigantic :+1:

BlangTech:
I updated this PR to OTRS version 5.0.13, made some bugfixes and added the feature to send attachmentfiles via mail and not only attach them to the ticket.
The question is: How can i provide my changes? I dont have push permission fpr this PR :-(

jepf:
We made a public extension for this:
https://github.com/znuny/Znuny4OTRS-GIArticleSend

It's based on this PR. For details see our PR #1539
Thanks to @fredericve and our customer PC-College.

fredericve:
@jepf cool! Thanks!

NeverMin:
@jepf 👍



2016-10-27 16:56 GMT+08:00 Jens Pfeifer <notifications@github.com>:

flexd:
@jepf Does this support sending PGP-encrypted emails?

flexd:
I see `ArticleSend` supports 
```
 EmailSecurity => {
            Backend     => 'PGP',                       # PGP or SMIME
            Method      => 'Detached',                  # Optional Detached or Inline (defaults to Detached)
            SignKey     => '81877F5E',                  # Optional
            EncryptKeys => [ '81877F5E', '3b630c80' ],  # Optional
        }
```

as input. If I could specify that in the API call to `GIArticleSend`, that would be very helpful! @jepf :)

thorsteneckel:
HI @flexd - thanks for your feedback. Could you please create an issue or even a pull request at our repository @ https://github.com/znuny/Znuny4OTRS-GIArticleSend/issues

We will then sync it back to our pull request.

flexd:
@thorsteneckel Thanks for the reply. I have just created an issue :)

flexd:
@thorsteneckel Thanks for the reply. I have just created an issue :)

mgruner:
Superseded by #1539, closing.

PR #816, opened 2015-11-12
Optimize GenericAgent management

moritz:
The old way was to first query the last execution times, and then deletes them individually.
When several GenericAgent jobs run in parallel (for example in a cluster, and/or when several
E-Mails are delivered into the PostMaster in parallel), this produces a quadratic number
of DELETE statements (which caused an outages for us).

Instead, simply delete all old run time values in a single delete statement, which
keeps the quadratic run time out of OTRS (and confined to the database, which
handles it much better). As a bonus, the code is also shorter and simpler :-).

### DIFF
--- a/Kernel/System/GenericAgent.pm
+++ b/Kernel/System/GenericAgent.pm
@@ -1433,21 +1433,11 @@ sub _JobUpdateRunTime {
     # get database object
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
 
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @moritz,

Thank you for your pull request, it was merged in https://github.com/OTRS/otrs/commit/7f36203a829dca82700b7dbfd5091134b79970e9.

((enjoy))
cr

PR #817, opened 2015-11-12
[Rel-5_0 Bug 11286] - Default-Value "X-Envelope-To" instead of "Envelope-To" in PostmasterX-Header

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11286

Hello @mgruner ,

As reporter suggested email tag X-Envelope-To is used as well as Envelope-To, really depends on mail servers that users use. Instead of replacing it, added additional tag in SysConfig that users can choose for there needs.

Heres description taken from one of our sources used in researching this issue.
http://people.dsv.su.se/~jpalme/ietf/mail-headers/mail-headers.html

Description | Email Tag |  Note
------------ | -------------  | -------------
If the recipient in the envelope (SMTP "MAIL FROM") is not included in the CC list, some mail servers add this to the RFC 2822 header field as an aid to clients which would otherwise not be able to display the envelope recipients. | X-Envelope-To:, Envelope-To: | Non-standard.

Updated PostMaster and EmailParse unit test to include these tags now in testing.

If you think we should just keep X-Envelope-To tag (which is more used, when used then Envelope-To), i will update this PR accordingly. Please, let me know what do you think about this suggestion and PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -9737,6 +9737,7 @@
                 <Item>X-Original-To</Item>
                 <Item>Delivered-To</Item>
                 <Item>Envelope-To</Item>
+                <Item>X-Envelope-To</Item>
                 <Item>Return-Path</Item>
                 <Item>X-OTRS-Owner</Item>
                 <Item>X-OTRS-OwnerID</Item>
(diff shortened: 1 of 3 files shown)


UdoBretz:
Hello Sanjin,

thanks for your pull request! It is merged now for OTRS 5 and master.

Thanks, and best regards,
Udo

PR #818, opened 2015-11-12
[ Master Bug 10825 ] - <OTRS_CUSTOMER_Body> in Reply-Template

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10825

Hi @mgruner 

As you can see on the bug report, reporter wanted to use notification in Reply-Template. It is not supported by now.  In my opinion there is a bug now because OTRS_CUSTOMER_BODY is replaced with HTML tags.  

Here is our proposal for solving it in this moment, we add a param in _Replace function to check if it is allowed replacing OTRS_AGENT_* and CUSTOMER_AGENT_* tags. It is the same for Signature and Salutation. I know that it is not the best solution, but I am not sure if you want replace all parameters for all modules.  In that case there should be more improvement in many functions ( e.g. Template(), Signature(), Salutation() ... in TemplateGenerator). In any case we shoul fix something because the current situation with replacing wrong date as it is mentioned in the bug report is not option.

Regards
Zoran

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -135,11 +135,12 @@ sub Salutation {
 
     # replace place holder stuff
     my $SalutationText = $Self->_Replace(
-        RichText => $Self->{RichText},
-        Text     => $Salutation{Text},
-        TicketID => $Param{TicketID},
-        Data     => $Param{Data},
-        UserID   => $Param{UserID},
+        RichText                    => $Self->{RichText},
+        Text                        => $Salutation{Text},
+        TicketID                    => $Param{TicketID},
+        Data                        => $Param{Data},
+        ReplaceAgentAndCustomerTags => 0,
+        UserID                      => $Param{UserID},
     );
 
     # add urls
@@ -246,12 +247,13 @@ sub Signature {
 
     # replace place holder stuff
     my $SignatureText = $Self->_Replace(
-        RichText => $Self->{RichText},
-        Text     => $Signature{Text},
-        TicketID => $Param{TicketID} || '',
-        Data     => $Param{Data},
-        QueueID  => $Param{QueueID},
-        UserID   => $Param{UserID},
+        RichText                    => $Self->{RichText},
+        Text                        => $Signature{Text},
+        TicketID                    => $Param{TicketID} || '',
+        Data                        => $Param{Data},
+        QueueID                     => $Param{QueueID},
+        ReplaceAgentAndCustomerTags => 0,
+        UserID                      => $Param{UserID},
     );
 
     # add urls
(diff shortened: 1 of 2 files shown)


mgruner:
@carlosgarciac please carefully check this.

carlosgarciac:
Hi @s7design,

I just review this issue, next my comments:

1.- I agree with you this is not a bug, it could be a missing feature, due smart-tag '<OTRS_CUSTOMER_BODY>' is not supported for now.
2.- I like the idea about 'ReplaceAgentAndCustomerTags', but IMO could be better to remove the un-supported tags in the template, before sent it to the '_Replace' function, we can use OTRS_CUSTOMER_DATA_ tags.
3.- OTRS_CUSTOMER_EMAIL and OTRS_COMMENT are also replaced with the Body sent to the _Replace function.
4.- In AgentTicketCompose the body send to the '$TemplateGenerator->Template' is not the correct one, it should be the body of the article without quoting then we could use the OTRS_COMMENT tag there, for the other calls to the template generator the $Data{Body} param is correct, IMO.
5.- On AdminTemplate->Example Template, the reference to 'Your email address is: "<OTRS_CUSTOMER_UserEmail>"' should not be valid.

Best regards,
Carlos García

s7design:
Hi @carlosgarciac 

Thanks for your detailed follow-up. It help us to make, I hope, better proposal for fixing this bug.
You can see that I changed logic for avoiding to replace not supported tags.

At the beginning of the _Replace function there is if statement which check if there is needed replacing Agent and Customer tags.

<pre>

+    if ( !$Param{ReplaceAgentAndCustomerTags} ) {
+
+        my @ListOfNonSupportedTag = qw/OTRS_AGENT_SUBJECT OTRS_AGENT_BODY OTRS_CUSTOMER_BODY OTRS_CUSTOMER_SUBJECT/;
+
+        # cleanup all not supported OTRS_CUSTOMER_ and OTRS_AGENT_ tags
+        my $NotSupportedTag = $Start . "(?:" . join( "|", @ListOfNonSupportedTag ) . ")" . $End;
+        $Param{Text} =~ s/$NotSupportedTag/-/gi;
+    }
</pre>

My idea was that we make some generic solution. There is @ListOfNonSupportedTag, which is static initialised now, but it could be extended in future with some SysConfig item that represent not supported tags in some modules or now if we would like to add some non supported tags it is easy for extanding. 
Not supported tags is replaced by '-'. I used the same logic as it is at the end of _Replace function.

<pre>
    # cleanup all not needed OTRS_CUSTOMER_DATA_ tags
    $Param{Text} =~ s/(?:$Tag|$Tag2).+?$End/-/gi;

    # cleanup all not needed OTRS_AGENT_ tags
    $Tag = $Start . 'OTRS_AGENT_';
    $Param{Text} =~ s/$Tag.+?$End/-/gi;
</pre>

If you prefer solution that we remove unsupported tags before _Replace, I will be able to update this PR and add such code for each template where it is necessary.  

Please let me know what do you think about that solution. 

Regards
Zoran

carlosgarciac:
Hi @s7design,

Thanks for your nice proposal.

IMO I prefer to merge your solution and mine in one:

1.- Create a new function, probably call  'RemoveNonSupportedTag' 
2.- There you can send the 'Text' and the 'ListOfNonSupportedTag' as params
3.- On this new function you can cleanup all not supported tags
4.- This function could be used on the places it's needed, probably with different values for un-supported tags, it could give us more flexibility for restricting the use of this smart tags.
5.- Just as reminder, don't forget UnitTests for this new function

Please let me know what do you think about it.


Best regards,
Carlos García

s7design:
Hi @carlosgarciac 

Thanks for suggestion. I will try to implement as you suggested. In this moment it looks fine. I hope it will works nice for this feature.

Regards
Zoran

s7design:
Hi @carlosgarciac 

I am working on this and I think that it is better that I solve it in new PR. I tried to rebase but there are so many conflict because of the previous two commits on this PR. Because of that I close this PR and I will make another one with solution that you suggested.

Reagrds
Zoran

PR #819, opened 2015-11-13
Master - fixed the list of agent and customer users for invalid entities

s7design:
Hi @mgruner 

Working on some bug with LDAP I saw that we forgotten fix format for invalid entities if there is not column Valid for users (agent and customer) 

In the following commit was fixed only for Customer Company 
https://github.com/OTRS/otrs/commit/70eb6efb8c164b613a18e43d24c8658ba90d6a3f

I thought it could be merged in master as well.

Regards
Zoran


mgruner:
Merged and registered as http://bugs.otrs.org/show_bug.cgi?id=11651. Thanks!

PR #820, opened 2015-11-13
Master - fixed clean up for customer user by Helper RestoreDatabase

s7design:
Hi @mgruner 

I tested a little bit Helper with RestoreDatabase and I saw that maybe there is a problem with setting CustomerUser to invalid if there is used RestoreDatabase.

There is PR with my fix. I updated Replace unit test as a  example for this fix.

It is my proposal, maybe I am wrong somewhere, please let me know what do you think about that.

Regards
Zoran


mgruner:
Merged in 8da9227fe674c03a19a675538e6922cc6000df2c, thanks. I added a comment at the end of Replace.t explaining how the cleanup happens.

PR #821, opened 2015-11-13
[ Master Bug 8055 ] - Disable Add/Edit customers when usign LDAP

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8055

Hi @mgruner 

I tried to fix that. This is my proposal.There is supported only add/edit in DB backends with ReadOnly config param for apropriate Source.

Please let me know what do you mean about that. I am planning to add Selenium test for this as well and I have some ideas how we can improve these two screens.

Regrards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
@@ -58,9 +58,6 @@
             </div>
 [% RenderBlockEnd("ActionList") %]
         </div>
-
-
-
         <div class="ContentColumn">
             <div class="WidgetSimple">
 [% RenderBlockStart("OverviewHeader") %]
@@ -134,6 +131,9 @@ $('.MasterAction').bind('click', function (Event) {
 [% RenderBlockStart("HeaderEdit") %]
                     <h2>[% Translate("Edit Customer") | html %]</h2>
 [% RenderBlockEnd("HeaderEdit") %]
+                    [% IF Config(Data.Source).ReadOnly == 1 %]
+                    <h2><span class="Warning">[% Translate("This customer backend is read only!") | html %]</span></h2>
+                    [% END %]
                 </div>
                 <div class="Content">
                     <form action="[% Env("CGIHandle") %]"  class="Validate" method="post">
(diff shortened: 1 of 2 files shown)


Kernel/Output/HTML/Templates/Standard/AdminCustomerCompany.tt
```diff
@@ -129,75 +129,81 @@ $('.MasterAction').bind('click', function (Event) {
 [% RenderBlockStart("OverviewUpdate") %]
                 <div class="Header">
 [% RenderBlockStart("HeaderAdd") %]
-                    <h2>[% Translate("Add Customer") | html %]</h2>
+                    <h2>[% Translate("Add Customer") | html %]
 [% RenderBlockEnd("HeaderAdd") %]
 [% RenderBlockStart("HeaderEdit") %]
-                    <h2>[% Translate("Edit Customer") | html %]</h2>
+                        <h2>[% Translate("Edit Customer") | html %]
```
mgruner: The indentation looks wrong here. Do you have a current OTRSCodePolicy module? I removed the auto-formatting for TT files there.

s7design: Thanks Martin, I will update this PR. It looks strange for me yesterday when CodePolicy tidied on that way my changes in tt files.

Regards
Zoran

s7design:
Hi @mgruner
I updated this PR. I current  OTRSCodePolicy. I think that there is some bug. I wanted to add text "This customer backend is read only!" in one h2 tag. In that case there was a problem with indentation after tidying by OTRSCodePolicy.
Please let me know what do you think if it is ok current solution, where there are two h2 tags. There is many ways how we can solve this, but in this moment there is not the main problem indentation. If you like solution that we disable submit button and add this text I will be able to find better solution to solve this issue with indentation.
Regards
Zoran

mrcbnsls:
Thanks, merged in e9677ff with some adaptions. I basically introduced a new message style especially for widgets. Please have a look at the CSS I've added if you like :-)

s7design:
Hi @mrcbnsls ,

I think that new style for message boxes is very nice. Good job :+1: 

Regards
Zoran

nilei:
Can you also disable the input fields if the data source sis ReadOnly = 1 ?

mrcbnsls:
@nilei, thanks, nice idea, added in ae49ae8e572a4181555b422c86d6f6c487502120. I also improved the check for the source module to be more flexible.

nilei:
@mrcbnsls, thanks for considering my idea, but I think you disabled too much :-) When I now access a user account of a read only source, also the preferences are disabled.
:-)
-- Cheers, Nils

mrcbnsls:
@nilei, could you explain a bit more what you're referring to? If the data source is readonly, it is not possible at all to update a customer user from this source in AdminCustomerUser. Am I getting something wrong?

nilei:
@mrcbnsls, the customer user preferences can be submitted even on read only sources because they are stored in OTRS' own customer_preferences table.

dominikklein:
@nilei, now the customer user preferences can also changed on ready only sources 609f4e7a747d1e7c63faabe58af746f9bf375d40.

nilei:
:+1:

PR #823, opened 2015-11-16
[Rel-5_0 Bug 11605] - New ticket notification names are not translated in agent preferences

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11605

Hi @mgruner ,

In our opinion, as mentioned in bug report, we fear this is unnecessary addition when you can change names of the notification. Having this in mind, we would like to propose a solution that could solve reporter issue.

What we could offer is translatable names of initially added notifications in the system on install. This way, if there are translated strings, user can get notification name on preferred language, as long as it remain genuine. This has been done with other types of data thats created on initial install ( e.g. auto-response, salutation, signature; will create different PR that will also Translate(Data.Name) for this data, if we choose this way to proceed) 

Perhaps we can discuss more on this topic where your input is most welcome. Looking forward to your reply.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminNotificationEvent.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminNotificationEvent.tt
@@ -102,7 +102,7 @@
 [% RenderBlockEnd("NoDataFoundMsg") %]
 [% RenderBlockStart("OverviewResultRow") %]
                         <tr [% IF Data.ValidID != 1%]class="Invalid"[% END %]>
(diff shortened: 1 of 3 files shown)


mrcbnsls:
I think it's fine to add translations for the initial notification names, although they can be changed.

But I think there is at least one more place where translation needs to be added in the code: PreferencesNotificationEvent.tt. Could you please change this, too?

s7design:
Hi @mrcbnsls ,

Updated PR to translate notification names in AgentPreference screen. 
Please let me know if there is anything else we can do.

Regards,
Sanjin

mrcbnsls:
Thanks for the update, merged in ada0988.

PR #824, opened 2015-11-16
Add `--sql-part` option to `Dev::Tools:Database::XMLExecute` console cmd

eliasp:
This allows to only execute the 'pre' or 'post' part of an XML SQL
definition, as those often can't be executed at once due to foreign key
restrictions.
This is similar to the 'split-files' feature of the
`Dev::Tools::Database::XMLExecute` console command.


mgruner:
Merged in 68a1a6a7ba4daea76c031a0d53c7fb353c4a6ea2, thanks a lot @eliasp! :+1:

PR #825, opened 2015-11-17
[Rel-5_0 Bug 11657] - Since upgrade Ticket::Frontend::AgentTicket*###Subject is empty

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11657

Hi @mgruner ,

There were missing default values in SysConfig for Subject fields. Added them in same fashion as it was used in 4.0.

Please let me know if you disagree with this and if these default values were left empty for a reason.

Regards,
Sanjin


mgruner:
Hello Sanjin,
thanks for the proposal, but they were left out on purpose to avoid having redundant subjects. You should either fill them in or make the article not mandatory. And you can still provide a different config setting if you like. Can you close the bug please?
Regards, Martin

PR #826, opened 2015-11-18
Fix bug#11663 - View type not saved to UserPreferenses in AgentTicketZoom

noritnk:
This PR is fix for bug#11663.
See http://bugs.otrs.org/show_bug.cgi?id=11663 .


mrcbnsls:
@noritnk, this doesn't work correctly for me. It only works for expanded/collapsed view, but as soon as the timeline view (which is part of the OTRS Business Solution™) is used, this is obviously not saved to the prefs.

noritnk:
Hi @mrcbnsls,
Thanks for reviewing.  You are right.
I fixed the bug that you pointed.

mrcbnsls:
Thanks for the update. It was still not working correctly under some circumstances and I found out why. It is possible that AgentTicketZoom gets called multiple times by the same action. If you open a ticket with ZoomExpand=0 (= single article view), the viewed article gets marked as seen (Subaction=MarkAsSeen) which will cause the prefs to be updated again in the same call, but with the wrong value. I now added some code to make sure the prefs only get updated on the main call.

PR #827, opened 2015-11-18
[Rel-5_0 Bug 10931] - Reply to note does not use the origin note subject

s7design:
http://bugs.otrs.org/show_bug.cgi?id=r

Hi @mgruner ,

Hereby is proposal for fixing this issue. Removed in IF condition to check for $Config->{Subject}. This way 'Reply to note' can take subject from original note it replies to, nevertheless if it's default value is set in sys config or not ( since default value was not used when creating subject for 'Reply to note' )

Updated Selenium test to check for subject value in 'Reply to note' and generally improved test.

If there are any issues regarding this PR, please let me know.

Regards,
Sanjin


zottto:
Merged with 12fbf14. Thanks!

PR #829, opened 2015-11-19
[Rel-5_0 Bug 11654] - Internal server error with limit 0 in out-of-office-dashboard

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11654

Hi @mgruner , 

It was possible to send param PageShown value 0 to Layout/PageNavBar function resulting in fatal error. With this small fix, we bypass it and use default value when 0 is set as limit in SysConfig.

Regards,
Sanjin


mrcbnsls:
Thanks, merged in b638a8c.

PR #831, opened 2015-11-20
Master - removed AAAcalendar.tt dummy template

s7design:
Hi @mgruner 

As I said in the mail, I have made PR where I removed obsolete dummy template. It was the easiest one for the begging.

This is my workflow:
* There are translated words for Calendar in Frramework.xml, and as you suggested I removed whole file.
* called command perl bin/otrs.Console.pl Dev::Tools::TranslationsUpdate ( without --generate-po)

Is it ok, should we update .po files as well? (I suppose not)
Please let me know if we should do something else. This one was easy, we should have more checks in the others AAA files.

Regards
Zoran


mgruner:
Hi Zoran,
this looks good. No changes in Kernel/Language or i18n/ please. This is only for you to verify that the required strings are still there (and not lost). But AAA* is not really the required strings. You need to check if the strings occur anywhere in the codebase, if not it can be dropped, otherwise you need to make sure it is marked as translatable (preferrably in all occurrences). Maybe you can automate this (the detection part).
Regards, mg

s7design:
Hi @mgruner 

I updated this PR, I believe that it is good now. We will continue to do on this task next days.

Regards
Zoran

mgruner:
Merged, thank you @s7design!

PR #832, opened 2015-11-23
[Rel-5_0 Bug 11231] - Handiling of web site urls in customer data

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11231

Hi @mgruner ,

Although we think this isn't a bug, we feel that there isn't explanation on how to use this field properly which can lead to potential issues as bug reported experienced. To prevent this and other future problems, regarding this topic, it would be nice to have field explanation for URL field in AdminCustomerCompany screen with example how to input field right.

Please if you think this is indeed bug and that inputing 'www.google.com' in field without adding protocol is expected to work, tell me, and i will work on solving it.

Regards,
Sanjin


mrcbnsls:
Won't merge for now. I think there was a misunderstanding on the bug report. Thanks anyway.

PR #833, opened 2015-11-23
Master - fixed dropdown for CustomerTicketOverview

s7design:
Hi @mgruner 

Working on some bugs we saw that there is wrong display in dropdown for CustomerTicketOverviewSortable SysConfig item in case there is set "No".

screen shot 2015-11-23 at 4 15 01 pm

I suppose the problem occurs after adding  modernize input fields. This config was added for bug 
http://bugs.otrs.org/show_bug.cgi?id=5987 in commit https://github.com/OTRS/otrs/commit/a1fb55f776a6852b5a23b81e05166b316930f0a7

I fixed Customer ticket search screen as well, in my opinion there was bug. There was added Sortable twice if CustomerTicketOverviewSortable was "Yes".

Please let me know what do you think about this proposal

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -1918,8 +1918,8 @@
         <SubGroup>Frontend::Customer::TicketOverview</SubGroup>
         <Setting>
             <Option SelectedID="">
-                <Item Key="" Translatable="1">No</Item>
-                <Item Key="Sortable" Translatable="1">Yes</Item>
+                <Item Key="0" Translatable="1">No</Item>
+                <Item Key="1" Translatable="1">Yes</Item>
             </Option>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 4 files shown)


mgruner:
This would not be backwards compatible, I guess?

mgruner:
Maybe you can just add '0' for the 'No' value and leave the other one as it is. The template can likely stay. Please test the upgrading case.

s7design:
Thanks @mgruner  it will be updated 

Regards
Zoran

s7design:
Hi,

I updated the PR. I added test case where is checked this sysconfig item. Maybe there will be needed your help with messages which I have used in the test.

Regards
Zoran

mrcbnsls:
Thanks, merged in 179fa71b3fc19c889f04c166aec6d41a7bb65e98.

PR #834, opened 2015-11-24
Master - removed AAAGenericInterface.tt dummy template

s7design:
Hi @mgruner,

I removed whole  AAAGenericInterface.tt dummy template. There were:
<pre>
- [% Translate("Webservice \"%s\" created!") | html %]
- [% Translate("Webservice \"%s\" updated!") | html %]
</pre>

However, they were not used anywhere, actually there are used:
<pre>
- [% Translate("Web service \"%s\" created!") | html %]
- [% Translate("Web service \"%s\" updated!") | html %]
and in AdminGenericInterfaceWebservice.pm there is also
- [% Translate("Web service \"%s\" deleted!") | html %]
</pre>

You know that there are used  "Web service" and "Webservice" in the framework. It is not big problem, but maybe one day we could fix that and change that there is used only one in entire system.

Regards
Zoran


mgruner:
Merged in 64fdadd, thanks! "web service" is the correct English term, you can correct all other visible instances that you stumble upon.

PR #835, opened 2015-11-24
Master - changed Language::Get function to Language::Translate

s7design:
Hi @mgruner 

Working on dummy translate template files I saw that there were in some places Get instead Translate function from Language module.  I fixed this here. If you want you can merge it.

regards
Zoran


mgruner:
Excellent, thank you! I also dropped ```Language::Get()``` in 3a1907f.

PR #836, opened 2015-11-24
Master - removed AAA*.tt dummy - calendar words

s7design:
Hi @mgruner 

There is removed dummy files for calendar words. I have a question here. 
For example in the Framework.xml there are for CalendarWeekDayStart
<pre>
                Item Key="0" Translatable="1">Sunday
                Item Key="1" Translatable="1">Monday
                Item Key="2" Translatable="1">Tuesday
                Item Key="3" Translatable="1">Wednesday
                Item Key="4" Translatable="1">Thursday
                Item Key="5" Translatable="1">Friday
                Item Key="6" Translatable="1">Saturday
</pre>

and again with Translatable for CalendarWeekDayStart::Calendar1 ... and for other calendar as well.
The same are with the names of months. I would like to ask here that I can know for further works. Is it OK if I delete "Translatable" for all words if there has been added before for another SysConfig item.

There are pros and cons if we add for each words in every config item and versa. For example if we one day delete some item with Translatable param, in that case we should think of adding Translatable param to another item. Of course if we add for each item, there is redundant data. I prefer to add only once. There is such case for TimeVacationDays.

Regards
Zoran


mgruner:
Merged in 5371f5c. Thanks!

Strings that are translatable must be marked so in all occurrences for logical reasons. Otherwise if you'd delete the first occurrence (that would be the only one marked) the string would be lost. That is why `yes`and `no` are marked hundreds of times, for example.

PR #838, opened 2015-11-25
Give otrs.Console.pl a --quiet option

moritz:
useful for cron jobs.

### DIFF
--- a/Kernel/System/Console/BaseCommand.pm
+++ b/Kernel/System/Console/BaseCommand.pm
@@ -83,6 +83,10 @@ sub new {
             Description => 'Do not perform ANSI terminal output coloring.',
         },
         {
+            Name        => 'quiet',
+            Description => 'Suppress informative output, only retain error messages.',
+        },
+        {
             Name        => 'allow-root',
             Description => 'Allow root user to execute the command. This might damage your system; use at your own risk.',
             Invisible   => 1,   # hide from usage screen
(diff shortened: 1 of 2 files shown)


mgruner:
Hi @moritz. So you tried the console API, I hope you like it? Regarding this PR, looks very nice, but you are probably aware of the fact that it will not work in all cases, for example in Admin::Package::* where there are messages printed by internal objects?

moritz:
Yes, it's nice.

I wasn't aware of the Admin:: packages printing things from internal objects, but I'm mostly concerned with Maint:: scripts that might be run from cron jobs.

Usuall cron jobs with non-empty output send emails, so they should only print somthing in the case of errors.

If I run into command that I want to run from cron, and where --quiet doesn't work, I'll simply send more pull requests to quiet them too :-)

mgruner:
It might be possible to (additionally) capture STDOUT around the ->Execute() call, what do you think? We use that a lot in the unit tests already.
Besides that, I'd appreciate unit test coverage for this new feature, please. We could also backport to OTRS 5 I guess.

carlosfrodriguez:
Just an small comment if you don't want to see the STDOUT in the daemon you can opt to enable capturing into a file with a SysConfig setting that we have specially for that. the drawback is that all output is logged into a file. and even if we also have a setting to remove this files based on its age it might not be optimal to log something that you just don't want to.

((enjoy))
cr

moritz:
@mgruner I've added a simple test for --quiet

mgruner:
Merged and backported to OTRS 5, thanks so much @moritz!

PR #839, opened 2015-11-25
[Rel-5_0 Bug 11096] - action URL overwrites permission settings

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11096

Hello @mgruner ,

Hereby is our proposal for fixing reporters issue. This could be solved in CustomerTicketOverview module as well, but with this proposal hopefully we can prevent other future potential similar issues with customer modules who branch into different subactions.

Added part where we go through each nav bar of customer module registration and searching if there are any subaction present in links. If some are found, matching it with current param subaction to check if that subaction module is accessible for current customer group.

Additionally there is Ticket::Frontend::CustomerDisableCompanyTicketAccess sysconfig that prevents customer to see company tickets that are not created by him. With this addition we can further limit it for certain groups and prevent 'force entry' by URL for desired subaction.

There is Selenium test to simulate this type of behaviour. Please let me know if there are any questions or issues regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/Web/InterfaceCustomer.pm
+++ b/Kernel/System/Web/InterfaceCustomer.pm
@@ -12,6 +12,7 @@ use strict;
 use warnings;
 
 use Kernel::System::Email;
+use Kernel::System::VariableCheck qw(IsArrayRefWithData);
 
 our @ObjectDependencies = (
     'Kernel::Config',
@@ -1071,7 +1072,7 @@ sub Run {
             return;
         }
 
-        # module permisson check
+        # module permisson check for action
         if ( !$ModuleReg->{GroupRo} && !$ModuleReg->{Group} ) {
             $Param{AccessRo} = 1;
             $Param{AccessRw} = 1;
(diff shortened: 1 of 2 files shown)


carlosgarciac:
HI @s7design,

I could see the code for 'module permisson check for action' is pretty similar, do you think is possible to reduce it?

Regards,
Carlos García

s7design:
Hi @carlosgarciac,

Yes, you are right. There is code for checking permission for module action. It is very similar to this one which we add for checking submenu action ( or in that case subaction)
I was trying to find solution that we omit duplicate code in first commit, but in that moment I didn't know how we could do this. I updated PR, where I added internal function for checking module permission. What do you think about this solution. I would like better solution, but there are many conditions which should be checked. 

The first IF statement (module permission check for action)  is necessary for checking module permission for current action. In case the module is restricted it is not needed contionue checking for subacton, because of that I add another IF statement (module permission check for submenu item) which is used only in case the current action module is allowed. Then it follows checking for subaction again only in case there are NavBar data for Module registration SysConfig item and there Group or GroupRo param. 

This is our proposal, please let me know if you have another suggestion.

Regards
Zoran

carlosgarciac:
Hi @s7design,

Thanks a lot for your nice PR, it was merged on de98219.

Best regards,
Carlos García

PR #840, opened 2015-11-25
Master - changed Language::Get to Language::Translate

s7design:
Hi @mgruner 

Working on the task for dummy templates, I saw that we forgot to change all calls of function Language::Get to Language::Translate when we had removed Get from Language module. I believe that I managed to find all of them now.

Regards
Zoran


mgruner:
Merged, thank you very much @s7design!

PR #841, opened 2015-11-25
[Rel-5_0 Bug 9061] - Initial queues are set to "Ticket lock after follow-up: yes"

s7design:
http://bugs.otrs.org/show_bug.cgi?id=9061

Hi @mgruner ,

Although we think that this is not really a bug since you can change this parameter to suit your needs when creating queue, it's nice to see consistency through system. Changed default value for 'Ticket lock after a follow up' to 'Yes' on initial load of AdminQueue;Subaction=Add screen.

Regards,
Sanjin


mgruner:
Hello Sanjin,
thanks a lot, merged with changes in 724630e. There was a bug in your solution: it must be `// 1` instead of `|| 1`, otherwise you cannot set to 'No' any more. :smile: I also extended the AdminQueue.t selenium test for this (recommended).
Best regards, Martin

PR #842, opened 2015-11-25
Bug 11688 - otrs.Console.pl Maint::Session::ListExpired doesn't list - it deletes! 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11688

Hi @mgruner,

As you can see there was a wrong message in the command, but behavior is good.

P.S. I have a suggestion. I have checked test for Session console command. There is only one test ( https://github.com/OTRS/otrs/blob/master/scripts/test/Console/Command/Maint/Sessions/Commands.t ) 
We can split it and make separate unit tests for each Session console command ( DeleteAll, DeleteExpired, ListAll and ListExpired)
Please let me know what do you think about that. We can do that in separate PR.

Regards
Zoran


mgruner:
Applied with slightly modified wording, thanks @s7design.

PR #844, opened 2015-11-26
Master - formated login messages

s7design:
Hi,

Working on some bugs for lost password in customer interface we saw that there are different format of messages in agent and customer interface.
We played a little bit with this. I believe it could be done better, and Marc's help will be welcomed.
If you like you can merge it.

Regards
Zoran


mrcbnsls:
Looks nice! Merged in d7dad54, thanks a lot.

PR #846, opened 2015-11-27
Simplify AgentTicketWatchView

moritz:
by using the NotTicketFlag search filter


mgruner:
Sorry Moritz, I had to revert this change again as there seems to be a bug:

bildschirmfoto 2015-11-27 um 13 20 10

Can you please review this and open a new PR?

PR #848, opened 2015-11-27
Master - follow up bug 11168

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11168
http://bugs.otrs.org/show_bug.cgi?id=10853

Hi @mgruner,

I tested behavior for the bug #10853 after fixing in the bug #11168. Actually there is a bug in fix:
There is a problem
https://github.com/OTRS/otrs/blob/master/Kernel/Modules/AgentTicketSearch.pm#L347

There are the same code in the both if  statement, for 'Open' and Closed 'StateType', and the second mistake is $Param{StateType} eq 'Closed' . It could be $GetParam{StateType} eq 'Closed'.

Fix that I have made for the bug #10853 is necessary bu only partially. There is a problem with the count numbers in customer information links in AgentTicketZoom (Open, Close)

Here are fix for this problem and updated tests as well. 

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketSearch.pm
+++ b/Kernel/Modules/AgentTicketSearch.pm
@@ -305,8 +305,9 @@ sub Run {
         $GetParam{ResultForm} = '';
     }
 
-    # get user object
-    my $UserObject = $Kernel::OM->Get('Kernel::System::User');
+    # get needed objects
+    my $UserObject  = $Kernel::OM->Get('Kernel::System::User');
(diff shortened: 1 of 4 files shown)


s7design:
New bug opened. It could be fixed with this PR as well 
http://bugs.otrs.org/show_bug.cgi?id=11786

Regards
Zoran

dvuckovic:
Merged with a few test tweaks. Thanks to S7 (Zoran) for the fix!

PR #850, opened 2015-11-30
[ Master Bug 9038 ] - Default Ticket Type in Emails

s7design:
Hi @mgruner,

There is our proposal for fixing this bug. Actually, we did not fix the bug, we added not for invalid ticket types. We used ideas as this is fixed with invalid customer company for ticket information widget.
Maybe, we can change  style for customer note.

The problem is because ticket type is hardcore if it is not defined.
https://github.com/OTRS/otrs/blob/master/Kernel/System/Ticket.pm#L306

Generally, there could be such problems for other entities which is in meantime set to invalid. (e.g for Queue:Raw, for State:open ... and other entities which are often set as default )

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -143,6 +143,14 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::Type::Default" Required="1" Valid="1">
+        <Description Translatable="1">Defines the default ticket type.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Core::Ticket</SubGroup>
+        <Setting>
+            <String Regex="" Translatable="1">Unclassified</String>
+        </Setting>
+    </ConfigItem>
     <ConfigItem Name="Ticket::Service" Required="1" Valid="1">
         <Description Translatable="1">Allows defining services and SLAs for tickets (e. g. email, desktop, network, ...), and escalation attributes for SLAs (if ticket service/SLA feature is enabled).</Description>
         <Group>Ticket</Group>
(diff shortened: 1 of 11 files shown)


zottto:
Hi @s7design ,

I recommend to not fix it this way. The best solution would be a new sysconfig option "Default TicketType" and to assign that to the tickets (and not TypeID 1).
Additionally in the admin interface for the TicketType, if someone wants to set a type to invalid, you can check if this is the default type and not allow the change until the default type is changed, too.

We can keep the note in AgentTicketZoom additionally to inform the agents about tickets with invalid ticket types.

Would be great, if you could update the PR accordingly.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for the follow-up. I updated PR as you suggested. I updated unit test for Type as well.
If you have any suggestion please let me know.

Regards
Zoran

zottto:
Hi @s7design,

looks really good already, but I recommend some slight changes:
- the Ticket::Type::Default should "default" by default (not "Unclassified")
- IMO the note that a type is invalid is not necessary in the customer interface
- at the moment, if the default type is assigned in the code, it is not checked that the default type really exists

At the moment, you are not able to change a ticket type if it is the default type. I would recommend to block the changes only if you change it to invalid and add a note to the screen, if you change the name to also update the sysconfig option Ticket:Type::Default accordingly.

Thanks,
Marc

zottto:
Sorry, but the Ticket::Type::Default should be "Unclassified" by default as you already added it to the sysconfig. I tested that with an old system...

s7design:
Hi @zottto 

Thanks for your suggestions it help me to improve code. I updated PR and there are also added new test cases in unit and Selenium tests.

Regards
Zoran

zottto:
Thanks, was merged with bc4ff655004cc304f5202be8b1c62bb6e05cad24 (master) and fc84d63695a22bb92f20580c691906201625a0f6 (rel-5_0).

PR #851, opened 2015-11-30
Master - removed AAA*.tt dummy preferences file

s7design:
Hi @mgruner 

Here is removed dummy preferences file. You can see, that I have added Translatable() function for some messages, and in some case I changed $LayoutObject->{LanguageObject}->Translate() with Translatable(). I think that there is not needed translate this message in these modules, because the message will be translated in Error.tt view ( 
https://github.com/OTRS/otrs/blob/master/Kernel/Modules/AgentPreferences.pm#L145
https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/Templates/Standard/Error.tt#L18 )
 
There are such cases here:
*Kernel/Output/HTML/Preferences/NotificationEvent.pm
*Kernel/Output/HTML/Preferences/Password.pm

Please let me know if I can improve it.

Regards
Zoran


mgruner:
Very nice, @s7design. Merged in a4d17072f6bcf137aee494fde90613af7ec40c01. Thanks a lot!

PR #852, opened 2015-12-01
[ Rel-5_0 Bug 11703 ] - An error while search

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11703

Hi @mgruner 

I added code in DBUpdate-to-5.pl to migrate Notify modules SySconfig items. I think that we omitted these ones.
I saw that during communication with bug reporter, you can see on the bug our conversation

Regards
Zoran


mgruner:
Very nice, merged, thank you!

PR #854, opened 2015-12-01
Fixed bug #11706 - Map -> Selections is not used

thorsteneckel:
See http://bugs.otrs.org/show_bug.cgi?id=11706 for further information.


mgruner:
Thanks for the suggestion, @thorsteneckel, but currently I'm not convinced that this really is a bug. Please see the information in the bug report.

thorsteneckel:
Thanks for your Feedback @mgruner . Maybe bug is not the right word? We think more it's a missing feature. From a practical point of view I don't understand why there is no way to give a classic "key" value a understandable display value. Some of our customers can't understand it, too. What do you think @nilei ? Thanks for your time.

nilei:
@thorsteneckel, thanks for involving me. I think that we should look for a more generic way to also translate other customer user data. Maybe we can add a "translatable" option to the map and let OTRS translate the value if a translation is available.

PR #855, opened 2015-12-02
cleaned up AAA*.tt dummy Ticket file

s7design:
Hi @mgruner,

Here is PR where I removed some of phrases/words for translating  in file AAATicket.tt. I hope that I did not cover too big chunk. If it is hard for follow up I will be able to split it in the smaller ones.

You can see in Ticket.xml that I added Translatable for some Description properties and in the list of Ticket data, and similar in AgentTicketZoom for HistoryTypeMapping. I have done it  to cover all word an phrases which should be translatable; some of them I will remove form AAABase.tt and the rest of AAATicket.tt. I thought it is good to fix it now. Is it OK?

* https://github.com/s7design/otrs/blob/a6efdd1bf20797888d1bb710b08e3be448fc8136/Kernel/Config/Files/Ticket.xml#L8340 
* https://github.com/s7design/otrs/blob/a6efdd1bf20797888d1bb710b08e3be448fc8136/Kernel/Config/Files/Ticket.xml#L8064
* https://github.com/s7design/otrs/blob/a6efdd1bf20797888d1bb710b08e3be448fc8136/Kernel/Modules/AgentTicketZoom.pm#L134

Please let me know if it is good, it will be nice if you merge it that I can continue to work on the rest of AAATicket.tt file.

Reagrds
Zoran


mgruner:
Very good, merged, thank you Zoran!

PR #857, opened 2015-12-03
Master - translate error and worning message

s7design:
Hi @mgruner 

Working on translate dummy files, I saw that there is not used translate in Error, Warning and similar ones views. Only in CustomerWarning.tt is used Translate for Message and Comment data. I added it for other views in this PR.

I change AgentTicketBulk.pm as an example how it could be done for all modules. It could be next task when I finish task with AAA*.tt dummy files.
If you like it you can merge to master.

Regards
Zoran


mgruner:
Merged, thanks!

PR #858, opened 2015-12-03
Master - proposal for adding translation for entities from DB in backend modules

s7design:
Hi @mgruner 

I have one proposal here. What do you think about adding Translatable function in System backend modules for phrases and words which represent initial  entities in database. Of course, there are Translatable in otrs-initial_insert.xml, and maybe this one is unneeded. These are specific word and often there are not anywhere in code where they could be marked for transliteration. 

If you think otrs-initial_insert.xml is enough I can just remove them from AAATicket.tt

Please let me know what do you think about my proposal.

P.S. Additional note:
I think these are not used anywhere.
- [% Translate("very low") | html %]
- [% Translate("low") | html %]
- [% Translate("normal") | html %]
- [% Translate("high") | html %]
- [% Translate("very high") | html %]

only these ones are used:

- [% Translate("1 very low") | html %]
- [% Translate("2 low") | html %]
- [% Translate("3 normal") | html %]
- [% Translate("4 high") | html %]
- [% Translate("5 very high") | html %]

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AAATicket.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AAATicket.tt
@@ -30,12 +30,6 @@
 [% Translate("4 high") | html %]
 [% Translate("5 very high") | html %]
 
-[% Translate("auto follow up") | html %]
-[% Translate("auto reject") | html %]
-[% Translate("auto remove") | html %]
-[% Translate("auto reply") | html %]
-[% Translate("auto reply/new ticket") | html %]
-
 # Templates
 [% Translate("Create") | html %]
 [% Translate("Answer") | html %]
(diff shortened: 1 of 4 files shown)


mgruner:
Hello Zoran,
initial_insert is enough. This is the place defining the entities, and this is therefore enough. Even if these entities are referenced in SysConfig (open, closed etc.) they don't have to be marked as translatable, as there they are only *used*, not defined. But marking them also does not hurt.
So I'll wait for you to remove the changes in the Perl files, ok?
Regards, mg

mgruner:
Can this be closed now, is it still relevant?

s7design:
We can close this one. Now there is Translatable only in otrs-initial_insert.xml, for these default entities names.
If you would like I will be able make a PR where I will go trough backend modules ( where it is necessary eg. Sate, Priority, Lock, AutoResponse, Notification name ... ) and add Translatable function for default entities as it is here in this proposal.

I can do that when I finish the rest of AAATicket.tt and AAABase.tt, or maybe it will not be needed.

What do you think about that?

Regards
Zoran

mgruner:
Hello Zoran, this will not be needed. Thanks.

PR #859, opened 2015-12-03
Master AAATicket translate part2

s7design:
Hi Martin,

There is the second part of removed words from AAATicket dummy templates.

Regards
Zoran


mgruner:
Merged, thanks a lot!

PR #860, opened 2015-12-04
Master - Removed strings from dummy template AAATicket

s7design:
Hi @mgruner 

This is the rest of AAATicket dummy template.
You can see I added Translatable for Notification event tooltip.

Regards
Zoran


mgruner:
Merged with one slight change in AgentTicketEscalation.tt. Thanks a lot!

PR #861, opened 2015-12-04
Removing trailing semicolon from grant command

edgardmota:
The semicolon at the end of the grant command (line 547) was causing error on remote database installation (error on granting permissions). Getting rid of it fix the issue.


mgruner:
Merged in master and rel-5_0. Thanks a lot @edgardmota!

PR #862, opened 2015-12-05
Fix bug#11721

noritnk:
This PR is fix for bug#11721.
See http://bugs.otrs.org/show_bug.cgi?id=11721 .


mgruner:
Merged, thanks for the fix @noritnk!

PR #865, opened 2015-12-07
Added ExternalTicketNumberRecognitionCompletion to outgoing mails

straubr:
OTRS does have the feature to recognize ExternalTicketNumbers and merge incoming mails. One of the requests we faced was, to add this number to outgoing mails if it's not in the subject already.

This PR adds a module that checks on outgoing mails whether a previously recognized ticketnumber is included in the subject and adds it if it's not. It does nothing if:
- The sender is not an agent
- The Article is not external
- ExternalTicketNumberRecognition is not activated
- The number is already in the subject

The format (left / right) can be configured in the sysconfig.

Also in this pull request I increased the number of recognition filters from 4 to 9.

As this is my first pull request, so please let me know if I don't comply with OTRS best practises.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -52,6 +52,17 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::ExternalTicketNumberCompletionFormat" Required="0" Valid="1">
+        <Group>Ticket</Group>
+        <SubGroup>Core::Ticket</SubGroup>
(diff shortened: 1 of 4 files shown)


mgruner:
Dear @straubr, thank you very much for your contribution. This is an interesting use case, and I am glad that you have found a working solution.

Currently I don't see a possibility of including this into the OTRS framework, mostly because of several architectural concerns. If you are interested in investing the effort, I can come up with a specification proposal for a generic solution that you could implement. Let me know if you are interested.

Best regards, Martin

PR #866, opened 2015-12-07
[ Rel-5_0 Bug 8971 ] - NewPhoneTicket toolbar button - does not check group settings

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8971

Hello @mgruner ,

Hereby is our proposal for fixing this issue. It could be advocated both ways as a bug and a new feature. IMO this is more of a new feature then a bug in system. Saying this, we thought that idea can have it's advantages so we implemented it.

Added in Header - Layout.pm check if Group key is set in SysConfig to confirm does user have permission to see that link and accordingly to show it or not. 
Additionally included more description for ToolBar modules in SysConfig, so user can be informed of feature and how to use it properly.
Lastly there is Selenium test in ToolBar/GroupRestriction.t for testing this functionality.

Please let me know if there are any question or issues regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -5716,7 +5716,7 @@
         </Setting>
     </ConfigItem>
     <ConfigItem Name="Frontend::ToolBarModule###1-Ticket::AgentTicketQueue" Required="0" Valid="0">
-        <Description Translatable="1">Toolbar Item for a shortcut.</Description>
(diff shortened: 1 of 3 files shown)


carlosgarciac:
Hi @s7design,

Thanks a lot for this nice fix, it was merged in e0f02c4 with some minor changes on the error message and comments.

Best regards,
Carlos García

PR #867, opened 2015-12-08
Fix AutoResponseSet if AutoResponseForWebTickets is set 0

moritz:
With this patch, the test passes regardless of the AutoResponseForWebTickets setting.


mgruner:
Merged, thank you @moritz.

PR #868, opened 2015-12-08
TemplateGenerator/AutoResponse test: Default language

moritz:
Makes the test pass independently of the globally configured default language


mgruner:
Merged, thanks.

PR #853, opened 2015-12-01
[ Rel-5_0 Bug 11705 ] - Ticket filters are displayed incorrectly

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11705

Hi
It was our mistake in the commit https://github.com/OTRS/otrs/commit/a01b6a25cc95ac06b5bd5bb45512761cb9d55f76

I have fixed it here.

Regards
Zoran

PR #870, opened 2015-12-09
[ Master Bug 11728 ] - Dashboard Widget: Attributes OwnerIDs is not working

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11728

Hi @mgruner ,

There is our proposal for fixing this bug. As you know there were many bugs similar to this one. I think that, with this possibility there you can add as many attributes as you want in SySConfig, it could cause many another problems with the filters on the Dashboard. Maybe there is needed big refactoring this code. 
I am not sure if we can do that, but for example if we add in atributes OwnerIDs, ResponsibleIDs, WatchUserIDs in the same time this filters will not be usable at all. I know this is extreme case but I saw that we fix this  TicketSearchSummary in any case when somebody want to add new attribute in SysConfig :) 

Maybe it is not so important and does not have top priority, but this is time for thinking about that and exchange ideas. Please let me know if we can help about that.

P.S. 
There is another PR which maybe could made merge conflict because the changes are on the same file. 
https://github.com/OTRS/otrs/pull/853/files 
If there will be need rebase PR please let me know.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
@@ -676,8 +676,8 @@ sub Run {
     # get config object
     my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
-    # show also watcher if feature is enabled
-    if ( $ConfigObject->Get('Ticket::Watcher') ) {
+    # show also watcher if feature is enabled and there is a watcher filter
+    if ( $ConfigObject->Get('Ticket::Watcher') && $TicketSearchSummary{Watcher} ) {
         $LayoutObject->Block(
             Name => 'ContentLargeTicketGenericFilterWatcher',
             Data => {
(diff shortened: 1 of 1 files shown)


Kernel/Output/HTML/Dashboard/TicketGeneric.pm
```diff
@@ -2103,16 +2103,16 @@ sub _SearchParamsGet {
...
         Watcher => {
-            WatchUserIDs => [ $Self->{UserID}, ],
-            LockIDs      => $TicketSearch{LockIDs} // undef,
+            WatchUserIDs => $TicketSearch{OwnerIDs} // [ $Self->{UserID}, ],
+            LockIDs      => $TicketSearch{LockIDs}  // undef,
         },
         Responsible => {
-            ResponsibleIDs => [ $Self->{UserID}, ],
-            LockIDs        => $TicketSearch{LockIDs} // undef,
+            ResponsibleIDs => $TicketSearch{OwnerIDs} // [ $Self->{UserID}, ],
```
mgruner: Should this be ```{ResponsibleIDs}``` instead?

mgruner:
Hello @s7design, I applied your PR with the one change mentioned. Please review this and let me know your feedback. Thanks a lot!

PR #872, opened 2015-12-09
Improved box sizing of input fields.

thorsteneckel:
Hi there,

thanks to @mrflix for giving us a hint to use this CSS attribute to align the input fields properly. This has positive effects for all kind of fields like dropdown, text etc.

Before:
before

After:
after

In case you will apply this PR it would be great if your could cherry-pick it for OTRS 3.3, 4 and 5, too.

Greetings,
  Thorsten

### DIFF
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Form.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Form.css
@@ -61,6 +61,7 @@ textarea {
     padding: 3px;
     font-size: 12px;
     color: #333;
+    box-sizing: border-box;
 }
 
 input[type=text]:focus,


rolfschmidt:
++

nilei:
++

mrcbnsls:
Thanks, nice. I think there are more places where this can be useful. I will try to have an eye on it :-)

Merged in 377b0ab, also for 5, 4 and 3 (exeptionally, because it's just such a small change).

mgruner:
@zottto: there is an issue with this. It breaks at least the search field in sysconfig and other places (with magnifying glass). Can you please either revert this or fix the problem *before* the next 5.0.4 release of OTRS?
@mrcbnsls: FYI.

zottto:
Had to revert this with 4aacdc6 because of styling problems in many other dialogs (SysConfig search input fiels, ACL editor, DanymicField admin dialog, ...). 
Adding box-sizing to all input fields in every context is a little bit too much, also because this change affects Agent and Customer interface.
Also reverted in 5,4 and 3.3.

@thorsteneckel Please change that to not break anything else in Agent or Customer interface.

Thanks,
Marc

thorsteneckel:
We'll see.

PR #873, opened 2015-12-09
Master - Removed strings from dummy template AAABase part 2

s7design:
Hi @mgruner 

Here is part 2 of cleaned dummy template AAABase

You can see in Defaults.pm that I add :

+    Translatable("Mr.");
+    Translatable("Mrs.");

I am not sure about that, your opinion will be welcomed. 

Regards
Zoran


mgruner:
Hello @s7design. I don't want any dependencies in Defaults.pm. Do we have any place in the code / data where Mr. / Mrs. are actually used? If not, we can just drop it.

s7design:
HI @mgruner 

I removed "Mr."  and "Mrs." 

There is in code only 'Mr/Mrs' as one phrase here
https://github.com/OTRS/otrs/blob/master/Kernel/System/Auth/Sync/LDAP.pm#L273

However, it is not the same as separated Mr. / Mrs, and because of that I just dropped it. 

Regards
Zoran

Kernel/Config/Files/Framework.xml
```diff
@@ -614,8 +614,8 @@
         <SubGroup>Core::LinkObject</SubGroup>
         <Setting>
             <Array>
-                <Item>Normal</Item>
-                <Item>ParentChild</Item>
+                <Item Translatable="1">Normal</Item>
```
mgruner: Why did you mark these? They are not on the AAA file.

s7design:
Hi @mgruner 

I updated PR, I removed Translatable for this two IDs. Thanks for follow-up, I was not sure about that when I had made PR.

Reagrads
Zoran

mgruner:
Hi Zoran, can you rebase this one please?

s7design:
Hi @mgruner 

I rebased it. I hope you will be able merge it now.

Regards
Zoran

PR #874, opened 2015-12-09
Fixed problem with UTF-8 argument passed to otrs.Console.pl (IB#1047660)

pboguslawski:
Please verify and fix:
https://dev.ib.pl/ib/otrs/issues/18

Regards,
Pawel


mgruner:
Hello @pboguslawski, thanks for your suggestion. I created a bug for this at http://bugs.otrs.org/show_bug.cgi?id=12008 but I solved it in a different way to also support terminals with non-utf8 encodings. Let me know if you see any problems.
https://github.com/OTRS/otrs/commit/224f784d5ae6dfd88522679320e7d066f4d0513d

PR #875, opened 2015-12-09
Fix `bin/otrs.SetPermissions.pl` when the application directory is accessed through a symlink

eliasp:
When the path, which is used to execute `bin/otrs.SetPermissions.pl`
contains a symlink component, the permissions will only be set for the
symlink itself, but `bin/otrs.SetPermissions.pl` won't recurse into the
OTRS directory itself, e.g. `/opt/otrs` points to `/opt/otrs-5.0.3` and
for the execution `/opt/otrs/bin/otrs.SetPermissions.pl` is used, then
the usage of `dirname` in `bin/otrs.SetPermissions.pl` will only dissect
the path of the executed script (to `/opt/otrs`), but won't resolve the symlink.

Any further mode/permission changes will now only be carried on the
symlink itself, which is just a file to those. By resolving the symlink
to the actual application directory first using `Cwd::abs_path()`, this problem
will be circumvented.

### DIFF
--- a/bin/otrs.SetPermissions.pl
+++ b/bin/otrs.SetPermissions.pl
@@ -21,6 +21,7 @@
 use strict;
 use warnings;
 
+use Cwd qw(abs_path);
 use File::Basename;
 use FindBin qw($RealBin);
 use lib dirname($RealBin);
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @eliasp. Thanks a lot for your proposal.
I wonder though that $RealBin already should resolve the symlinks and point to the "real" path. This also works fine for me:
```
sudo otrs5test/bin/otrs.SetPermissions.pl --web-group=staff --otrs-user=martin --admin-group=wheel
Setting permissions on /ws/otrs5
Skipping /.gitignore
Skipping /.tx
Skipping /.tidyall.d
Skipping /.settings
Skipping /.gnupg
Skipping /.git
``` 

otrs5test is a symlink to otrs5 in this case. I'll close this PR for now; please feel free to reopen with more information!

Best regards, Martin

PR #877, opened 2015-12-11
[ Rel-5_0 Bug 9837 ] - HTML content doesnÂ´t work in all notification tags

s7design:
http://bugs.otrs.org/show_bug.cgi?id=9837

Hello @mgruner ,

Hereby is our proposal for fixing this issue. Followed by example how HTML articles are handled for notifications, added similar check for auto response. Modified unit test AutoResponseSent.t to simulate auto response with <OTRS_CUSTOMER_BODY/SUBJECT[n]>  tags for HTML article.
Additionally, felt free to clean up code in TemplateGenerator.pm

Please let me know what do you think about this proposal and if there are any questions or issues regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -117,10 +117,13 @@ sub Salutation {
         ID => $Queue{SalutationID},
     );
 
+    # get HTMLUtils object
+    my $HTMLUtilsObject = $Kernel::OM->Get('Kernel::System::HTMLUtils');
+
     # do text/plain to text/html convert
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @s7design. Thanks a lot for your proposal. Unfortunately we had to rollback the HTML insertion in the notifications for two main reasons: impossibility of handling broken external HTML and possible interferences of the HTML code/styling with our templates. Now we are back with the original plain text insertion that we had before.
I also added a comment to the bug asking for the real problem here; right now I do not understand it.
I'll close this PR for now. Thank you! Please feel free to reopen if you disagree.
@carlosgarciac: ok?
Best regards, Martin

PR #878, opened 2015-12-11
Master - Removed strings from dummy template AAABase part 3

s7design:
Hi @mgruner 

Here is one more PR for dummy template AAABase.

You can see here that I added Translate for login message in Login.tt and CustomerLogin.tt, and I also ass Translatable for all messages in InterfaceAgent.pm and InterfaceCustomer.pm.

P.S. Working on this task I can see that we will have to likely update any else modules with Translate or Tranlatable function ( eg. for Notify, and errorScreen function). But we can discus about that when I finish work on  AAABase

Reagrds
Zoran


mgruner:
Excellent, thank you!

PR #880, opened 2015-12-16
[ Rel-5_0 Bug 9241 ] - Ticket searching in GenericAgent not working as expected

s7design:
http://bugs.otrs.org/show_bug.cgi?id=9241

Hello @mgruner ,

Hereby is our proposal for fixing this issue. There were no ContentSearch params when calling TicketSearch() in AdminGenericAgent.pm resulting in broken results when trying to search by part of a string. Added in backend part as well those parameters, so found tickets can be processed. Generic agent with these changes behave in same fashion as AgentTicketSearch allowing results by partial search.

Updated Selenium test to simulate this kind of behaviour. Please let me know what do you think about this proposal.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericAgent.pm
+++ b/Kernel/Modules/AdminGenericAgent.pm
@@ -1258,23 +1258,27 @@ sub _MaskRun {
 
     # perform ticket search
     my $Counter = $TicketObject->TicketSearch(
-        Result          => 'COUNT',
-        SortBy          => 'Age',
-        OrderBy         => 'Down',
-        UserID          => 1,
-        Limit           => 60_000,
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Sanjin,
I actually don't understand your solution. As long as ConditionInline is 1, the input string will always be split up by QueryCondition. In the ticket search this can be changed with the setting ```Ticket::Frontend::AgentTicketSearch###ExtendedSearchCondition```. That is the only way to make searching for a string (keeping word order) possible. You can use the Fred SQLLog module to see the generated statements.
I'll close this PR for now.
Best regards, Martin

PR #881, opened 2015-12-17
Removed strings from dummy template AAABase part4

s7design:
Hi @mgruner 

Here is one more PR for dummy template AAABase.

Regards
Zoran


mgruner:
Merged, thank you very much!

PR #883, opened 2015-12-18
Removed strings from dummy template AAABase part5

s7design:
Hello @mgruner 

This is the last commit for removing strings from dummy template. I did not removed file AAABase.tt. I am not sure about some strings and your help will be welcomed.

The folowing ones there are only in Defaults.pm. Where we can set Translatabele for these strings.
<pre>
<code>
[% Translate("Office") | html %]
[% Translate("Phone") | html %]
[% Translate("Fax") | html %]
[% Translate("Mobile") | html %]
[% Translate("Zip") | html %]
[% Translate("City") | html %]
[% Translate("Street") | html %]
[% Translate("Country") | html %]
[% Translate("Location") | html %]

[% Translate("Edit the system configuration settings.") | html %]
</code>
</pre>

For these I am not sure if they should be translatable
<code>
<pre>
[% Translate("Signed") | html %]
[% Translate("Sign") | html %]
[% Translate("Crypted") | html %]
[% Translate("Crypt") | html %]
</code>
</pre>
I can set here https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/ArticleCheck/PGP.pm#L256 and the same for SMIME. But is it needed?

And these are for rich text editor:
<code>
<pre>
# richtext editor
[% Translate("Bold") | html %]
[% Translate("Italic") | html %]
[% Translate("Underline") | html %]
[% Translate("Font Color") | html %]
[% Translate("Background Color") | html %]
[% Translate("Remove Formatting") | html %]
[% Translate("Show/Hide Hidden Elements") | html %]
[% Translate("Align Left") | html %]
[% Translate("Align Center") | html %]
[% Translate("Align Right") | html %]
[% Translate("Justify") | html %]
[% Translate("Header") | html %]
[% Translate("Indent") | html %]
[% Translate("Outdent") | html %]
[% Translate("Create an Unordered List") | html %]
[% Translate("Create an Ordered List") | html %]
[% Translate("HTML Link") | html %]
[% Translate("Insert Image") | html %]
[% Translate("CTRL") | html %]
[% Translate("SHIFT") | html %]
[% Translate("Undo") | html %]
[% Translate("Redo") | html %]
</code>
</pre>

Please let me know if I can make some improvement.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminACL.pm
+++ b/Kernel/Modules/AdminACL.pm
@@ -14,6 +14,7 @@ use warnings;
 our $ObjectManagerDisabled = 1;
 
 use Kernel::System::VariableCheck qw(:all);
+use Kernel::Language qw(Translatable);
 
 sub new {
     my ( $Type, %Param ) = @_;
@@ -34,8 +35,9 @@ sub Run {
 
     my $ACLID = $ParamObject->GetParam( Param => 'ID' ) || '';
 
-    my $SynchronizeMessage
-        = 'ACL information from database is not in sync with the system configuration, please deploy all ACLs.';
+    my $SynchronizeMessage = Translatable(
+        'ACL information from database is not in sync with the system configuration, please deploy all ACLs.'
+    );
 
     my $SynchronizedMessageVisible = 0;
 
--- a/Kernel/Modules/AdminPackageManager.pm
+++ b/Kernel/Modules/AdminPackageManager.pm
@@ -13,6 +13,7 @@ use strict;
 use warnings;
 
 use Kernel::System::VariableCheck qw(:all);
+use Kernel::Language qw(Translatable);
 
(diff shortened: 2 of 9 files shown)


mgruner:
Hello Zoran,

thanks for your research so far. The first group of strings you mention is about column names for the customer user db. Here we could have a fake data insert element in initial_insert.xml that is commented out and has the right name for each column marked as translatable.

The second group looks like it could be made translatable in the file you mentioned, perhaps also the corresponding "Value" string; you should check this in detail.

The third group could be placed in a commented section in RichTextEditor.tt *if* it were needed, but I am not really convinced about this. If the string don't appear in any file in the codebase (as it seems), they could be removed maybe?

Regards, Martin

mgruner:
Merged the current state, thank you!

PR #887, opened 2015-12-22
Removed strings from dummy template AAABase part6

s7design:
Hi @mgruner,

There is tidied up the rest of AAABase dummy template. 

I added a fake insert in otrs-initial_insert.xml, I think it is good solution maybe also for some things in future if it will be necessary. 

The strings for sign and crypt were added as you sugested. The Value are translated as well, because of that I add Translatable for this Value AgentTicketZoom.tt

I checked about rich-text string, they are not necessary to be translatable in the framework. There are translate files here for this third-party plugin :
https://github.com/OTRS/otrs/tree/master/var/httpd/htdocs/js/thirdparty/ckeditor-4.5.4/lang

Regards
Zoran


mgruner:
Hello Zoran, thanks a lot! I made a slight change as I reconsidered the fake insert. Optimally the strings are marked where they are *defined*. So in the end I added a no-op method ```Translatable()``` to Defaults.pm and used it in the Customer Map where the strings are actually defined. This avoids the dependency and reduces the risk of getting out of sync. Please let me know your opinion!
Regards, Martin

PR #888, opened 2015-12-23
Master - Translate some strings in Default.pm

s7design:
Hi @mgruner 

I like your idea about adding Translatable() no-op to Default.pm. It is better, as you said, in that case there is set for translation in the place where the strings are. 

In this PR I added Translatable to  some other strings which have been omitted before. If you like it you can merge it.

PS. I fix some spell mistake as well.

Regards
Zoran


mgruner:
Excellent, thanks!

PR #889, opened 2015-12-23
Master - Removed GetTableData from AdminQueue

s7design:
Hi @mgruner 

As we had spoken before about removing GetTableData, I removed it form AdminQueue. You can see I changed appropriate tests. I added RestoreDatabase instead manually delete.

I changed a little GroupList, now it is similar to other backend module function for get List of entities. Please check if it is good now. I run all test and it has the same behavior, but I removed extra cache and return statement.  Now it is set in cache only List which is used in that moment ( Valid or Invalid ones). 

For GetFollowUpOptionList I added new function, it could be update GetFollowUpOption, but new function in this moment seemed better. 

Regards
Zoran

### DIFF
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -346,39 +346,55 @@ sub GroupList {
     );
     return %{$Cache} if $Cache;
 
-    # create the valid list
-    my $ValidIDs = join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet();
+    # get valid ids
+    my @ValidIDs = $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet();
 
-    # build SQL
-    my $SQL = 'SELECT id, name FROM groups';
+    # get group data list
+    my %GroupDataList = $Self->GroupDataList();
 
-    # add WHERE statement
-    if ($Valid) {
-        $SQL .= ' WHERE valid_id IN (' . $ValidIDs . ')';
+    my %GroupListValid;
+    my %GroupListAll;
+    KEY:
+    for my $Key ( sort keys %GroupDataList ) {
+
+        next KEY if !$Key;
+
+        # add group to the list of all groups
+        $GroupListAll{$Key} = $GroupDataList{$Key}->{Name};
+
+        my $Match;
+        VALIDID:
+        for my $ValidID (@ValidIDs) {
+
+            next VALIDID if $ValidID ne $GroupDataList{$Key}->{ValidID};
+
+            $Match = 1;
+
+            last VALIDID;
+        }
+
+        next KEY if !$Match;
+
+        # add group to the list of valid groups
+        $GroupListValid{$Key} = $GroupDataList{$Key}->{Name};
     }
 
-    # get database object
-    my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
-
-    # get group data from database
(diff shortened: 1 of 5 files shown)


Kernel/System/Group.pm
```diff
@@ -346,55 +346,39 @@ sub GroupList {
...
-    my %GroupListValid;
-    my %GroupListAll;
-    KEY:
-    for my $Key ( sort keys %GroupDataList ) {
-
-        next KEY if !$Key;
-
-        # add group to the list of all groups
-        $GroupListAll{$Key} = $GroupDataList{$Key}->{Name};
+    # create the valid list
```
mgruner: Please revert the changes in Group.pm. This is carefully designed cache usage for performance optimization.

Kernel/System/Queue.pm
```diff
@@ -636,6 +636,58 @@ sub GetFollowUpOption {
     return $Return;
 }
 
+=item GetFollowUpOptionList()
```
mgruner: It seems that there are no unit tests yet for this function, can you please check?

Kernel/System/Group.pm
```diff
@@ -346,55 +346,39 @@ sub GroupList {
...
-        for my $ValidID (@ValidIDs) {
-
-            next VALIDID if $ValidID ne $GroupDataList{$Key}->{ValidID};
+    # build SQL
+    my $SQL = 'SELECT id, name FROM groups';
 
-            $Match = 1;
+    # add WHERE statement
+    if ($Valid) {
+        $SQL .= ' WHERE valid_id IN (' . $ValidIDs . ')';
```
mgruner: BTW: if you send Valid => 0 you would not have a WHERE clause...

Kernel/System/Queue.pm
```diff
@@ -636,6 +636,58 @@ sub GetFollowUpOption {
...
+
+    # create the valid list
+    my $ValidIDs = join ', ', $Kernel::OM->Get('Kernel::System::Valid')->ValidIDsGet();
+
+    # build SQL
+    my $SQL = 'SELECT id, name FROM follow_up_possible';
+
+    # add WHERE statement
+    if ($Valid) {
+        $SQL .= ' WHERE valid_id IN (' . $ValidIDs . ')';
```
mgruner: Same here. Valid => 0 -> no WHERE clause. UTs would show you.

mgruner:
Please see my comments.

s7design:
HI @mgruner 

I rebased this PR and updated based on your suggestions. I added test for GetFollowUpOptionList in Queue.t. Regarding to your suggestion about _'Valid => 0 -> no WHERE clause'_ in my opinion where clause is not necessary if Valid = 0. In that case there will be fetched all follow-up p options no matter valid or not. I wanted that, please let me know if I wrong somewhere. 

I hope it is well know, i will be able to change something if you have any suggestion. 

Regards
Zoran

mgruner:
Nice, thank you Zoran. There was a problem in Group.t that causes failure on postgresql: it deliberatly tries to insert a group that already exists (this must fail). But after that the transaction is aborted and all subsequent SQL statements are ignored. I fixed this by adding a check in GroupAdd before the INSERT statement. This is relevant for the activation of RestoreDatabase in the other unit tests; please use postgresql for that as it is stricter than other databases. Best regards, Martin

PR #890, opened 2015-12-24
[ Master - Bug 11771 ] - ACL AgentTicketPhone

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11771

Hi @mgruner 

Here is proposal how this bug could be fixed. To be honest it is not bug in my opinion. It could be new behavior and it would be helpful but such behavior cause many side effect, for example with AJAX update fields.

I implemented select mandatory selection fields if there is only one value. I tried to add similar code for each of these fields on load screen, however after testing in details I saw it is not good solution. There is problem if there is for example one valid Queue, it will be set as one value on load e.g AgentTicketPhone, and if there is ACL which contain this Queue. In  that case AjaxUpdate will not check ACL and it could confuse customer of system. 

Because of that I said it is proposal for fixing, it is not final solution. I did not want waste more time in this moment, but if this solution is acceptable I will be able to improve it and add Selenium test as well.

P.S. I deleted commented code in AgentTicketEmail.pm for conversion Signature

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketEmail.pm
+++ b/Kernel/Modules/AgentTicketEmail.pm
@@ -1702,11 +1702,18 @@ sub Run {
             }
         }
 
-        # convert Signature to ASCII, if RichText is on
-        if ( $LayoutObject->{BrowserRichText} ) {
+        # select mandatory fields if there is only one value
+        my @DestKeys = keys %{$NewTos};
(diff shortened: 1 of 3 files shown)


mgruner:
I agree with you that this is not a bug, please see my comments in the bug.

mgruner:
If we ever decided to change this, a JavaScript based generic solution might be interesting.

PR #891, opened 2015-12-24
Master - switched to Verified methods part1

s7design:
Hi @mgruner 

This is first part of the task for switch to Verified methods. I update some Selenium test randomly, but I am planing that we  continue  in some order one by one tests. I updated also AdminEmail.t and  AdminSignature.t which you committed 13 days ago. I wanted to cover all three methods.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminEmail.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminEmail.t
@@ -80,7 +80,7 @@ $Selenium->RunTest(
         # check client side validation
         my $Element = $Selenium->find_element( "#Subject", 'css' );
         $Element->send_keys("");
-        $Element->submit();
+        $Element->VerifiedSubmit();
 
         $Self->Is(
             $Selenium->execute_script(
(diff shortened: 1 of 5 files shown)


mgruner:
Looks good. At least in one place the VerifiedSubmit() would not be needed as no form reload is expected (error instead), but it is probably good to be consistent and use it whenever possible.

Whenever we cannot use it (because of closing window), a comment should be added explaining the reason IMHO.

If you go through the tests, this might be a chance of removing the Selenium based ticket creation in test files that should test other things. Right now usually AgentTicketPhone is called to create the ticket; this could be replaced by a simple backend call. What do you think?

mgruner:
Merged, thank you very much! One more thing: can you please create all the future UT-related PRs for rel-5_0 as we need them there as well?

mgruner:
I will then also cherry-pick them to master.

s7design:
Thanks Martin,

I will make PRs for rel-5_0, it is not problem. And of course it is good advice that we create tickets in beckend where it is possible.

Regards
Zoran

PR #892, opened 2015-12-28
Rel-5_0 AdminGroup updated test

s7design:
Hi Martin,

I saw that we omitted to switch one submit in this test. Actually, I changed two submit, one of them is not necessary as you said in the previous PR but we will do that to be more consistent. 

<pre>
         $Selenium->find_element( "#GroupName", 'css' )->clear();
-        $Selenium->find_element( "#GroupName", 'css' )->submit();
+        $Selenium->find_element( "#GroupName", 'css' )->VerifiedSubmit();
</pre>

However, the second one is omitted and there is necessary VerifiedSubmit.
Now, I make in rel-5_0 and next update test will be done in this one.

Regards
Zoran


mgruner:
Merged, thanks.

PR #893, opened 2015-12-28
Rel-5_0  - Selenium update output ticket 

s7design:
Hi Martin,

There is one PR with switched Selenium test to Verified methods. You can see some Waitfor and sleep I removed but some of them are left. Please check if it is good now. I tested in PhantomJS and FF, and I think that these Wait for are  still necessary ( Overview* tests).

Please let me know if I can improve it.

Regards
Zoran


mgruner:
Yes, WaitFor is still needed if no page reload happens (but only AJAX). I thought that the mod_perl waits are also needed (because mod_perl needs to pick up the new ZZZProcessManagement.pm file), but if not, even better. Merged, thanks!

PR #899, opened 2015-12-30
[ Rel 5 0 Bug 11710 ] - Using ArticleFilter causes numeration issues 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11710

Hi Martin,

This is out proposal for fixing. As you can see in the comments of the bug reports, we were not sure is it necessary that there is numbering of ticket with filter the same as it is without filter. Maybe our solution is not the best one. Please let me know if we can improve it.
I think that we can slow down system in some case because there is loop through all article only to get numbers of the articles for filtered or paginated articles. 

What do you think about all of that?

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -782,6 +782,14 @@ sub MaskAgentZoom {
         $Pages = ceil( $ArticleCount / $Limit );
     }
 
+    my $Count;
+    if ( $ConfigObject->Get('Ticket::Frontend::ZoomExpandSort') eq 'reverse' ) {
+        $Count = scalar @ArticleBox + 1;
+    }
+    else {
+        $Count = 0;
+    }
+
     # get all articles
     my @ArticleContentArgsAll = (
         TicketID                   => $Self->{TicketID},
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Hi @s7design,

I think the UT has several issues. Please check again:

(1) Sometimes the filter dialog close button can't be clicked because the dropdown field is still open and is hiding the submit button.

(2) After doing this:

````
$SysConfigObject->ConfigItemUpdate(
    Valid => 1,
    Key   => 'Ticket::Frontend::ZoomExpandSort',
    Value => 'normal',
);
````

you probably need to wait to make sure this has been applied correctly.

(3) Doing this:

````
$Selenium->find_element( "#ResetArticleFilter", 'css' )->VerifiedClick();
````

is probably not enough, because it's not a real link but a javascript trigger. Better wait until the button is gone (which is the case once the filter has been removed).

(4) After resolving these issues, I still get this error:

````
AgentTicketZoomArticleFilter.t #53 - ZoomExpandSort: reverse - Article #4 – Fourth Test Article found on page with original numeration - article filter on
````

Can you please have a look? I know the PR is very old, but I'm currently trying to get down the number a bit :-D

Thanks!

Marc

**Comment by s7design** [comment]
Hi @mrcbnsls 

Selenium test is updated:
(1) Expanded dropdown menu is collapsed to prevent submit button overlapping.
(2) After config item update, 'VerifiedRefresh' is added.
(3) After click on reset filter button, 'WaitFor' is added to be sure button has gone.

Regards
Zoran

**Comment by mrcbnsls** [comment]
Thanks for the update, but the UT is still failing for me:

````
 not ok 54 - ZoomExpandSort: reverse - Article #4 – Fourth Test Article found on page with original numeration - article filter on
.Selenium command success: $VAR1 = {
  'args' => [],
  'command' => 'executeScript',
  'script' => 'return typeof(Core) == "object" && typeof(Core.App) == "object" && Core.App.PageLoadComplete',
  'session_id' => '1181ca26-e599-4079-b2b7-6afdc00b6cdf'
};

 not ok 56 - Exception in Selenium': An element could not be located on the page using the given search parameters: #ResetArticleFilter,css selector at (eval 100) line 235.
````

**Comment by s7design** [comment]
Hi @mrcbnsls 

We changed the test a little bit. You will see there are added WaitFor and VerifiedRefresh() instead of VerifiedClick for test case in 'filter dialog'. 

In any case the test worked fine in our local system:

```
+-------------------------------------------------------------------+
/opt/otrs/scripts/test/Selenium/Agent/AgentTicketZoomArticleFilter.t:
+-------------------------------------------------------------------+
...............................................................................
=====================================================================
 Product:     OTRS 5.0.x git
 Test Time:   32 s
 Time:        2016-09-26 15:05:53
 Host:        localhost
 Perl:        5.14.2
 OS:          Linux
 Vendor:      ubuntu 12.04
 Database:    postgresql 9.1.20
 TestOk:      79
 TestNotOk:   0
=====================================================================

I know that it sounds strange again a test work in one system and did not in another one :) but I don't have idea what could be the problem.

Regards
Zoran
```

mrcbnsls:
Works fine and looks good now, thanks! Merged in 0b762f1 and cb76dd8.

PR #902, opened 2015-12-31
Rel-5_0 - Selenium update Admin part 2

s7design:
Hello @mgruner ,

Hereby are updated Selenium tests. There was an issue with Admin.t when trying to access AdminSMIME. Added certificate and private keys so page can be loaded and tested, which are cleaned on the end.

AdminGenericAgent.t had an issue if test failed or prematurally ended, so tickets remained. Corrected that adding RandomID on ticket title which is search parameter for generic agent.

Regards,
Sanjin

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
@@ -23,7 +23,7 @@ $Selenium->RunTest(
         # get helper object
         my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
 
-        # create and login test user
+        # create test user and login
         my $Language      = 'de';
         my $TestUserLogin = $Helper->TestUserCreate(
             Groups   => ['admin'],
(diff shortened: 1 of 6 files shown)


scripts/test/Selenium/Agent/Admin/AdminDynamicField.t
```diff
@@ -193,16 +173,12 @@ JAVASCRIPT
 
                 $Selenium->find_element(
                     "//a[contains(\@data-query-string, \'Subaction=DynamicFieldDelete;ID=$DynamicFieldID' )]"
-                )->click();
-
-                $Selenium->WaitFor(
```
mgruner: I believe this is still needed as the delete operation does not cause an immediate reload.

scripts/test/Selenium/Agent/Admin/AdminGenericInterfaceWebservice.t
```diff
@@ -131,10 +133,7 @@ $Selenium->RunTest(
 
             # delete web service
             $Selenium->find_element( "#DeleteButton",  'css' )->click();
-            $Selenium->find_element( "#DialogButton2", 'css' )->click();
-
-            # wait until delete dialog has closed an action performed
```
mgruner: Same here.

mgruner:
I merged this with the two changes mentioned. Please make sure to carefully check that the required WaitFor calls are still in place. This is always the case if there is no immediate page reload but still something needs to be waited for. Thanks!

PR #903, opened 2016-01-04
Rel-5_0 - Selenium update Admin part 3

s7design:
Hello @mgruner 

Hereby are updated selenium tests for Admin modules. AdminPackageManager.t is not working if test failed or prematurely ended. When test package is left in the system. Added backend check on start of the test for it and removing accordingly.

Please let me know if there are any issues regarding this PR.

Regards,
Sanjin


mgruner:
Very nice, merged, thanks!

PR #907, opened 2016-01-07
Do not let the user know whether a ticket exists or not

reneeb:
This would mean an information leak and could be used to try to get more information.


mgruner:
Merged, thank you very much!

PR #908, opened 2016-01-08
Fix Bug#11786 - Reference of undefined variable in AgentTicketSearch

noritnk:
This PR is fix for bug#11786.
See http://bugs.otrs.org/show_bug.cgi?id=11786 .


s7design:
Please see https://github.com/OTRS/otrs/pull/848

Thanks 
Zoran
S7Design

mgruner:
PR applied to fix the one issue mentioned in the bug. Thanks @noritnk!

PR #911, opened 2016-01-09
Rel-5_0 selenium update Output

s7design:
Hi Martin,

Here are some updated Selenium tests from Selenium/Outout with swiched to Verified methods.

Regards
Zoran


mgruner:
Very nice, thanks, merged!

PR #912, opened 2016-01-10
Rel-5_0 selenium update Output/NavBar

s7design:
Hi Martin,

You can see that I removed "sleeping to allow mod_perl to pick up the changed config files". There were some such sleeping in the test which are updated in this PR. I know that you mentioned that it is necessary in your system and I suppose the main OTRS test server, however it is not needed on my system. Please let me know if you see some problem testing on your system these Selenium tests. I tested it with FF and PhantomJS as well. All tests work fine for me without these "sleeps".

I will be able to rollback on previous state if it is necessary.

Regards
Zoran


mgruner:
Please keep the mod_perl sleeps for now. Do you have mod_perl actually enabled? You won't see the problem otherwise.

mgruner:
I saw it recently also for a SysConfigUpdate() right in the middle of a test, that a sleep is needed afterwards to have the test work.

s7design:
Hi Martin,

Thanks, I updated PR. All mod_perl sleeps are returned back.
As you know I cannot test it because it works the same on my system. Please let me know if I can change something else.

Regards
Zoran

mgruner:
Merged, thank you very much!

PR #918, opened 2016-01-12
Rel-5_0 - switched to VerifiedRefresh

s7design:
Hi Martin,

There are updated Selenium test which had updated the previous days where there was refresh() function. 

There are in some tests in Output/Dashboard and  Output/ToolBar but it will be done there in separate PRs when we switch other Verified methods as well.

Regards
Zoran


mgruner:
Merged, thanks. Nice, we even improved places where WaitFor was missing.

PR #920, opened 2016-01-12
Rel-5_0 - Selenium updated test

s7design:
Hi Martin,

Here is updated Selenium test  WindowHeight.t which was added yesterday and some more tests that were in different directories. 

Regards
Zoran


mgruner:
Merged, thanks a lot!

PR #923, opened 2016-01-13
Rel-5_0 - Selenium updated test - Output/ToolBar part 1

s7design:
Hi Martin,

There are the first part of updated Selenium test from Output/Toolbar

Regards
Zoran


mgruner:
Merged, thank you very much!

PR #926, opened 2016-01-18
Bugfix: Load missing Customer Company names (CustomerID search).

balazj:
Please, take a look at [http://bugs.otrs.org/show_bug.cgi?id=11808](http://bugs.otrs.org/show_bug.cgi?id=11808).

We got 600+ registered customers in our system. 

System loads only 250 Customers, so it might happen that some Customer information is not included. This patch allows to load all Customers available in the system.

This fix could be applied on the "master" branch also.

If you have any questions, please don't hesitate to ask.

BR,
Jaroslav

### DIFF
--- a/Kernel/Modules/AgentCustomerInformationCenterSearch.pm
+++ b/Kernel/Modules/AgentCustomerInformationCenterSearch.pm
@@ -46,6 +46,7 @@ sub Run {
 
         my %CustomerCompanyList = $Kernel::OM->Get('Kernel::System::CustomerCompany')->CustomerCompanyList(
             Search => $ParamObject->GetParam( Param => 'Term' ) || '',
+            Limit => 0,
         );
 
         # add CustomerIDs for which no CustomerCompany are registered


mgruner:
Hello Jaroslav, thanks a lot for your proposal. I don't fully understand your problem. IIRC this search is for the autocomplete result list that is only triggered if you input a few characters for the search ("Term"). So you would never get a full result list with all customers, and that might also not be so helpful in this place. Can you please explain?

balazj:
Yes, I already thought about this - this bug occurs only when there are a lot of hits ( > 250). That actually happened when our user searched for the term "**". I can agree with you that this field is not designed for this purpose (it anyway displays only first 20 results), however we can't predict users behavior. 
If you consider this as a performance issue, i will understand if you don't merge this request.

mgruner:
Hello @balazj. Thanks for the update. Yes, this is also a performance safety net. Thanks!

PR #927, opened 2016-01-18
[ Rel-5_0 Bug 9241 ] - Ticket searching in GenericAgent not working as expected - ver2

s7design:
Hello @mgruner ,

Thank you for your feedback on previous PR regarding this issue. You are right regarding ConditionInline param. Although without 'ContentSearchPrefix' and 'ContentSearchSuffix' params in TicketSearch() for GenericAgent, if user want to search by only 2-3 words extracted from ticket title he must then add * before and after each word. This behaviour is not present in AgentTicketSearch where those params are sent to TicketSearch() function. 
On the end i think it's better to go forward with this approach ( without prefix and suffix params ) so there is clear option left for user if and how they want to use ticket search function in AdminGenericAgent. Additionally it can prevent some unexpected tickets to show in search result causing further unwanted ticket actions on those tickets.

Added new sysconfig 'Ticket::GenericAgentTicketSearch###ExtendedSearchCondition' to control ConditionInline param. In description of a config added how to use appropriately if user wants to search by multiple words that are not next to each-other e.g. (*key1*||*key2*)

Update Selenium tests to search first tickets when this sysconfig is enabled, expecting no results found, since search input is "*Ticket $TestTicketRandomID Generic*". After disabling sysconfig, test find expected tickets and continue with module testing.

Please review this updated proposal and let me know if there are any issues or remarks.

Kind regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -9288,6 +9288,17 @@
             </Hash>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::GenericAgentTicketSearch###ExtendedSearchCondition" Required="1" Valid="1">
+        <Description Translatable="1">Allows extended search conditions in ticket search of the generic agent interface. With this feature you can search e. g. ticket title with this kind of conditions like "(*key1*&amp;&amp;*key2*)" or "(*key1*||*key2*)".</Description>
+        <Group>Ticket</Group>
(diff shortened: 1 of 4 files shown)


mgruner:
Merged, thanks a lot, @s7design. Please note that after all config setting changes that appear during a Selenium test, we need a sleep 1; to make it work on fast mod_perl systems.

PR #928, opened 2016-01-19
Bug 11810 - <OTRS_TICKET_State> not translated

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11810

Hi Martin,

This is our proposal for fixing. There was a problem because in the Standard Template was not the language selected in TemplateGenerator module.

I extended Selenium test with this test case as well. In my opinion we can add Language parameter also for Signature and Salutation.
In hint for these two entities there is possibility to add:

<code>
...

<OTRS_TICKET_*>
    Options of the ticket data (e. g. OTRS_TICKET_TicketNumber, OTRS_TICKET_TicketID> OTRS_TICKET_Queue, OTRS_TICKET_State). 

...
</code> 

Regards
Zoran

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -412,6 +412,25 @@ sub Template {
         );
     }
 
+    # get user language
+    my $Language;
+    if ( defined $Param{TicketID} ) {
+
+        # get ticket data
+        my %Ticket = $Kernel::OM->Get('Kernel::System::Ticket')->TicketGet(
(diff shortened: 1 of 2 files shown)


Retiebralanak:
Hi S7,

thanks a lot for the fix & test. It will be included in 5.0.8.

5.0.8 (2c83a8cd963c85c5afa6b064b165836526d7e28d)
master (de7e4125fbd6a516d4a6227326ccebda4534010a)

Regards,
Jan

PR #932, opened 2016-01-22
[ Rel-5_0 Bug 11819 ] - Error when trying to edit/create new customer agent while SMIME activated 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11819

Hi Martin,

There was a problem because there did not updated some modules in the script  DBUpdate-to-5.pl
Now, I check all sysconfig items which have Module param. I hope that now it works for all items. 

If you want I can split this PR in two ones, because there is fixed bug with changes for   "Preferences Modules" and fixed "Article View Modules" that is not mentioned in the bug. I thought that it is better solve in one PR. 

Regards
Zoran

### DIFF
--- a/scripts/DBUpdate-to-5.pl
+++ b/scripts/DBUpdate-to-5.pl
@@ -1243,11 +1243,12 @@ sub _MigrateConfigs {
 
         # update module location
         my $Module = $Setting->{$ToolbarModule}->{'Module'};
-        if ( $Module !~ m{Kernel::Output::HTML::ToolBar(\w+)} ) {
-            next TOOLBARMODULE;
-        }
+        next TOOLBARMODULE if ( $Module !~ m{Kernel::Output::HTML::ToolBar(\w+)} );
 
         $Module =~ s{Kernel::Output::HTML::ToolBar(\w+)}{Kernel::Output::HTML::ToolBar::$1}xmsg;
+        next TOOLBARMODULE if ( $Setting->{$ToolbarModule}->{'Module'} eq $Module );
+
+        # update sysconfig if it is necessary
         $Setting->{$ToolbarModule}->{'Module'} = $Module;
 
         # set new setting,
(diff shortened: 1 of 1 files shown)


scripts/DBUpdate-to-5.pl
```diff
@@ -1339,33 +1339,45 @@ sub _MigrateConfigs {
     print "...done.\n";
     print "--- Preferences group modules...";
 
-    # Preferences groups
-    $Setting = $ConfigObject->Get('PreferencesGroups');
+    # Preferences Modules
+    my @PreferencesGroupTypes = (
+        {
+            Path => 'PreferencesGroups',
```
mgruner: This looks very good in general. However, I have one question: why do you use a list of hashes here? Would not a simple list ```@PreferencesGroupTypes = qw(PreferencesGroups CustomerPreferencesGroups)``` be enough? I see that it is already used on other places in this function, but maybe you could consider a simplification.

s7design: Thanks Martin, you are right. I will change it and update this PR very soon.
Regards
Zoran

s7design:
Hi @mgruner 

I updated this PR as you suggested

Regards
Zoran

mgruner:
Very nice, thanks! I have one more small request: the settings should only be updated if they actually change (Module vs. NewModule).

s7design:
Hi @mgruner,

Thanks for your suggestion, I think it is better now to update only sysconfig which are changed. Please let me know if my proposal good now.

Regards
Zoran

mgruner:
Merged with slight changes, I modified two more places (cluster configuration and customer queue selection generic) to only update if needed. Please let me know in case you see any problem there. Thanks!

s7design:
Thanks Martin, I tested it and it works fine. I agree with you it is better now with your changes.

Regards
Zoran

PR #933, opened 2016-01-22
Rel-5_0 - Added unittest for AutoResponseWithoutQueue - bug#6350

s7design:
Hi @mgruner,

Here is the unit test for AutoResponseWithoutQueue. I updated test with cleaning-up by RestoreDatabase. The test is added in TemplateGenerator/AutoResponse.t, I hope it is OK but I can make new one if it is necessary. 

Regards
Zoran


carlosgarciac:
Hi @s7design,

thanks a lot for this unit test, it was included in 2ddb53d.

Best regards,
Carlos García

PR #843, opened 2015-11-26
[Rel-5_0 Bug 9950] - Ticket Split takes uses system address as the customer

s7design:
http://bugs.otrs.org/show_bug.cgi?id=9950

Hello,

Hereby is our proposal for fixing this issue. Next to already existing queue check, added to check if system address is in Article{From} parameter when Article{To} is not a Queue.

Created Selenium test to simulate three different split article cases covering different scenarios. 

Please let me know if there are any questions or issue regarding this PR. 

Regards,
Sanjin

PR #934, opened 2016-01-26
[ Rel-5_0 Bug 11826 ] - Split ignores CC

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11826

Hello @mgruner ,

This is our proposal for fixing this reporters issues. Must admit that reproducing this kind of behaviour is more artificial then natural work flow. Current OTRS Split function IMO works as expected. Created Selenium test to simulate this type of behaviour.

Please if you could consider PR ( https://github.com/OTRS/otrs/pull/843 ) together when evaluating this proposal.

Kind regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketPhone.pm
+++ b/Kernel/Modules/AgentTicketPhone.pm
@@ -319,6 +319,9 @@ sub Run {
                     $ArticleFrom = $Article{To};
                 }
             }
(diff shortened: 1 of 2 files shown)


mgruner:
Hello Sanjin, thanks a lot for your proposal, but at this point I am not (yet) convinced that this really is a bug. Ticket split is not intended to copy the current ticket.

PR #940, opened 2016-01-27
[ Rel-5_0 Bug 11816 ] - Ticket::Service::KeepChildren parameter - invalid Parent Service

s7design:
http://bugs.otrs.org/show_bug.cgi?id=118161

Hello @mgruner .

Hereby is proposal for fixing reporters issues.

In Preferences/CustomService.pm there was missing param KeepChildren.

In AgentTicketService.pm , even though there was no KeepChildren param, in order to reach sub level of service ( child ) we have to be able to click on parent to expand it. Edited Valid param to 0, now all service are visible even if they are valid or not. This is same fashion as how AgentTicketQueue works ATM.

Additionally i went through code and edited KeepChild param in AdminGenericAgent, AdminNotificationEvent and AdminSLA to get value from Config. In AdminSLA where we call in Overview screen for ServiceList, removed KeepChild param since we ask for all services not just valid ones, IMO there is no real point for it there, since it doesn't have an effect.

Modified Selenium test for AgentTicketService to now check tickets in child service, when parent service is set to invalid reproducing reporter issue.

Lastly, would like to ask you what do you think of proposal for future task to implement some notice for invalid ticket parameters for Queue, Service, ect. , like how it is done for Customer. 
screen shot 2016-01-27 at 1 22 46 pm

Please review this proposal and let me know if there are any issues or questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericAgent.pm
+++ b/Kernel/Modules/AdminGenericAgent.pm
@@ -830,7 +830,7 @@ sub _MaskUpdate {
         # get list type
         my %Service = $Kernel::OM->Get('Kernel::System::Service')->ServiceList(
             Valid        => 1,
-            KeepChildren => 1,
+            KeepChildren => $ConfigObject->Get('Ticket::Service::KeepChildren'),
             UserID       => $Self->{UserID},
         );
         my %NewService = %Service;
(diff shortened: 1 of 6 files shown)


UdoBretz:
Thanks a lot! Merged!

PR #941, opened 2016-01-28
[ Rel 5 0 Bug 11822 ] - S/MIME support is disabled in Kernel::Config::SMIME

s7design:
Hi @mgruner,

Hera is our proposal for fixing this bug and improving AdminSMIME screen in case SMIME is disabled.
You can see, I have made two different widget which displayed in cases SMIME sysconfig item is disabled or some other SMIME config item is not configured well.

I added result table with NoDataFoundMsg message in case SMIME  is not working or disabled, it is better layout for me, but I wanted to ask you what do you think about that.

I check if SMIME configured  well at the beginning, since I  removed if statement that had been before on other places in AdminSMIME.pm module.

If you like it I will be able to improve AdminPGP screen, I would do that in another PR. Now in AdminPGP you can go to Action=AdminPGP;Subaction=Add even CryptObject is not created. I suggest removing Action widget to avoid such cases. 

screen shot 2016-01-28 at 12 11 11 pm

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminSMIME.pm
+++ b/Kernel/Modules/AdminSMIME.pm
@@ -15,8 +15,6 @@ use Kernel::Language qw(Translatable);
 
 our $ObjectManagerDisabled = 1;
 
-use Kernel::Language qw(Translatable);
-
 sub new {
     my ( $Type, %Param ) = @_;
 
@@ -30,7 +28,7 @@ sub new {
 sub Run {
     my ( $Self, %Param ) = @_;
 
-    # get needed objects
+    # get objects
     my $ParamObject  = $Kernel::OM->Get('Kernel::System::Web::Request');
     my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
 
@@ -43,6 +41,7 @@ sub Run {
         $Output .= $LayoutObject->NavigationBar();
 
         $LayoutObject->Block( Name => 'Overview' );
+        $LayoutObject->Block( Name => 'Notice' );
         $LayoutObject->Block( Name => 'Disabled' );
         $LayoutObject->Block( Name => 'OverviewResult' );
         $LayoutObject->Block(
@@ -71,6 +70,7 @@ sub Run {
         );
 
         $LayoutObject->Block( Name => 'Overview' );
+        $LayoutObject->Block( Name => 'Notice' );
         $LayoutObject->Block( Name => 'NotWorking' );
         $LayoutObject->Block( Name => 'OverviewResult' );
         $LayoutObject->Block(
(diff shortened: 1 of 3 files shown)


carlosfrodriguez:
Hi @s7design,

Thank you for pull request, I see that in the TT file there is some code duplication Where only a string is different why not set an TT if condition to either display the disabled or not working message within the same TT block? does it make sense?

((enjoy))
cr

s7design:
Hi @carlosfrodriguez 

Thanks for sugestion, I updated PR. I hope it is better now.
There is only part of code which is duplicated,
`
 [% Translate("Use this feature if you want to work with SMIME keys.") | html %]`

but I think it is OK. We can if you like put this message below the button, and there will not be duplicated code.

```
[% RenderBlockStart("Disabled") %]
                    <h3><span class="Warning">[% Translate("This feature is disabled!") | html %]<span></h3>
                    <fieldset>
                        <div class="Field SpacingTop">
                            <button class="CallForAction Fullsize Center" type="submit" value="[% Translate("Enable it here!") | html %]">
                                <span><i class="fa fa-unlock-alt"></i> [% Translate("Enable it here!") | html %]</span>
                            </button>
                        </div>
[% RenderBlockEnd("Disabled") %]
[% RenderBlockStart("NotWorking") %]
                    <h3><span class="Error">[% Translate("This feature is not working!") | html %]<span></h3>
                    <fieldset>
                        <div class="Field SpacingTop">
                            <button class="CallForAction Fullsize Center" type="submit" value="[% Translate("Configure it here!") | html %]">
                                <span><i class="fa fa-unlock-alt"></i> [% Translate("Configure it here!") | html %]</span>
                            </button>
                        </div>
[% RenderBlockEnd("NotWorking") %]
                        <p class="FieldExplanation">
                            [% Translate("Use this feature if you want to work with SMIME keys.") | html %]
                        </p>
                        <div class="Clear"></div>
                    </fieldset>
```

Regards
Zoran

mrcbnsls:
This PR had been merged in the meantime, but I had to revert it due to a failing UT (AdminSMIME.t). Could you please investigate on this? I think this is again a timing problem (maybe in connection with mod_perl) when you use $SysConfigObject->ConfigItemUpdate(). Please make sure to wait for some seconds before refreshing the screen using VerifiedRefresh().

Also, I could see that the notice box is always there but empty if no message is to be displayed (see screenshot).

s7design:
Hi @mrcbnsls 

I will try to fix it. I updated this about Notice in PR https://github.com/OTRS/otrs/pull/1057

Regards
Zoran

s7design:
Hi @mrcbnsls 

I updated PR, I hope it works better now.
Please let me now how it works in you system. I tested and it works for me.

Regards
Zoran

mrcbnsls:
Ok, lets give it another try :D 

Merged in 6885d8039921f5e99f9f0d5dd94877ee7af58f2e and 3109095. Thanks!

PR #942, opened 2016-01-28
[ Rel-5_0 Bug 11834 ] - Ticket Title is not correct in a csv/excel output

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11834

Hello @mgruner ,

Hereby is proposal for fixing this bug and additional unification of the system output on search ticket. 

In config replaced Subject with Title for CSV/Excel output for agent and customer search. Added Translatable for customer headers in table.

There is similar issue in PDF output, changed that. 

CustomerTicketSearch normal output, changed to show Title as well and removed 'Body' from output, since IMO if we show ticket title there is no need to show body of first article.

In AdminGenericAgent, changed in ticket result table to show ticket title instead of article subject.

I am aware that these changes are not critical to the system, but i believe that they will provide nice user experience when outputting ticket search and reading data. 
Please let me know if you disagree with these changes or if there are any questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -2408,7 +2408,7 @@
                 <Item Translatable="1">CustomerID</Item>
                 <Item Translatable="1">CustomerName</Item>
                 <Item Translatable="1">From</Item>
-                <Item Translatable="1">Subject</Item>
+                <Item Translatable="1">Title</Item>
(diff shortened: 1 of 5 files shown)


zottto:
Hi @s7design,

I closed this bug as invalid. The underlying problem is quite different to your fix proposal. Please see my comment on Bugzilla for details.

Marc

PR #944, opened 2016-01-28
Rel-5_0 - OTRS - updated unit test - part 1

s7design:
Hi @mgruner 

This is the first PR with updated unit test. You can see I added helper object with  RestoreSystemConfiguration and RestoreDatabase where it was necessary. I also change rand function with Helper GetRandomID() methods. 

Of course there were some else changes. As you know there would be many possible things which we could change but I am not sure how wide we should go. If you have any suggestion during this process updating unit test please let me know and we will be able to change it.

In AuthSession I add RestoreDatabase becouse in the test all test are deleted and if there was active session it will be deleted. At the end with transaction in the test the previous state will be restored.

Regards
Zoran

### DIFF
--- a/scripts/test/Auth.t
+++ b/scripts/test/Auth.t
@@ -12,7 +12,16 @@ use utf8;
 
 use vars (qw($Self));
 
-# get needed objects
+# get helper object
+$Kernel::OM->ObjectParamAdd(
+    'Kernel::System::UnitTest::Helper' => {
+        RestoreSystemConfiguration => 1,
+        RestoreDatabase            => 1,
+        }
+);
+my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
+
+# get config object
 my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
 # configure auth backend to db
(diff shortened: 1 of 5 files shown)


mgruner:
Merged, very nice, thanks. Actually I want to ask you to also remove all rand() calls from the unit test files as you go through them and replace them with GetRandomID() or GetRandomNumber(), a new method that I just added in case only a number is possible.

s7design:
Thanks @mgruner . I will do that. I think that it is good now that there are two rand function for different cases.

In that case, I think that we can change for example merged test from this PR Cache.t

```
            # create key
            ----#my $Key = $Helper->GetRandomID();
            ++++#my $Key = $Helper->GetRandomNumber();

            # copy strings to safe the reference
            my $StringRef1 = $String1;
            my $StringRef2 = $String2;

```

If you like I can do that for Cache.t test, and of course we will take care about that in the next tests which we do.

Regards
Zoran

mgruner:
Sure. Thanks.

Martin Gruner
Team Lead R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

PR #945, opened 2016-01-29
Bug 8298: Use of "Submit" and "Save" buttons

nielsdimmers:
This pull request contains the following changes (All are textual, no functionality has been changed):

In general, settings are saved, so on almost every admin screen, the “Sumbit” button has been changed to “Save”, with a few exceptions:
- otrs/index.pl?Action=AdminSMIME;Subaction=ShowAddCertificate - I changed the “Submit” button to “Add”, because the user is adding a certificate here
- otrs/index.pl?Action=AdminSysConfig;Subaction=Import - I changed the “Submit” button to “Import”, since the user is importing settings here

Mostly, agent actions are “submitted”, so the submit button can stay whenever there’s a creation of a ticket, adding of an article or sending of a message. Everywhere where there are settings involved (like the settings on the dashboard or in the ticket overview screen) I changed “submit” to “Save” or to “Save Changes” if the last one was more in line with similar situations. There are two exceptions:
- otrs/index.pl?Action=AgentTicketBulk - I changed the “Submit” button to “Excecute Bulk Action”, because the user wants to execute the bulk action
- otrs/index.pl?Action=AgentTicketMove - I changed the “Submit” button to “Move”, since the ticket is moved to a different queue

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminAttachment.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminAttachment.tt
@@ -155,7 +155,7 @@
                         <div class="Clear"></div>
 
                         <div class="Field SpacingTop">
-                            <button class="Primary CallForAction" type="submit" value="[% Translate("Submit") | html %]"><span>[% Translate("Submit") | html %]</span></button>
+                            <button class="Primary CallForAction" type="submit" value="[% Translate("Save") | html %]"><span>[% Translate("Save") | html %]</span></button>
                             [% Translate("or") | html %]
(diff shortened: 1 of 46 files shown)


mgruner:
Hello @nielsdimmers, impressive, thank you!
@mrcbnsls, can you please review this? I believe you had a similar item on your list. We should process this timely to avoid merge conflicts.

mrcbnsls:
Nice, merged in b03f43b1f4be8359f1e2e84367d3626a93b4f75e. Thanks a lot!

PR #946, opened 2016-01-29
[ Rel-5_0 ] - menu moduls update

s7design:
Hi @mgruner 

Working on some bug I saw that Spam menu item changed and now it moves Ticket to Junk queue. It have been done in this commit https://github.com/OTRS/otrs/commit/8ce6b493b63456fc1e3d71a97d708ecb06cbba6d

I saw that it was not changed in the TicketMenu/Menu.t Selenium test. I wanted to update this test and then I saw there are some error in Fred, and because of that I also fix code in /TicketMenu/Generic.pm. There was a problem if there is not $Param{Config}->{Action} and then $Config will be empty.
`
        my $Config = $ConfigObject->Get("Ticket::Frontend::$Param{Config}->{Action}");`

There was also a problem with  <li id="nav-[% Data.Name %]"> in AgentTicketZoom.tt In some case id was with space for example "nav-Phone Call Outbound".

Title attribute  for cluster menu item is not neccecccary, acctually it was empty ( titlt="") 
It is not needed because on hover the submenu is expanded. Simmilar to empty title, the class for ul of cluster container is not needed, because there is not value, it was also empty.
I used idea from AgentNavigationBar.tt to make this two screen consistent, since I add NameForID.

And the last thing that I changed Ticket::Frontend::PreMenuModule###460-Spam, I think it could be confusing that Spam action in Zoom and Overview screens  have different behavior one move ticket in Junk and the other in Delete queue.

I did not open a bug for that, but if you would like I will be able to do that. If you think this changes is necessary you can marge it. Please let me now what do you think about all of that.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -1002,7 +1002,7 @@ sub MaskAgentZoom {
             }
 
             $Item->{NameForID} = $Item->{Name};
-            $Item->{NameForID} =~ s/ /-/g if( $Item->{NameForID} =~ m/ / );
+            $Item->{NameForID} =~ s/ /-/g if ( $Item->{NameForID} =~ m/ / );
 
             if ( !$Menus{$Menu}->{ClusterName} ) {
 
@@ -1024,15 +1024,15 @@ sub MaskAgentZoom {
         for my $Cluster ( sort keys %MenuClusters ) {
 
             my $NameForID = $Cluster;
-            $NameForID =~ s/ /-/g if( $NameForID =~ m/ / );
+            $NameForID =~ s/ /-/g if ( $NameForID =~ m/ / );
 
             $ZoomMenuItems{ $MenuClusters{$Cluster}->{Priority} . $Cluster } = {
-                Name  => $Cluster,
+                Name      => $Cluster,
                 NameForID => $NameForID,
-                Type  => 'Cluster',
-                Link  => '#',
-                Class => 'ClusterLink',
-                Items => $MenuClusters{$Cluster}->{Items},
+                Type      => 'Cluster',
+                Link      => '#',
+                Class     => 'ClusterLink',
+                Items     => $MenuClusters{$Cluster}->{Items},
                 }
         }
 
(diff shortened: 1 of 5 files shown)


Kernel/Modules/AgentTicketZoom.pm
```diff
@@ -1001,6 +1001,9 @@ sub MaskAgentZoom {
                 $Item->{Class} = "AsPopup PopupType_$Menus{$Menu}->{PopupType}";
             }
 
+            $Item->{NameForID} = $Item->{Name};
```
mgruner: I believe this is not needed in Perl. You can do this in TT directly.

Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt
```diff
@@ -373,16 +373,16 @@
                     <div class="ActionRow Cluster">
                         <ul class="Actions">
 [% RenderBlockStart("TicketMenu") %]
-                            <li id="nav-[% Data.Name %]">
+                            <li id="nav-[% Data.NameForID %]">
```
mgruner: You can instead use something like [% Data.Name  | replace(' ', '-') | html %]

scripts/test/Selenium/Output/Ticket/Menu.t
```diff
@@ -197,92 +213,146 @@ $Selenium->RunTest(
         # create menu module test params
         my @MenuModule = (
             {
-                Name   => "Back",
-                Action => "AgentDashboard",
+                Name      => "Back",
+                NameForID => "Back",
```
mgruner: I think this could also be computed in the loop below, to save some lines.

mgruner:
Please see my inline comments. Looks like a good improvement in general.

Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt
s7design: Thanks Martin, I tried something similar. And my first intention was that I solve it in TT, but I see now that I made mistake. I tried something like  [% Data.Name.replace(' ', '-') | html %] and alll menu items were duplicated, becouse of that I suggested solution where I replaced space in Perl module.

Thanks, I will try this solution. I believe it will work.

Regards
Zoran

s7design:
Thanks for follow-up, I will check all of that and update PR soon.

Regards
Zoran

s7design:
This PR is updated, I believe it is much better now.

Regards
Zoran

mgruner:
Merged, thank you Zoran!

PR #948, opened 2016-02-01
Rel-5_0 - OTRS - updated unit test - part 3

s7design:
Hi @mgruner 

Here is one more PR with updated unit tests. I updated the previus ones where omitted come in list of the helper param. 
You can see CustomerUserService.t, I think that clean-up code which was at the end of the test is not necessary any more because clean-up is done by RestoreDatabase.

Regards
Zoran


mgruner:
Merged, thanks a lot!

PR #949, opened 2016-02-01
Upload cache performance fix

pboguslawski:
OTRS upload cache cleaning up is done on every FormID creation which may
hurt performance, especially if there are many cached files on slower
storage (i.e. shared NFS directory).

This mod moves cleaning upload cache to separate scheduler task fired
every hour.

When WebUploadCacheModule=FS is used, files are also spread between
different subdirectories to avoid many files in one directory (for
better performance and easier cleaning up).

Related: https://dev.ib.pl/ib/otrs/issues/20
Author-Change-Id: IB#1048779

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-27 Upload cache cleanup command name changed, unit test added.
  - 2016-02-18 Upload cache performance fix.
  - 2016-02-01 Upload cache performance fix.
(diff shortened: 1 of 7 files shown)


mgruner:
Looks very nice, Pawel! Can you please rename the command to Maint::WebUploadCache::Cleanup and add a unit test for it as well?

pboguslawski:
Command renamed, unit test added.

Regards,
Pawel

mgruner:
Thanks a lot @pboguslawski. Excellent work, also the test cases. Good job! Merged in 4af77eeddb1668dec197497dcec0cb7881ae7331.

PR #951, opened 2016-02-02
Rel-5_0 - OTRS - updated unit test - part 4

s7design:
Hi @mgruner 

This is the next part of the updated unit tests. You will see I also updated CustomerGroup, there were some spell mistakes.

In some test I made Helper object with RestoreDatabase, but actually I would like only to clean up cache. Maybe we can update Helper.pm and add RestoreCache for such cases (e.g Event.t - there were Valid and DynamicField in cache after running this test ). It is not big problem but maybe we can leave full clean system after any unit test.

Regards
Zoran

### DIFF
--- a/scripts/test/CustomerGroup.t
+++ b/scripts/test/CustomerGroup.t
@@ -273,7 +273,7 @@ my $ResetMembership = sub {
     }
 };
 
-# reset memberchip
+# reset membership
 $ResetMembership->(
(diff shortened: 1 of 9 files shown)


mgruner:
Hi Zoran,
I believe it is ok to use RestoreDatabase in that case, it will not hurt.
Best regards, Martin

mgruner:
Merged, very nice, thanks!

s7design:
Thanks, I agree with you. I wanted only to inform you. 

Regards
Zoran

PR #952, opened 2016-02-02
Removed GetTableDate from frontend modules - SystemAddressQueueList

s7design:
Hi @mgruner 

This is PR with removed GetTableData from frontend modules. 

Working on SystemAddress module I saw that there is maybe one problem. It is possible to add system addresses with the same Email address, Display name and Queue ( test@localhost, test, Junk - you can add such identical system addresses as many as you want  ). Is it OK, or we should add some protection in SystemAddress module.

There are not any more this deprecated method in frontend said.
There is only in the following modules in the backend:

<pre>
/opt/otrs/Kernel/System/Signature.pm:
  230  
  231      # sql
  232:     return $Kernel::OM->Get('Kernel::System::DB')->GetTableData(
  233          What  => 'id, name',
  234          Valid => $Valid,

/opt/otrs/Kernel/System/StdAttachment.pm:
  454  
  455      # return data
  456:     my %Relation = $DBObject->GetTableData(
  457          Table => 'standard_template_attachment',
  458          What  => 'standard_attachment_id, standard_template_id',
  ...
  493  
  494      # return data
  495:     return $DBObject->GetTableData(
  496          Table => 'standard_attachment',
  497          What  => 'id, name, filename',

/opt/otrs/Kernel/System/CustomerUser/DB.pm:
  378  
  379      # get data
  380:     my %Users = $Self->{DBObject}->GetTableData(
  381          What  => "$Self->{CustomerKey}, $Self->{CustomerKey}, $Self->{CustomerID}",
  382          Table => $Self->{CustomerTable},
</pre>
I am planning fix it on these places and we will be able to remove GetTeblaData function. I will do that in another PR.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketEmail.pm
+++ b/Kernel/Modules/AgentTicketEmail.pm
@@ -2051,12 +2051,7 @@ sub _GetTos {
             );
         }
         else {
-            %Tos = $Kernel::OM->Get('Kernel::System::DB')->GetTableData(
-                Table => 'system_address',
-                What  => 'queue_id, id',
-                Valid => 1,
-                Clamp => 1,
-            );
+            %Tos = $Kernel::OM->Get('Kernel::System::SystemAddress')->SystemAddressQueueList();
         }
 
         # get create permission queues
--- a/Kernel/Modules/AgentTicketPhone.pm
+++ b/Kernel/Modules/AgentTicketPhone.pm
@@ -2167,12 +2167,7 @@ sub _GetTos {
             );
         }
         else {
-            %Tos = $Kernel::OM->Get('Kernel::System::DB')->GetTableData(
-                Table => 'system_address',
-                What  => 'queue_id, id',
-                Valid => 1,
-                Clamp => 1,
-            );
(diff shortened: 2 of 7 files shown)


mgruner:
Merged, thank you! I don't know about forcing this to be unique. 

@nilei: do you think it would be good to add some protection against duplicate system addresses? What exactly should be unique, the email address maybe?

s7design:
To be honest Martin, I am also not sure about that. I don't have experience with real installation  OTRS in production system, but it looks me that it could make some confusion. Now there is possible to have one email address with different display name, or the same display name for two or more email addresses.

We can thinks about that. If we can help about improving this behavior feel free to tell me and we will be able to do that.
Some of possible situation are displayed in the picture below:

screen shot 2016-02-04 at 10 33 40 am

Regards
Zoran

nilei:
Hi @mgruner + @s7design,

I agree with Martin, email addresses should be unique as otherwise we cannot control to which queue an incoming email would be routed to.

Alternatively we could add some priority to the data sets, but I think this will only cause confusion and no real benefit.

For upgrading I would recommend to not change the existing data but change the logic to stop on first match and add a Support Data Collector module to check duplicates.

Cheers, Nils

PR #656, opened 2015-07-29
Modularize AgentTicketZoom

moritz:
Start to modularize AgentTicketZoom by putting the "Ticket Information" sidebar widget into a separate module, and a generic module loader for more sidebar widgets.

If this meets your approval, I'll also factor out the "Customer Information" and the "Linked Objects" widgets into separate modules.

PR #953, opened 2016-02-02
Modularize AgentTicketZoom, PR 2.0

moritz:
This is basically the same as #656, but based on the current master; this was easier than trying to rebase or merge. It also changes the location of the modules based on feedback by @mgruner.

It moves all widgets except the article view into separate modules, and at the same time introduces the option to extend AgentTicketZoom with custom widgets in a clean way.


mgruner:
Merged with one slight change: I also put the templates in a subdirectory ```AgentTicketZoom/*.tt```. Thanks a lot @moritz!

s7design:
Hi all,

Nice for that improvement, I think it is much better with @moritz improvement. 

We have mentioned before that there should update some Selenium tests. I suppose we can do that now.

Reagrds
Zoran

mgruner:
Hello @s7design, thanks for reminding me. Yes, I would be grateful if you could enhance AgentTicketZoom.t to test with toggling of the widgets (for LinkTable also Sidebar vs. Main positioning) and also a some widget content testing.

s7design:
OK thanks, it will be done the best we can. I have tested this features manually before applying  Moritz's changes and i have some ideas which test case we can include in the tests.

Regards
Zoran

PR #955, opened 2016-02-04
[ Rel-5_0 Bug 11422 ] - List in the Dashboard: standard priority is ignored if you use a filter

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11422

Hello @mgruner ,

Hereby is proposal for fixing issues in Dashboard screen with auto-sorting columns on filter actions.

Added new param on Filter actions 'AddFilters'. Based on this param we can distinguish column click ( for ordering ) and filter actions. Before they acted as same event which was resulting in changing column order on filter action.

Added Selenium test in which we test reporter case and different sorting issues that were present in this module.

Additionally added Translatable() for strings that were not in the system and Translate() for column titles.

This proposal is for rel-5_0, there are some changes compared to master. If you need we can provide different PR for master with this fix proposal.

Please let me know if there are any questions or issues regarding this PR.

Kind regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
@@ -40,6 +40,9 @@ sub new {
         $Self->{$Item} = $ParamObject->GetParam( Param => $Item ) || $Param{$Item};
     }
 
+    # get add filters param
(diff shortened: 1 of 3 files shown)


s7design:
Hello @mrcbnsls ,

This PR is rebased and Selenium test is updated for stability.

Regards,
Sanjin

mgruner:
Superseded by #1598, closing.

PR #956, opened 2016-02-04
Removed deprecated method GetTableDate

s7design:
Hi @mgruner 

In this PR I removed GetTableDate method. There have been done the following steps:

* There are redefined two methods: SignatureList and StdAttachmentList
* added unit tests for this methods
* removed deprecated methods: StdAttachmentSetResponses, CustomerUser/DB/CustomerUserList() and CustomerUser/CustomerUserList() and of course GetTableData itself. 
* fixed some spelling mistake in the tests on which I had done recently ( 'does not contains' instead 'does not contain') and 
* fixed illogicality such checking if nonexistent item is equal with something
<pre>
$Self->False(
    exists $SystemQueues{'1'} && $SystemQueues{'1'} == $SystemAddressID,
    "SystemAddressQueueList() does not contains the queue 1 of the SystemAddress $SystemAddressID",
);
</pre> 

Please let me know, what do you think about all of that.

Regards
Zoran

### DIFF
--- a/Kernel/System/CustomerUser.pm
+++ b/Kernel/System/CustomerUser.pm
@@ -180,32 +180,6 @@ sub CustomerSearch {
     return %Data;
 }
 
-=item CustomerUserList()
-
-return a hash with all users (depreciated)
-
-    my %List = $CustomerUserObject->CustomerUserList(
-        Valid => 1, # not required
-    );
-
-=cut
-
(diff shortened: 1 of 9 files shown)


mgruner:
Merged, thank you very much! I also removed CustomerUserList from LDAP.pm.

PR #957, opened 2016-02-04
Rel-5_0 - OTRS - updated unit test - part 6

s7design:
Hi @mgruner 

There is one more PR with updated unit test where are added Helper object with params for cleaning-up system after running tests. I also fixed some other test code, I hope I improved it but if you have any suggestion about that please let me know.

Regards
Zoran

### DIFF
--- a/scripts/test/Loader.t
+++ b/scripts/test/Loader.t
@@ -180,6 +180,6 @@ for my $Test (@JSTests) {
     );
 }
 
-# cleanup is done by RestoreDatabase
+# cleanup cache is done by RestoreDatabase
 
 1;
--- a/scripts/test/Loader.t
+++ b/scripts/test/Loader.t
@@ -14,21 +14,28 @@ use vars (qw($Self));
 
 # get needed objects
 my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
-my $HelperObject = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
 my $LoaderObject = $Kernel::OM->Get('Kernel::System::Loader');
 my $MainObject   = $Kernel::OM->Get('Kernel::System::Main');
 
+# get helper object
+$Kernel::OM->ObjectParamAdd(
+    'Kernel::System::UnitTest::Helper' => {
+        RestoreDatabase => 1,
+    },
+);
+my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
+
+my $Home = $ConfigObject->Get('Home');
+
 {
     my $CSS = $MainObject->FileRead(
-        Location => $ConfigObject->Get('Home')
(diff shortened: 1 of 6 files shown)


scripts/test/ReferenceData.t
```diff
@@ -14,9 +14,17 @@ use vars (qw($Self));
 
 use Kernel::System::VariableCheck qw(:all);
 
-# get needed objects
+# get config object
 my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
+# get helper object
+$Kernel::OM->ObjectParamAdd(
```
mgruner: This seems unneeded as there are no SysConfig calls here, only ConfigObject that changes only runtime?

scripts/test/MailAccount.t
```diff
@@ -12,9 +12,7 @@ use utf8;
 
 use vars (qw($Self));
 
-use Kernel::System::ObjectManager;
-
-# get needed objects
+# get mail account object
```
mgruner: No RestoreDatabase here?

scripts/test/Loader.t
```diff
@@ -14,21 +14,28 @@ use vars (qw($Self));
...
 # get needed objects
 my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
-my $HelperObject = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
 my $LoaderObject = $Kernel::OM->Get('Kernel::System::Loader');
 my $MainObject   = $Kernel::OM->Get('Kernel::System::Main');
 
+# get helper object
+$Kernel::OM->ObjectParamAdd(
+    'Kernel::System::UnitTest::Helper' => {
+        RestoreDatabase => 1,
```
mgruner: Do we really need this here? I believe the Loader does not change the DB.

mgruner:
Hello Zoran, please see my inline comments.

scripts/test/MailAccount.t
s7design: It is actually not need here, there is MailAccountDelete at the end of the test.

scripts/test/ReferenceData.t
s7design: Thanks, I will fix it, I added RestoreDtaabase, but I did not think that it is only ConfigObject.

scripts/test/Loader.t
s7design: I addded it only to clear cache, I will change comment at the end in:
# cleanup cache is done by RestoreDatabase

s7design:
The PR is updated. Please let me know if I can fix anything else.

Regards
Zoran

mgruner:
Merged, thanks. I still added RestoreDatabase to MailAccount.t as it avoids write operations in the DB and we can really make sure there are no leftovers in any case, also for future additions to the test.

PR #961, opened 2016-02-05
Rel 5 0 ut restore database part7

s7design:
Hi @mgruner 

One more PR with updated unit tests. I implemented some your sugestion from the preius commit about clean-up cache data and RestoreDatabase in PGP.t even though there is PublicKeyDelete() in test

Regards
Zoran

### DIFF
--- a/scripts/test/PID.t
+++ b/scripts/test/PID.t
@@ -14,7 +14,6 @@ use vars (qw($Self));
 
 # get needed objects
 my $HelperObject = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
-my $DBObject     = $Kernel::OM->Get('Kernel::System::DB');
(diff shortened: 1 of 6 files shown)


scripts/test/PID.t
```diff
@@ -14,7 +14,6 @@ use vars (qw($Self));
 
 # get needed objects
 my $HelperObject = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
```
mgruner: No RestoreDatabase here?

s7design: Yes we could add here, the same reason even though there is PIDDelete. I will update it right now.

s7design:
I updated PR

Regards
Zoran

mgruner:
Merged, thanks!

PR #972, opened 2016-02-06
Add support for X-OTRS-Title in PostMaster/NewTicket

reneeb:
This can be helpful when a postmaster filter wants to change/set the ticket title based on some calculations/checks...

### DIFF
--- a/Kernel/System/PostMaster/NewTicket.pm
+++ b/Kernel/System/PostMaster/NewTicket.pm
@@ -228,7 +228,7 @@ sub Run {
     my $NewTn    = $TicketObject->TicketCreateNumber();
     my $TicketID = $TicketObject->TicketCreate(
         TN           => $NewTn,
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @reneeb. Thanks for the suggestion! It seems however, that `X-OTRS-Title` is not supported anywhere else (for example in FollowUp.pm) so this would be an inconsistent addition. Additionally it would be good to cover this with unit tests as well. What do you think?

reneeb:
If you're interested, I would add the Header in other packages as well and add unittests

mgruner:
@reneeb please do so. If we have a consistent state we can merge this.

mgruner:
@reneeb, when do you think you will be able to work on this?

reneeb:
next week...

reneeb:
Improved version is available: #1127

PR #974, opened 2016-02-08
Fixed output for Age data in Ticket Information sidebar widget

s7design:
Hi @mgruner 

Working on the Selenium tests for new sidebar widgets we saw that there is small mistake with output Age data in Ticket Information widget.

You can see in the picture:

screen shot 2016-02-08 at 11 43 15 am

Here, it is fixed and now output for Age data is again human readable from CustomerAge method.

P.S. If you would like you could marge it and we will be able to include it in the selenium tests as well. 

Regards
Zoran


mgruner:
Thanks!

PR #976, opened 2016-02-09
Rel-5_0 - OTRS - updated unit test - part 8

s7design:
Hi @mgruner 

One more part of updated Selenium tests

Regards
Zoran


mgruner:
Merged with one fix for Service.t (use GetRandomID in the loop to get different IDs for the different services). Thanks!

PR #979, opened 2016-02-09
[ Rel-5_0 Bug 11854 ] - Error with the Spam

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11854

Hello @mgruner ,

Here's our proposal for fixing this bug. Added so that ACL restriction on 'Spam' and 'Delete' menu module in AgentTicketZoom show error that is used for ACL restrictions. 

If you think it's needed i can change it so that ACL remove also menu modules button in AgentTicketZoom if match is satisfied and leave this type of error to show when restricted queue added directly in URL.

Modified AgentTicketMove.t Selenium tests to simulate this type of behaviour.

Kind Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketMove.pm
+++ b/Kernel/Modules/AgentTicketMove.pm
@@ -239,6 +239,15 @@ sub Run {
         $Error{DestQueue} = 1;
     }
 
+    # check if destination queue is restricted by ACL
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Merged in f787c6edefff10530d7cdd070ea80201b2504474, thanks.

PR #982, opened 2016-02-09
Add new feature: define default headers for emails

reneeb:
Sometimes someone might want to send some headers in each email sent via OTRS. Currently there is no way to define such headers. With this patch it would be possible...

Many mail servers do not send "out of office" replys for mails where those headers in the sample config are set. Some of my customers do not want any "out of office" replys at all, so they want to set the email headers.

Another use case is when OTRS "talks" to another system and you want to send a static X-Header to identify the system.


reneeb:
replaces PR #761

mgruner:
Merged in abe33d6, thanks a lot @reneeb! I renamed the config setting to Sendmail::DefaultHeaders to make it more consistent with the other settings.

PR #983, opened 2016-02-10
Rel-5_0 - OTRS - updated unit test - part 10

s7design:
Hi @mgruner 

There is one more PR with updated unit tests.
I had a problem to use Helper object and RestoreDatabase in SupportDataCollector.t test. I added simple CleanUp() method.

Regards
Zoran


mgruner:
Merged, thank you very much!

PR #984, opened 2016-02-10
Enable asynchronous loading of AgentTicketZoom widgets

moritz:
This patch enables automatic asynchronous loading of the new AgentTicketZoom sidebar widgets.

This is not huge benefit for the Customer Information widget, for which I've enabled it now, but is very much appreciated for extensions that either have to talk to external APIs, or have to do potentially slow things (like list all the attachments in a ticket with several thousand articles).

It mostly works, just the triangles for toggling the visibility of the widget don't work. That's because the content of the widget is not available at the time the toggle functionality is attached. So we'd need a way to run the Core.UI init sequence on just the new DOM elements. And that's totally out of my depth, so feedback (or even a patch) from a frontend expert would be very much appreciated.

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -14,6 +14,7 @@ use warnings;
 our $ObjectManagerDisabled = 1;
 
 use POSIX qw/ceil/;
+use Digest::SHA1 ();
 use Kernel::System::EmailParser;
 use Kernel::System::VariableCheck qw(:all);
 use Kernel::Language qw(Translatable);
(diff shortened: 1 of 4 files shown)


moritz:
I've found a solution to the Javascript problem (see commit 51de002), so I'd consider this mergable. Feedback very welcome!

moritz:
Hi, any updates or feedback on this pull request?

moritz:
Hi @mrcbnsls, @mgruner,

any updates? I'd have to have this pull request rots until there are conflicts to resolve...

Cheers,
Moritz

mrcbnsls:
Hey Moritz,

A big sorry for letting you wait for so long. First of all, thank you for all the effort you've made to bring this into OTRS!

We've been discussing this internally and came to no real conclusion. Though we understand the approach and could imagine use cases for it, we dislike the idea of having this in the framework at the moment. Currently, there is no widget in ticket zoom which would be predestinated for asynchronous loading. It's nice that you've enabled it for the customer information widget, but its more like a proof that it works and no real improvement to the current behavior. Users will probably more likely notice the delay of having the customer information displayed rather than noticing a reduced loading time of AgentTicketZoom.

Our current thought is to not merge this PR. If you'd like to have asynchronous loading for a certain widget, the modularized approach you've added with a recent PR would probably allow for implementing it directly into the widget itself, without touching the framework code.

I'll close this for now. Please feel free to share your thoughts about it. Thanks!

Marc

jensoliver:
Hi Marc,

can we discuss this tomorrow as it might be helpful for one of the requests for OTRS6?

mrcbnsls:
Hi @jensoliver,

Of course we can discuss it. I will not be available tomorrow, but please feel free to add your feedback to this PR or to write me an email :)

Marc

moritz:
Hi @mrcbnsls,

while I agree that the core framework itself has no real use case for it, there are certainly extensions that could greatly benefit from it. For example http://opar.perl-services.de/dist/TicketAttachments works nicely, but can introduce a delay of several seconds for large tickets if loaded synchronously.

Likewise we some proprietary extensions inhouse that query data from external sources that might be slow and/or unreliable.

Sure, we can implement asynchronous loading for each of these widgets separately, but that leads to quite some code duplication (of which OTRS has enough as-is, if you ask me), and puts the burden on those that want to extend the platform.

So the question, from my perspective, is whether OTRS wants to position itself as an open platform, where such a feature makes quite some sense, or as an application, where you reject such a feature because it's not immediately beneficial.

So I'd appreciated it if you could reconsider your decision to reject this.

Cheers,
Moritz

mgruner:
Hello @moritz. I'm not sure if this is fully fair; the fact that we merged the modularization of AgentTicketZoom showed the platform aspect of OTRS already, don't you think? What we want to avoid is having code in the Framework that is not really used and tested, but still part of the API. I have already experienced a lot of crazy things like "this worked once and is now gone" - even if "this" was never really an official feature. @mrcbnsls said that this was a close decision and we are open for arguments, as always I believe.

moritz:
Hi @mgruner,

you're right, the merge of the modularization of AgentTicketZoom shows that you are open to extension APIs. The question is just how much. And I do remember dicussions with you and @ManuelHecht around that topic, and your response was along the lines of "if you need an extension API, submit a pull request for it", and it did not include a clause "but only if it's also beneficial for the OTRS core itself".

So, I guess the question is, can I do something to alleviate your fears that this API becomes unused and/or untested? Would a port of a publicly available extension to use that API be enough?

I haven't run them myself, but my understanding is that the Selenium tests should already cover it. If not, I can try to improve coverage in this area.

Cheers,
Moritz

Kernel/Modules/AgentTicketZoom.pm
```diff
@@ -14,6 +14,7 @@ use warnings;
 our $ObjectManagerDisabled = 1;
 
 use POSIX qw/ceil/;
+use Digest::SHA1 ();
```
mgruner: This is untypical. Why do you need it, to escape IDs for JavaScript? That could be done in a less expensive way, I guess.

Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt
```diff
@@ -144,8 +144,20 @@
 
     <div class="LayoutFixedSidebar SidebarLast">
         <div class="SidebarColumn">
-            [% FOREACH widget IN Data.SidebarWidgets; %]
-                [% widget %]
+            [% FOREACH Widget IN Data.SidebarWidgets; %]
+                [% IF Widget.Async; %]
+                    <div id="[% Widget.ElementID %]"></div>
+                    [% WRAPPER JSOnDocumentComplete %]
```
mgruner: This would have to be rewritten to comply with the JavaScript code in master. Almost all JS was moved out of the template files and into JS namespaces. It would probably be useful to add method to Core.Agent.TicketZoom and use $Layout->AddJSData() to inform it about widgets to be loaded.

var/httpd/htdocs/js/Core.AJAX.js
```diff
@@ -481,10 +481,11 @@ Core.AJAX = (function (TargetNS) {
      * @param {jQueryObject} $ElementToUpdate - The jQuery object of the element(s) which should be updated
      * @param {String} URL - The URL which is called via Ajax
      * @param {Function} Callback - The additional callback function which is called after the request returned from the server
+     * @param {Bool} Replace - replace the DOM container with the response instead of updating the contents
```
mgruner: We would like to avoid this addition. Maybe it could be solved with container elements.

moritz:
@mgruner I'll look into updating this PR, though probably not this week.

mgruner:
Closing, please see the follow-up PR.

PR #985, opened 2016-02-11
[ Rel-5_0 Bug 11791 ] - Can't use FullTextSearch with Russian words

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11791

Hello @mgruner ,

Hereby is our proposal for fixing this issue. Changed 'Ticket::SearchIndex::Filters' to look for WorldCharacters instead of a-z A-Z 0-9. Now Cyrillic characters that are 2 bytes and Chinese characters that are 3 bytes, when encoded, are not removed by filter. They are added in the article_search DB on ArticleCreate/ArticleUpdate events and can be used by FullText TicketSearch.

Added also small improvement in apply filter 'for loop;. Where we escape it if string is empty as a result of previous filter action.

Modified ArticleSearchIndex.t unit test to simulate reporter issue and created some additional cases that can help in future testings. 

Kind Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -526,7 +526,7 @@
             <Array>
                 <Item><![CDATA[[,\&\<\>\?"\!\*\|;\[\]\(\)\+\$\^=]]]></Item>
                 <Item><![CDATA[^[':.]|[':.]$]]></Item>
-                <Item><![CDATA[^[^a-zA-Z0-9]+$]]></Item>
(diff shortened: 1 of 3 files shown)


mgruner:
Manuel agreed on this.

dvuckovic:
Merged in e71bcdacb43998ae84ce49fd23bf2621d6be7602 and 1bc116d30111afd75d4a4b91df628e2dba03537c. Thanks for the fix!

PR #986, opened 2016-02-11
[Master - Bug 11862] - Ticket Notification Management - 'Send by default' has no impact

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11862

Hi @mgruner 

As you can see on the bug report, there was not so clear what meant "Send by default". I changed it a little bit, if you would like some other filed explanation, please let me know and I will be able to change it.

In any case this is not a bug. This field is used to activate by default ticket notification when new agent is created. It forks fine for me. There is not influence on created agents if you change this field, it will affected only on new ones.

Regards
Zoran


mgruner:
Thanks! There is now an alternative proposal in #995.

PR #987, opened 2016-02-12
Pass pref

MBRG:
Added some instructionary comments regarding password policy options and modified the error messages to give the user some more feedback.

### DIFF
--- a/Kernel/Output/HTML/Preferences/Password.pm
+++ b/Kernel/Output/HTML/Preferences/Password.pm
@@ -177,7 +177,7 @@ sub Run {
 
     # check if password is not matching PasswordRegExp
     if ( $Config->{PasswordRegExp} && $Pw !~ /$Config->{PasswordRegExp}/ ) {
-        $Self->{Error} = Translatable('Can\'t update password, it contains invalid characters!');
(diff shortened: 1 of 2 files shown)


Kernel/Output/HTML/Preferences/Password.pm
```diff
@@ -177,7 +177,7 @@ sub Run {
 
     # check if password is not matching PasswordRegExp
     if ( $Config->{PasswordRegExp} && $Pw !~ /$Config->{PasswordRegExp}/ ) {
-        $Self->{Error} = Translatable('Can\'t update password, it contains invalid characters!');
+        $Self->{Error} = Translatable('Can\'t update password, it does not match the RegExp stated in PasswordRegExp! (See SysConfig: Framework -> Frontend::Agent::Preferences) and/or contact your administrator.');
```
mgruner: Hello @MBRG. Thanks a lot for your proposal.
I believe this message is too technical and regular users will not understand it. How about this one: "This password is forbidden by the current system configuration. Please contact your admin if you have additional questions."? Would that be ok?

mgruner:
Hello @MBRG. Despite your non-reaction I went ahead and merged with the proposed change in 993d7dd4533963095c6a19f7340f35cc13a1a668. Please let me know if you agree. Thanks a lot for your contribution!

PR #988, opened 2016-02-12
[ Rel-5_0 Bug 11866 ] - Customerhistory destroyies the cutsomerinformation

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11866

Hello @mgruner ,

Hereby is proposal for fixing reporter issue. In AgentTicketEmail and AgentTicketPhone there was no break line before CustomerTickets <div>, which was resulting in mixing CustomerInformation with CustomerTickets, when CustomerInformation is longer then default.

Regards,
Sanjin


dvuckovic:
Hi S7, thanks for your bugfix, it's been merged.

PR #991, opened 2016-02-15
Rel-5_0 - OTRS - updated unit test - part 12

s7design:
Hi @mgruner 

There is one more PR with updated unit tests. 
I had a problem with using Helper object with RestoreDataBase for test ACL/DB/ACL.t and CloudService/Backend/Configuration.t. There were error in test results with RestoreDataBase  param I used simple CleanUp for cache, I checked in DB state is cleaned after running these tests.

Regards
Zoran


mgruner:
Hello Zoran, thanks, I merged this. I also added RestoreDatabase to the two tests you mentioned. We just have to make sure that there are no errors in DB queries by improving the test cases (using real numbers for inexistent IDs) or the code (finding duplicates in Perl rather than in the DB's unique index). Best regards, Martin

mgruner:
Please let me know if this works for you.

PR #992, opened 2016-02-15
bug4839: closed (un)successful -> closed (un)successfully 

nielsdimmers:
hi @mgruner ,

This pull request is for bug4839: 'Wrong english wording for ticket state "closed successful"'
Changed every occurence of closed successful and closed unsuccessful in the code. Added two modules to the upgrade script to update the wrong states if the OOB states are being used.
After merging this update it is highly recommended to run the "DBUpdate-to-6.pl" script or change the states in the system manually. Otherwise any tests closing tickets, handling closed tickets or states might fail.

Regards,

Niels


nilei:
@nielsdimmers, for the update to OTRS 6, do you also consider ACLs, ticket process configurations, postmaster filter, SysConfig, generic agent jobs and statistics?

-- Cheers, Nils (without "e" :-)

nielsdimmers:
@nilei,
I just did a grep on the entire system and changed every occurence of "closed (un)successful" to "closed (un)successfully". There aren't that many occurences of these words, mostly in tests but just a few in the code. In the default system, there aren't any ACL's or filters, should I actually consider those? I used Kernel::System::AdminState to update the state name in the database, and the sub used there also updates any occurences in the config.

-- Niels

nilei:
@nielsdimmers 

Yes, you have to consider ACLs, ticket process configurations, postmaster filter, SysConfig, generic agent jobs and statistics for existing setups. I think this cannot be solved 100% as one may have an ACL matching with regular expression on e.g. "^closed success.+" or even worse on "NOT "closed successful"", which would be unintentional true if the name is changed to "closed successfully".

If this PR will be accepted, the DBUpdate***.pl script and detailed migration documentation is at the core of success for this change!

-- Cheers, Nils

carlosfrodriguez:
Apparently Travis check failed due to a timing issue, most possible not related to the fix.

((enjoy))
cr

mgruner:
Hello all, thanks for your proposal and feedback. I'll close this PR for now as it is not complete yet. If a clean migration for the most important use cases is added, we can consider applying this change to master. Again, thanks a lot!

PR #993, opened 2016-02-15
Rel-5_0 - OTRS - ProcessManagement - updated unit test - part 1

s7design:
Hello @mgruner ,

Here's updated first part of unit tests for ProcessManagement. Removed sub function in tests for creation of ticket and process objects. IMO it is no longer necessary. 

Please if you could verify it, tests perform as expected with these changes. 

Regards,
Sanjin


mgruner:
Merged, thanks!

PR #995, opened 2016-02-17
Changed the wording of Send by default and the explanatory message.

MBRG:


mgruner:
Hello @MBRG, thanks! @carlosgarciac, can you please review this? IMHO the label should not be so long, maybe it can stay as it is. Regarding the explanation, what about "Should the notification be sent to *assigned recipient* agents who have not yet made a choice in their preferences?"

MBRG:
Hi @mgruner , @carlosgarciac,
if the lable should be kept short we should rename it to "Enabled by default" as the current term "Send by default" caused confusion in the recent past.

carlosgarciac:
Hi @mgruner, @MBRG,

IMO the issue is we should provide a couple of text 'Label and Description' that helps the administrator to understand the immediate effect of enable/disable this field, that means this value is used as Default Value choosing a notification in the Agent Preferences, as result of this we have the sending of the notification, that should be transparent for the Agent but explained to the Administrator.

Then my suggestion could be (sorry for my bad English):
Label: Default value on Agent Preferences (perhaps not short enough)
Explanation: This is the default value for who have not yet made a choice in their preferences, if the box is enabled the notification will be send to assigned recipient agents.

Best regards,
Carlos García

mgruner:
Sounds good. Label could be "Active by default in agent preferences". Explanation: "This is the default value for assigned recipient agents who didn't make a choice for this notification in their preferences yet. If the box is enabled, the notification will be sent to such agents."

MBRG:
Hi, 
yes, I like "Active by default in agent preferences". 
About the explanation: "Activates this notification by default for defined recipient agents. If the box is enabled, the notification will be sent to agents, who didn't make a choice for this notification in their preferences yet." 

A little different from the placement of the words.

s7design:
Hi all,

I suppose you know for PR https://github.com/OTRS/otrs/pull/986/files where I provided this changes as bugfix for the bug #11862

Regards
Zoran

carlosgarciac:
Hi @MBRG, @mgruner and @s7design,

At the end I decided to use the texts from @mgruner, since IMO explain with more detail the immediate effect on preferences and the impact of the default value.

The change was included in 5f84f9f.

Closing.


Best regards,
Carlos García

PR #996, opened 2016-02-17
[ Rel-5_0 Bug 11875 ] - Printed timestamp is shown with GMT

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11875

Hello @mgruner ,

Hereby is our proposal for fixing reporters issues. Changed variable $Time in PDF print output for 'printed by' time, to call FormatTimeString() function which includes 'UserTimeZome' preference and modify time by it.

Bug is reported for AgentTicketPrint and CustomerTicketPrint, but changed AgentTicketSearch and CustomerTicketSearch as well for PDF output.

Please let me know if there are any questions or issues regarding this proposal.

Kind Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketPrint.pm
+++ b/Kernel/Modules/AgentTicketPrint.pm
@@ -176,11 +176,16 @@ sub Run {
         );
     }
 
-    # get PDF object
-    my $PDFObject = $Kernel::OM->Get('Kernel::System::PDF');
+    # get needed objects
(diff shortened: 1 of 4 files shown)


jepf:
Merged with OTRS 5 (33d715f1fef6f02719efdfc05fe14d5cb1a3830c), thanks to S7.
OTRS 4 version in progress.

PR #998, opened 2016-02-18
Rel-5_0 - OTRS - updated unit test - part 13

s7design:
Hi @mgruner 

There are updated test for console command CustomerCompany/Add.t and CustomerUser/Add.t. You can see that I tried solve problem adding and updating existing CustomerCompany. There is a problem because Company name is Unique and CustomerID is primary key and I am not sure if I covered all conflicted cases.  I tested a lot test cases, and if you like I can extend unit test/CustomerComany but your opinion in that moment will be welcomed.

I made new function CustomerList because the function CustomerCompanyList return a little bit strange result. 
```

%List = {
          'example.com' => 'example.com Customer Inc.',
          'acme.com'    => 'acme.com Acme, Inc.'
        };
```

and it is not usable for reversed List ( 'example.com Customer Inc.', - **customer_id name**)

I did not want update more test in this PR because I have dilemma if it is good that I have done.

Regards
Zoran

### DIFF
--- a/Kernel/System/CustomerCompany.pm
+++ b/Kernel/System/CustomerCompany.pm
@@ -120,6 +120,24 @@ sub CustomerCompanyAdd {
         }
     }
 
+    my %ExistingCustomerCompany = $Self->{ $Param{Source} }->CustomerList( Valid => 0 );
+    my %ExistingCustomer = reverse %ExistingCustomerCompany;
+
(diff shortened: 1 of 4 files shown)


mgruner:
Hello Zoran,
thank you very much. I agree basically with your solution, but at this point I would not want to change the API of these modules. I have therefore reverted the changes in Kernel/ and simplified the test case to not test for the duplication problem. That is good enough for now.
Best regards, Martin

PR #1002, opened 2016-02-19
Improved rendering for custom development

rolfschmidt:
Hi OTRS,
if you are doing some custom development to hide dynamic fields for AgentTicketClose for example and it will get rerendered in Chrome browser then the chrome will be not able to calculate the width of the hidden field and crash very hard, because of this endless loop. I added this small improvement to prevent the crash.

Big thanks also to @thorsteneckel !

Best regards,
Rolf

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -2122,7 +2122,7 @@ Core.UI.InputFields = (function (TargetNS) {
                     // redraw event is critical on hidden elements because
                     // e.g. chrome can't calculate the width of hidden elements correctly
                     // so we skip it for hidden elements
-                    if ( !$(this).is(':visible') ) return;
+                    if (!$SearchObj.is(':visible')) return;
 
                     CloseOpenSelections();
                     if (Filterable) {
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -507,7 +507,7 @@ Core.UI.InputFields = (function (TargetNS) {
 
                     // If first selection, we must shorten it in order to display it
                     if (i === 0) {
-                        while (OffsetLeft + $SelectionObj.outerWidth() >= MaxWidth) {
+                        while (MadWidth > 0 && OffsetLeft + $SelectionObj.outerWidth() >= MaxWidth) {
                             $TextObj.text(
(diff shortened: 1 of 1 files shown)


mrcbnsls:
@zottto, could you please have a look? Wouldn't it be better to instead just check if the current element is visible before doing anything on it?

zottto:
Hi @rolfschmidt,
can you please explain the usecase for this change a little bit more in detail?
In line#995 of Core.UI.InputField.js the initialization for the modern inputfields is only started if the underlying select field is visible. And only than the ShowSelectionBoxes function is executed which contains the MaxWidth calculation.
So in my understanding, there can naver be the problem that ShowSelectionBoxes tries to calculate the width of an invisible element.

Thanks for clarification,
Marc

thorsteneckel:
Hi @zottto 

On initialization the field is shown and gets hidden in the workflow of interactions of the user. So in the first place there is nothing wrong here. Within the initialization the redraw binding is set: https://github.com/rolfschmidt/otrs/blob/fd85259148dbe777eb175987e0401ea001326ad6/var/httpd/htdocs/js/Core.UI.InputFields.js#L2120.

Now the user interacts and changes some fields and some select fields get hidden. The change triggers a AJAX request too which triggers the custom redraw binding: https://github.com/OTRS/otrs/blob/master/var/httpd/htdocs/js/Core.AJAX.js#L315.

Now the initialized, hidden field gets "redraw"ed and ends in an endless loop caused by the MaxWidth = 0.

I hope I could make it clearer. If not let us know.

- Thorsten

zottto:
Hi @thorsteneckel ,
thanks for the clarification, makes sense now.

But wouldn't it be better to handle that problem directly in the redraw event fucntion and exit this function if the field is not visible at the moment of the redrawing?
Additionally we could also apply your change to the while-loop in ShowSelectionBoxes to make it more robust.

What do you think?

Marc

rolfschmidt:
Heyhoo @zottto ,
I added the check for the redraw event and also tested it. Your idea of the check in the event works nice :+1: Typo is fixed also.

zottto:
Hi @rolfschmidt,
as you can see, the unit tests are now failing. Could you please correct that?
Additionally, I think the visibility check should not be done on the SelectObj (which is the select element, which is always invisible for modernized input fields), but on the SearchObj, because that is the field, where Chrome fails on getting the width from.

Sorry for bouncing again... 

Marc

zottto:
Merged with d35d419 (master) and 96486f2 (OTRS5). Thanks!

PR #1003, opened 2016-02-19
Rel-5_0 - OTRS - updated unit test - part 14

s7design:
Hi @mgruner 

In this PR I updated Console command selenium tests for Role and Group. I also improved code for Group and Role add and update functions with checks if there is duplicate entities in DB, because all of that I updated Group.t and Role.t in this PR as well.

Regards
Zoran


mgruner:
Merged, thanks! I also removed the ```@ObjectDependencies``` from Group.t, seems unneeded?

PR #1005, opened 2016-02-22
Bug 11885 - uninitialized value in scripts/DBUpdate-to-5.pl

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11885

Hi @carlosgarciac and @mgruner 

Here is fixed mistake which the reporter mentioned in the bug report.

Regards
Zoran


mgruner:
Merged, thank you!

PR #1006, opened 2016-02-22
Rel 5 0 ut restore database part15

s7design:
Hi @mgruner 

There are updated Selenium test from Console/Command. I improved some tests with adding test entities like queues, users, roles etc. I believe that it is better then updating  default values ( Junk, root@localhost ... ) 

Regards
Zoran


mgruner:
Very nice, merged, thank you!

PR #1011, opened 2016-02-24
Rel 5-0 ut restore database part19

s7design:
Hi @mgruner 

- updated Console/Command selenium tests

You can see that I add in many tests RestoreDatabase even though there is only check exit code because in all of these Ticket Console commands there is changes in DB ( or could be - ArchiveCleanup, EscalationIndexRebuild ... etc ).

Regards
Zoran


mgruner:
Merged, thanks a lot!

PR #1012, opened 2016-02-25
Rel 5 0 ut restore database part20

s7design:
Hi @mgruner,

There is updated Selenium test from ProcessManagement/TransitionAction/

I tried to fix the issues because of that PostgreSQL makes problem for the following test cases:

https://github.com/OTRS/otrs/blob/master/scripts/test/ProcessManagement/TransitionAction/TicketResponsibleSet.t#L142
https://github.com/OTRS/otrs/blob/master/scripts/test/ProcessManagement/TransitionAction/TicketOwnerSet.t#L142

I did not manged to solve it, or to be honest I did not invest enough time, if you would like I can try to solve it.

Regards
Zoran

### DIFF
--- a/scripts/test/ProcessManagement/TransitionAction/TicketCustomerSet.t
+++ b/scripts/test/ProcessManagement/TransitionAction/TicketCustomerSet.t
@@ -15,19 +15,24 @@ use vars (qw($Self));
 use Kernel::System::VariableCheck qw(:all);
 
 # get needed objects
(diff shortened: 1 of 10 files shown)


mgruner:
Hello Zoran, thanks for the PR, merged. I fixed the TicketResponsibleSet() and TicketOwnerSet() methods and reactivated the test cases.

s7design:
Great thanks, it is much better, with this we covered more test cases

Regards
Zoran

PR #1015, opened 2016-02-26
[Rel-5_0 Bug 11897] - Text in the Queue View is not translateable

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11897

Hi @mgruner 

I think that we fix the issues that reported mentioned in the bug. I am not sure is there nedded Interpolate directive. I left it but please check this.
`
title="[% Translate(Data.Description) | Interpolate | html %]`

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentTicketOverviewMedium.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentTicketOverviewMedium.tt
@@ -63,7 +63,7 @@
             <ul class="Actions InlineActions">
                 <li class="ResponsiveActionMenu"><a href="#" id="ResponsiveActionMenu[% Data.TicketID | html %]" title="[% Translate("Open / Close ticket action menu") | html %]"><i class="fa fa-bars"></i></a></li>
 [% RenderBlockStart("InlineActionRowItem") %]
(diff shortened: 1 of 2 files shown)


carlosfrodriguez:
Hi @s7design,

I see in the code for Medium.pm that for example the name is already translated in Perl, before creating the TT block so I think that it might be better to have the same solution for both cases.

```
push @ActionItems, {
                HTML        => $Output,
                ID          => $Item->{ID},
                Name        => $LayoutObject->{LanguageObject}->Translate( $Item->{Name} ),
                Link        => $LayoutObject->{Baselink} . $Item->{Link},
                Target      => $Item->{Target},
                PopupType   => $Item->{PopupType},
                Description => $Item->{Description},
                Block       => $Item->{Block},
            };
```

What do you think about translating it in Perl?

((enjoy))
cr

s7design:
Thanks @carlosfrodriguez 
I updated PR, I have seen now that there is Translate in Perl Module for Name of menu item and I added for Description as well. I analyzed a little if it would be better there was only Translate on TT side. What do you think about that options. I will be able to update Preview and Medium Perl modules and remove Translate from all places there and add in TT if you think it is good solution. 

Regards
Zoran

s7design:
Hi @carlosfrodriguez 

I see some errors in Travis. I checked them, but it is not mistake that caused by my changes, I think that there are some mistakes in the rest of the framework. 

Regards
Zoran

PR #1016, opened 2016-02-29
Master - update Selenium tests - TicketZoom

s7design:
Hi @mgruner 

As we discussed by mail about it, there is very strange case. I am not sure what is real problem here. I could not reproduce the problem on my local system. We analyzed it and maybe there is problem in Customer Information because we saw that there was wrong data for CustomerUserAdd(). However it is so strange, why it works on our local system :) 

In TicketInformation we change way how the escalation times are tested. I believe it is better to get Ticket data before checking these values. In previous version there could be problem if escalation time is changed in meantime for example instead 29m there could be 30m for fast system or 28m for slow system. I am not sure about that i just suppose.

Regards
Zoram


mgruner:
Merged, thanks. Let's see.

PR #1018, opened 2016-02-29
Rel 5 0 ut restore database part23

s7design:
Hi @mgruner 

- updated Unit tests

You will see that removed comment in ArchiveSystem.t for some test cases. They work for me, but I am not sure if it works on other systems. Please check that.

https://github.com/OTRS/otrs/pull/1018/files#diff-8d17ec54c3ba4ea8a5f360f9456f032fL222

Regards
Zoran


mgruner:
Merged, thank you!

PR #1021, opened 2016-02-29
Marked messages for translation

urbalazs:
Hi @mgruner,
This is the second part of marking messages for translation. Please also cherry-pick this into rel-5_0. After commit it would be great to sync translation with Transifex, because in this PR (and the previous one) a lot of strings have been added, so let you give some time for translators to translate them into their languages - before releasing 5.0.8.


mgruner:
All done as you suggested. Thanks @urbalazs! ++

PR #1024, opened 2016-03-02
Master - proposal for fix too long DF label and fix some bugs

s7design:
Hi @mgruner 

There is our proposal for fix long label of DynamicField. I only truncated on frontend, it is the simplest way. Maybe I can invest more time if you  don't like this fix.  

Working on this I saw some bug I think it i s because of the previous changes with 

`use base 'Kernel::Output::HTML::Base';`

There was a problem because there are not some params in $Self. I saw these issues in the Fred.

This is example how it looks if DF label is too long:
df_long_label

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/TicketZoom/LinkTable.pm
+++ b/Kernel/Output/HTML/TicketZoom/LinkTable.pm
@@ -22,7 +22,7 @@ sub Run {
     my $LinkListWithData =
         $Kernel::OM->Get('Kernel::System::LinkObject')->LinkListWithData(
         Object           => 'Ticket',
-        Key              => $Self->{TicketID},
+        Key              => $Param{Ticket}{TicketID},
         State            => 'Valid',
         UserID           => $Self->{UserID},
         ObjectParameters => {
--- a/Kernel/Output/HTML/TicketZoom/TicketInformation.pm
+++ b/Kernel/Output/HTML/TicketZoom/TicketInformation.pm
@@ -184,12 +184,14 @@ sub Run {
 
     # check if ticket is normal or process ticket
     my $IsProcessTicket = $Kernel::OM->Get('Kernel::System::Ticket')->TicketCheckForProcessType(
-        'TicketID' => $Self->{TicketID}
+        'TicketID' => $Ticket{TicketID},
     );
+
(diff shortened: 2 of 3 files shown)


s7design:
Hi @mgruner 

I forgot yesterday to note that without my changes in TicketInformation about 

my $DisplaySettings = $ConfigObject->Get("Ticket::Frontend::AgentTicketZoom"); and
https://github.com/OTRS/otrs/pull/1024/files#diff-723ec6c0b4b19c4b69e7303b97a0ce5fR259

DF is not appears in the Zoom screen.

Regards
Zoran

Kernel/Output/HTML/TicketZoom/TicketInformation.pm
```diff
@@ -184,12 +184,14 @@ sub Run {
...
-        'TicketID' => $Self->{TicketID}
+        'TicketID' => $Ticket{TicketID},
     );
 
+    my $DisplaySettings = $ConfigObject->Get("Ticket::Frontend::AgentTicketZoom");
+
     # overwrite display options for process ticket
     if ($IsProcessTicket) {
-        $Param{WidgetTitle} = $Self->{DisplaySettings}->{ProcessDisplay}->{WidgetTitle};
+        $Param{WidgetTite} = $DisplaySettings->{ProcessDisplay}->{WidgetTitle};
```
mgruner: ?

Kernel/Output/HTML/TicketZoom/LinkTable.pm
```diff
@@ -22,7 +22,7 @@ sub Run {
     my $LinkListWithData =
         $Kernel::OM->Get('Kernel::System::LinkObject')->LinkListWithData(
         Object           => 'Ticket',
-        Key              => $Self->{TicketID},
+        Key              => $Param{Ticket}{TicketID},
```
mgruner: This has been fixed in the meantime.

mgruner:
Actually I believe this is not a real world case. If the administrator defines labels that are too long, they still need to be shown and cannot be truncated. They might be different only in the parts that are missing. Another option might be to improve the CSS to allow for this label length. @mrcbnsls, what do you think?

PR #1028, opened 2016-03-03
Rel 5 0 ut restore database part28

s7design:
Hi @mgruner ,

- Updated Unit tests

Regards,
Zoran

### DIFF
--- a/scripts/test/Daemon/DaemonModules/SchedulerTaskWorker/AsynchronousExecutor.t
+++ b/scripts/test/Daemon/DaemonModules/SchedulerTaskWorker/AsynchronousExecutor.t
@@ -185,7 +185,7 @@ my @Tests = (
 
 # get task handler objects
 my $TaskHandlerObject
-    = $Kernel::OM->Get('Kernel::System::Daemon::DaemonModules::SchedulerTaskWorker::AsynchronousExecutor');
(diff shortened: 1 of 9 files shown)


scripts/test/Daemon/DaemonModules/SchedulerTaskWorker/AsynchronousExecutor.t
```diff
@@ -179,7 +185,7 @@ my @Tests = (
 
 # get task handler objects
 my $TaskHandlerObject
-    = $Kernel::OM->Get('Kernel::System::Daemon::DaemonModules::SchedulerTaskWorker::AsynchronousExecutor');
+    = $Kernel::OM->Get('Kernel::System::Daemon::DaemonModules::SchedulerTaskWorker::AsynchronousExecutorF');
```
mgruner: There seems to be a typo here.

mgruner:
I also get warnings about "rollback ineffective with AutoCommit enabled". Could it be that there are new DBObjects created? In that case RestoreDatabase would not work.

s7design:
Hi @mgruner 

Thanks, yes it was my mistake. I updated PR

Regards
Zoran

mgruner:
Merged, thanks! I also added a check to the Helper to cause a test failure if the rollback fails to make sure this does not happen in other test cases as well.

PR #1029, opened 2016-03-03
Fixed tooltip left

s7design:
Hi @dominikklein,@mgruner and @mrcbnsls 

Working on OTRSDynamicFieldAttachment package we saw that there is mybe a little UX issues with Tooltip for Server errors (max number of attachments and size of attachment)

attachment_tooltip

Acctually there is wrong style in the framework. I changed  margin-left on 0px. Perhaps it may be completely removed.

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Tooltip.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Tooltip.css
@@ -22,7 +22,7 @@ div.Tooltip {
 }
 
 div.Tooltip.TongueLeft {
-    margin-left: -37px;
+    margin-left: 0px;
 }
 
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Hey Zoran,

I tend to disagree on this change. In fact, the tooltip looks broken in the case you mentioned, but looks fine in all other cases (as far as I can see):

bildschirmfoto 2016-03-17 um 17 12 44

There should be a little tongue on the tooltip indicating which field the tooltip is referring to. Why is it missing in your example? Also, the text is strangely intended.

I'll close this PR, but please feel free to comment on it if you still think this is a framework issue which needs to be fixed. Thanks!

Marc

PR #1035, opened 2016-03-07
Replace link of inline content regexp fix

pboguslawski:
Upstream regexp fix introduced in
http://bugs.otrs.org/show_bug.cgi?id=7516#c1
may be too loose in some corner cases, i.e. if other img attribute also
has ContenID string inside like

```
<img src="/app/index.pl?Action=PictureUpload&amp;FormID=1111111111.2222222.33333333&amp;ContentID=11112323234343434344434342343234@aaaaaa"other_img_attribute="ContentID=foo@bbbbbb">
```

ContentID in regexp should be rather searched only inside src attribute
not whole img.

Related: https://dev.ib.pl/ib/otrs/issues/23
Author-Change-Id: IB#1049566

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-07 Replace link of inline content regexp fix.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
  - 2016-03-03 Fixed bug#[11872](http://bugs.otrs.org/show_bug.cgi?id=11872) - TicketGet function returns SolutionTime variable.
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Pawel,
I added the example you gave to RichTextReplaceLinkOfInlineContent.t and there is no change after applying your diff. Can you please extend your PR with a unit test case that demonstrates the issue? Unit test coverage is something that we would expect for bug fixes. And if there is really a bug, it would be helpful to also have an entry at bugs.otrs.org that we can reference to in the CHANGES.md file.

pboguslawski:
Hello,

> Can you please extend your PR with a unit test case that demonstrates the issue? 

Just merged commit with test cases.

> it would be helpful to also have an entry at bugs.otrs.org that we can reference to in the CHANGES.md file

Please consider referencing in CHANGES.md github PRs or bugs.otrs.org to avoid redundancy.

Regards,
Pawel

mgruner:
Merged in 918c6aacb5d16d0cc09de4db353ce66eb5f07105 and backported, thanks a lot @pboguslawski! I applied one small modification, test cases still work.

PR #1034, opened 2016-03-05
Fixed bug#3833

urbalazs:

PR #1037, opened 2016-03-07
Translatable out of office message

pboguslawski:
Out of office strings in user name like

   User Name *** out of office till YYYY-D-M/2d ***

should be translatable not hardcoded english string.

Related: https://dev.ib.pl/ib/otrs/issues/24
Author-Change-Id: IB#1004603

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-07 Out of office message should is now translatable.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
  - 2016-03-03 Fixed bug#[11872](http://bugs.otrs.org/show_bug.cgi?id=11872) - TicketGet function returns SolutionTime variable.
  - 2016-03-03 Fixed bug#[8631](http://bugs.otrs.org/show_bug.cgi?id=8631) - "ghost" tickets after merge.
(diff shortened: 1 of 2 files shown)


urbalazs:
I have a pull request for the same problem: https://github.com/OTRS/otrs/pull/1034
It shouldn't split the translatable string into substrings, because translators can not be able to understand the meaning of the string, if those are in separated messages. For example this one:

<pre>'*** ' . $LanguageObject->Translate('out of office till') . " $TillDate ($Till " . $LanguageObject->Translate('days') . ') ***';</pre>

Results 2 strings for translators:
out of office till
day

My solution:
<pre>$Kernel::OM->Get('Kernel::Language')->Translate('*** out of office till %s (%s day) ***', $TillDate, $Till);</pre>

Results only one string for translators:
*** out of office till %s (%s day) ***

pboguslawski:
@urbalazs: Thanks for the hint. We've fixed it in cb8834d and added correct date formatting and small optimization.

mgruner:
Hello @pboguslawski and @urbalazs. Thanks for your suggestions so far! There's one thing I don't understand about this PR. How do you determine the user's language? Usually we use the LanguageObject inside of the LayoutObject, which we cannot do in User.pm (backend). In this case I'd assume that always the default language would be used, which would defeat the purpose of this PR. Besides that, GetUserData will perform caching, so if the data is requested from another user, it might get returned with the wrong language for the out of office string...

I'll close this PR for now. Feel free to provide more information and a PR which solves the mentioned issues. I'm not sure if this can be solved with the current (bad) architecture of the out of office messages...

pboguslawski:
Hello,

Thank you for pointing problem with using LanguageObject in User.pm.

Please see workaround to allow one to customize out of office message without using LanguageObject: https://github.com/OTRS/otrs/pull/1079

Regards,
Pawel

PR #1040, opened 2016-03-07
Solved Bug8249 - system addresses should be unique

nielsdimmers:
Hi @mgruner,
I solved this bug by creating an additional check for already existing system e-mail adressess.
Regards,
Niels


mgruner:
Hello Nils,
thanks a lot for your contribution! This seems to be broken atm. You cannot save an existing address any more. Please change that and also add coverage for the bugfix to the Selenium tests in AdminSystemAddress.t.
Best regards, Martin

mgruner:
Hello @nielsdimmers. Thanks a lot for your contribution. Since it is incomplete, I will close it for now. Thanks again.

PR #1041, opened 2016-03-08
Fixed an error while self-creating customer account

pboguslawski:
When one tries to self-create customer account from customer panel,
OTRS throws an error

Can't call method "Translate" on an undefined value at
/opt/otrs/Kernel/System/Web/InterfaceCustomer.pm line 840.

Related: https://dev.ib.pl/ib/otrs/issues/25
Author-Change-Id: IB#1049592


mgruner:
Merged in dca1523, thanks!

PR #1045, opened 2016-03-09
Rel-5_0 - improved tests -  ported Language object to OM

s7design:
Hi @mgruner 

Herby I updated some tests with ported Language object to OM. 
In AdminEmail.t there is added sysconfig SendmailModule  'Kernel::System::Email::DoNotSendEmail'. I think that it is better if there is such sysconfig in all tests where some mails/notifications are sent.

Regards
Zoran

### DIFF
--- a/scripts/test/Selenium/Agent/Admin/AdminACL.t
+++ b/scripts/test/Selenium/Agent/Admin/AdminACL.t
@@ -13,7 +13,6 @@ use utf8;
 use vars (qw($Self));
 
 use Selenium::Remote::WDKeys;
-use Kernel::Language;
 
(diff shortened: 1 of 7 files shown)


mgruner:
Hello Zoran,

I actually don't have a problem with local instances of the Language object, since this is not really a singleton. Especially if you want to have several languages in one test file. So I would want to keep the code as it is.
Regarding the emails: in many of our selenium tests there is a potential for emails being sent (notifications etc.). This probably needs to be covered by the environment setup code to avoid too much duplication in our files. Our unit test servers have something like that.

Best regards, Martin

PR #1047, opened 2016-03-09
Rel-5_0 - improved tests - set SendmailModule

s7design:
Hi @mgruner 

Herby I updated some tests, actually I set SendmailModule to Kernel::System::Email::DoNotSendEmail'. As I mentioned in the previous PR, I think that it is improvement because I saw testing some of these test that if this sysconfig Item is not set well test could work more than 1 minute.

Regards
Zoran


mgruner:
Please see my comment on #1045.

PR #1049, opened 2016-03-10
Rel 5 0 ut restore database part39

s7design:
Hi @mgruner 

- updated unit tests

Regards
Zoran


mgruner:
Hello Zoran,

I don't see an improvement in these test files. If we have a Kernel::System::ObjectManager->new(), there should also be a use statement for it, right? I know that the parent process already has it loaded, but for a syntactically correct .t file it is needed. Please let me know if there is anything that I overlooked.

Best regards, Martin

PR #1050, opened 2016-03-10
Changed wording, as this option is still valid in newer versions of I…

MBRG:
…E (like 11).

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -450,7 +450,7 @@
         </Setting>
     </ConfigItem>
     <ConfigItem Name="DisableMSIFrameSecurityRestricted" Required="0" Valid="1">
-        <Description Translatable="1">Disable restricted security for IFrames in IE. May be required for SSO to work in IE8.</Description>
+        <Description Translatable="1">Disable restricted security for IFrames in IE. May be required for SSO to work in IE.</Description>
         <Group>Framework</Group>
         <SubGroup>Core::Web</SubGroup>
         <Setting>


mgruner:
Hi @zottto. Can you please review this? If it is only relevant for IE8, it could also be removed, right?

MBRG:
Hi, we tested this today on a system and with IE 11 and it did have an effect (a positive one). So we should not drop the option altogether.

zottto:
Hi @MBRG, hi @mgruner,

the setting is relevant for all versions of IE, but we added it to our code as an replacement for the sandbox attribute, that is not supported by older IEs.
The sandbox attribute is supported starting IE10, so we could remove this special IE setting for OTRS 5 or at least for master.

What do you think?

Marc

mgruner:
Hi Marc,

will SSO work in IE10+ then without the setting?

Martin Gruner
Team Lead R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

MBRG:
Hi,

this came up because on an OTRS 5 system and SAML SSO with IE11 was bringing up errors and was not showing the ticket zoom correctly. A "script disbled" message was shown instead of the ticket content. After changing this option, the content was shown correctly.

So in this setting, this option solved a problem on the system which would not be solvable otherwise with the current knowledge.

zottto:
Hi @MBRG,

this option disables a IE-specific security setting that was introduced for older IEs that do not understand the general sandbox attributes on iframes.
Because IE10+ understand those sandbox attributes now, the IE-specific security setting is not needed anymore in these browsers.
We will remove the IE-security setting completely which also makes the sysconfig setting unnecessary.

zottto:
Improved description for OTRS 5 (eebea78) and 4 (3cfed1f).

zottto:
Removed config setting and IE-specific attribute completely for master (a5c173a).
Thanks for your help!

PR #1052, opened 2016-03-10
Rel 5 0 ut restore database part39

s7design:
Hi @mgruner 

I updated tests and rollbacked a 'use' for Kernel::System::ObjectManager.  I tested it without it and the test worked, CodePolicy and Travis did not show any error or warning so I thought it is possible to avoid it. 

To be honest I was not sure about that, but sometimes if I am not sure I commit solution in witch I am not sure absolutely, but if it is not possible I can learn something new. Being in comfort zone is good but there we could not improve ourselves :)

I hope it is better now, there were not so much changes but I believe it is better that we correct this typos and some clean caches.

Regards
Zoran

### DIFF
--- a/scripts/test/ObjectManager/Can.t
+++ b/scripts/test/ObjectManager/Can.t
@@ -108,10 +108,13 @@ for my $Directory ( sort @DirectoriesToSearch ) {
                 "$Module | $1->$2()",
             );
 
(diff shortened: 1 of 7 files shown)


mgruner:
Merged, thanks Zoran!

PR #1055, opened 2016-03-10
Be sure that a outputfilter post hook is always generated

rolfschmidt:
Hi,
what about to change the behaviour of the hooks generation to always render it by minimum 1 to be sure that the hook get always rendered. In my opinion this makes the hooks much greater and i think that the current behaviour can make a lot of trouble, here my example:

You want to add code after the "DynamicField" block in different masks. If you now use the hook and some masks do not define dynamic fields in sysconfig then also the outputfilter will not match, because the hook does not exist.

Best regards,
Rolf

### DIFF
--- a/Kernel/Output/Template/Document.pm
+++ b/Kernel/Output/Template/Document.pm
@@ -84,27 +84,30 @@ sub _InstallOTRSExtensions {
                 my $BlockName = $stash->get('BlockName');
                 my $ParentBlock = $stash->get('ParentBlock') || $stash->{_BlockTree};
 
-                return if !exists $ParentBlock->{Children};
-                return if !exists $ParentBlock->{Children}->{$BlockName};
-
                 my $TemplateName = $stash->get('template')->{name} // '';
                 $TemplateName = substr( $TemplateName, 0, -3 );    # remove .tt extension
                 my $GenerateBlockHook =
                     $Context->{LayoutObject}->{_BlockHookSubscriptions}->{$TemplateName}->{$BlockName};
 
-                for my $TargetBlock ( @{ $ParentBlock->{Children}->{$BlockName} } ) {
-                    $output .= "<!--HookStart${BlockName}-->\n" if $GenerateBlockHook;
-                    $output .= $Context->process(
-                        $TargetBlock->{Path},
-                        {
-                            'Data'        => $TargetBlock->{Data},
-                            'ParentBlock' => $TargetBlock,
-                        },
-                    );
-                    $output .= "<!--HookEnd${BlockName}-->\n" if $GenerateBlockHook;
-                }
-                delete $ParentBlock->{Children}->{$BlockName};
+                if ( exists $ParentBlock->{Children} && exists $ParentBlock->{Children}->{$BlockName} ) {
+                    for my $TargetBlock ( @{ $ParentBlock->{Children}->{$BlockName} } ) {
+                        $output .= "<!--HookStart${BlockName}-->\n" if $GenerateBlockHook;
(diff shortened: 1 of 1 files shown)


thorsteneckel:
++

carlosfrodriguez:
Hi @mgruner,

I'm not sure if this is the best solution or not, but the concept is really interesting, we had also some problems in certain packages where the hooks are not displayed because of the block is never called, due a configuration change or so. Then you can't always trust in the hooks, only when the call of them is hard coded.

If we could have a mechanism to request the hooks to be placed where they suppose even if the block is not really called it will be a great help for everyone.

((enjoy))
cr

mgruner:
Hello @rolfschmidt. Certainly very interesting! But I see a logical problem here: child blocks of a non-rendered block will still not appear, even if they are called for via hooks.

carlosfrodriguez:
Hi @mgruner,

Yes that is a problem, and currently I don't know how to engage, but in the ideal world we should be able to print the hooks no matter the blocks are actually called or not, even got the child blocks.

Provably we will need to remove all the blocks and use pure TT to do that I don't know.

((enjoy))
cr

jepf:
I've built output filters and I'm sure there are many other output filters in which the existence of hooks is being used to decide what happens or will be output. In these cases the presence of a hook is being evaluated as "this block has been output and has content".

These output filters would break if hooks for empty blocks would be output.

rolfschmidt:
@jepf we also could add a new sysconfig type to beat this problem. In my opinon we also could create a new sysconfig for hooks which should be always rendered, e.g.:

```
    <ConfigItem Name="Frontend::Template::GenerateBlockHooksALWAYS###100-blub" Required="1" Valid="1">
        <Description Translatable="1">Generate HTML comment hooks for the specified blocks so that filters can use them.</Description>
        <Group>blub</Group>
        <SubGroup>Core</SubGroup>
        <Setting>
            <Hash>
                <Item Key="AgentTicketZoom">
                    <Array>
                        <Item>Tree</Item>
                        <Item>TicketDynamicField</Item>
                    </Array>
                </Item>
```

I'm really open in the way how we fix this, but in my opinion there is a large part of functionality killed because of the missing pre filters and i just search for ways how i could migrate all the pre filter stuff to post and i would be pretty happy if we could find a way to have more hooks. This would bring much more flexibility to the framework and to custom development.

```
Hello @rolfschmidt. Certainly very interesting! But I see a logical problem here: child blocks of a non-rendered block will still not appear, even if they are called for via hooks.
```

@mgruner do you maybe have an idea where we could integrate this stuff better to solve this problem more comprehensive?

@thorsteneckel  and me also open for skype brainstormings and i also would support you with pull requests for this topic. We are really interested in a solution for this topic.

mgruner:
Hello @rolfschmidt. I agree with @jepf that it depends on your point of view if you would consider a behaviour change to be better or worse. Regarding the problem with sub-blocks, I don't see a solution currently for this.
In fact the blocks are a custom TT extension that only exist because of backwards compatibility with DTL (or rather, an easy migration path). It would be desirable to get rid of them in future (that might take long). Maybe we should look for different solutions for extending OTRS templates. @mrcbnsls recently found that Bugzilla seems to have a nice TT based hook system that allows you to place snippets loaded at certain points in the template, and they are TT and have all the data available -- I haven't looked at this in detail yet, and we might need a different solution.
We can certainly skype if you want.
Best regards, Martin

mgruner:
I'll close this for now. Looking forward to improved proposals, though I honestly believe that a good implementation for this would be quite hard if possible.

PR #1057, opened 2016-03-11
Rel-5- Improved AdminPGP screen

s7design:
Hi @mgruner 

I saw that AdminSMIME was updated with PR which we had provided before. Herby, I updated  AaminPGP screen with improvment in case PGP is disabled or not working.

I also improve AdminSmime screen, becous there is Notice sidebar wiget alweys. I t is nessesary only if there is block Disabled or NotWorking.

If you like it you can marge it.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminPGP.pm
+++ b/Kernel/Modules/AdminPGP.pm
@@ -28,22 +28,59 @@ sub new {
 sub Run {
     my ( $Self, %Param ) = @_;
 
-    # get needed objects
-    my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
+    # get layout objects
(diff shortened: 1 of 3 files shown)


s7design:
Hi all,

I updated this PR, there are only modules and test for PGP. The modules for SMIME were deleted, because it is fixed in this PR https://github.com/OTRS/otrs/pull/941

Regards
Zoran

mrcbnsls:
Thanks, merged in ea2a42ebc87addef5b12eca1cd502022dcaaf17a and bfc997e9bf9d6a06e8ced9dd8a419ee0383a9712 with some small improvements.

PR #1058, opened 2016-03-11
Allow quoting font family names in Frontend::RichText::DefaultCSS

pboguslawski:
If you want to quote font family name containing spaces
in Frontend::RichText::DefaultCSS and try to put single quotation
marks like

font-family: Verdana, 'Bitstream Vera Sans', Tahoma; font-size: 10pt; color: #000000;

you'll break rich editor CSS styling; if you try to use double quotes -
it won't work in HTML e-mail body where DefaultCSS content is
already put inside double quotes of style attribute.

This patch fixes it and allowes Frontend::RichText::DefaultCSS to
contain single quotes.

Related: https://dev.ib.pl/ib/otrs/issues/26
Author-Change-Id: IB#1004607

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-11 Allow quoted font family names in Frontend::RichText::DefaultCSS.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
  - 2016-03-03 Fixed bug#[11872](http://bugs.otrs.org/show_bug.cgi?id=11872) - TicketGet function returns SolutionTime variable.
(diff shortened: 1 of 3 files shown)


zottto:
Hi Pawel,
thanks for this contribution.
Changing the surrounding quoting does not fix this issue for me in a proper way. In CSS, both font names in single and double quotes are allowed. With your change, font-names in double quotes would now brake the code.

I would suggest a proper quoting of the string to make sure, both font names in single and double quotes work and do not break the functionality.

I suggest to try to add a template filter like `[% Config("Frontend::RichText::DefaultCSS") | html %]` (or similar) to solve this.

Please update this PR in this way.

Thanks,
Marc

zottto:
Hi Pawel,
with the JS refactoring made in the master branch, the usage of single and double quotes in the default CSS is now possible.
The commit 5020fe0 makes the usage also possible in HTML mails.
This should make your PR unnecessary.

Thanks for your help!
Marc

PR #1059, opened 2016-03-11
Enhanced CSS styling of rich text editor and HTML e-mails

pboguslawski:
This enhancement allows you to modify CSS style of OTRS rich text
editor (i.e. line spacing) using new SysConfig parameter
Frontend::RichText::EditingAreaCSS. Example setting for standard
line spacing and 10px margin around text area:

.cke_editable { line-height: normal; margin: 10px; }

This parameter may contain one line with one or more CSS definitions;
its value is merged into RichText.EditingAreaCSS parameter just
before original

body { [% Config("Frontend::RichText::DefaultCSS") %] }

This modification provides also new SysConfig parameter
Frontend::RichText::MailCSS that allows you to add extra CSS style
definitions that are embedded into html/head/style node of HTML
e-mails; this can be used to modify style of HTML e-mail elements
like anchors or table headers, which cannot be done with standard
DefaultCSS parameter. Example setting of
Frontend::RichText::MailCSS:

a, a:visited, a:active
{
    color: #3D80B2;
    text-decoration: none;
}

a:hover
{
    color: #3D80B2;
    text-decoration: underline;
}

th, td, caption {
    font-family: Verdana, 'Bitstream Vera Sans', 'DejaVu Sans', Tahoma, Geneva, Arial, Sans-serif;
    font-size: 10pt;
}

Related: https://dev.ib.pl/ib/otrs/issues/27
Author-Change-Id: IB#1011028

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-11 Enhanced CSS styling of rich text editor and HTML e-mails.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
  - 2016-03-03 Fixed bug#[11872](http://bugs.otrs.org/show_bug.cgi?id=11872) - TicketGet function returns SolutionTime variable.
  - 2016-03-03 Fixed bug#[8631](http://bugs.otrs.org/show_bug.cgi?id=8631) - "ghost" tickets after merge.
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -431,11 +431,27 @@
         </Setting>
     </ConfigItem>
     <ConfigItem Name="Frontend::RichText::DefaultCSS" Required="1" Valid="1">
-        <Description Translatable="1">Defines the default CSS used in rich text editors.</Description>
+        <Description Translatable="1">Defines the default CSS used in rich text editors and HTML mail body.</Description>
         <Group>Framework</Group>
         <SubGroup>Core::Web</SubGroup>
         <Setting>
             <String Regex="">font-family:Geneva,Helvetica,Arial,sans-serif; font-size: 12px;</String>
+        </Setting>
+    </ConfigItem>
+    <ConfigItem Name="Frontend::RichText::EditingAreaCSS" Required="0" Valid="1">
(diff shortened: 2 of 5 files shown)


zottto:
Hi Pawel,

I like the idea behind your PR. But I don't understand, why DefaultCSS and EditingAreaCSS are splitted up. Wouldn't it make more sense to only have one setting that defines all the CSS for the editor?
And for the outgoing HTML mails, I also would prefer only one setting, that defines all CSS for the mails.

What do you think?

Marc

pboguslawski:
Hi,

(1) DefaultCSS in standard OTRS code is being used in two places: in html web editor and for styling outgoind e-mails. E-mail clients are poor html renderers and may require quirks that should not be applied to editor. Thats make DefaultCSS very limited in our opinion (much care must be taken to test changes in DefaultCSS to avoid breaking some web browser or - more difficult probably - some mail client). We wanted to leave this setting without changes for backward compatibility also (changing its purpose may break CSS config during upgrade?).

(2) We wanted to be able to apply style changes to HTML editor only like line spacing that is hardcoded in ckeditor (and not present in HTML emails), Thats why separate setting for editor only.

(3) DefaultCSS is applied as outgoing e-mail body style; we wanted to be able to modify styles of anchors in e-mails and use embedded styles in e-mail head section and leave DefaultCSS without changes /see (1)/ so why separate CSS setting in e-mail head section.

Separate parameters are not perfect but we find it better way to fine-tune CSS styling in OTRS web/e-mails.

Feel free to change CSS handling the better way if you find something more elegant, functional and compatible (mail clients vs web browsers rendering differences, existing DefaultCSS configurations).

Regards,
Pawel

zottto:
Hi @pboguslawski,

thanks for your answer. I know, that DefaultCSS is used in both editor and e-mails. :-) But for me, keeping DefaultCSS and two additional settings (one for the editor, one for the mails) is too much, the DefaultCSS is redundant at this point.

For sure, removing the DefaultCSS option will brake things after an upgrade. This has to be handled in the update script (converting the content from DefaultCSS and moving it into both new settings, if DefaultCSS was changed and not the default value).

Please update the PR in this way and I will merge it into the master branch for the next major version of OTRS.

Thanks,
Marc

pboguslawski:
Hi,

> two additional settings (one for the editor, one for the mails) is too much, the DefaultCSS is redundant at this point.

Thanks for your comments. We'll stay with 3 separate settings; this way its possible to pack all common stuff (i.e. font name and size) to DefaultCSS and all the quirks specific to e-mail or web clients /editors to separate places which isn't that bad in my opinion. Additionally we have backward compatibility with existing DefaultCSS.

Regards,
Pawel

mgruner:
I'm closing this PR for now as no agreement was reached.

PR #1060, opened 2016-03-12
 fixed bug 11930 

ThomasMaven:
fix for error 500 while performing the search with "\" sign at the end - http://bugs.otrs.org/show_bug.cgi?id=11930


mgruner:
Hi @ThomasMaven, and thanks for your proposal. I applied a different fix, please see the bug entry for details. 

ThomasMaven:
Looks ok to me :+1:

PR #1063, opened 2016-03-15
Insert current time and version into the generated POT file

urbalazs:
Fixed <a href="http://bugs.otrs.org/show_bug.cgi?id=11892">BUG#11892</a>
However the header of the generated POT file is still wrong (see <a href="http://bugs.otrs.org/show_bug.cgi?id=11909">BUG#11909</a>).
This PR can be cherry-picked into REL_5-0, because those one is also affected.


mgruner:
Applied in b73eee81a0dfcd48f62adb0a61e123c40e1e2466 and backported. I omitted the Version because this would not work correctly with Packages. I will add some comments in the bug as well. Thanks!

PR #1064, opened 2016-03-15
Marked default public interface as translatable

urbalazs:
Hi @mgruner,
This is a small improvement for default public interface. This string is displayed when no modules with public interface are installed. This PR can be cherry-picked into REL_5-0, because the .tt file hasn't changed since then. OTRS 4 is also affected, but it has different template.


mgruner:
Merged and backported, thanks a lot!

PR #1065, opened 2016-03-16
Marked My Service tooltip as translatable

urbalazs:
Please, apply this PR to REL-5_0, too!


mgruner:

PR #1066, opened 2016-03-16
Marked title attributes in template files as translatable

urbalazs:
All <tt>title=""</tt> attributes are now translatable. Grep helps me to find them:
<kbd>grep --exclude=\*.{t,wadl} -rne 'title="[a-zA-Z]' ./</kbd>
If this PR is acceptable, then I will make an other one to rel-5_0.

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminProcessManagementActivityDialog.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminProcessManagementActivityDialog.tt
@@ -169,7 +169,7 @@
                             </ul>
                             <ul class="AllocationList Tablelike W90pc" id="AssignedFields">
 [% RenderBlockStart("AssignedFieldRow") %]
-                                <li data-id="[% Data.Field | html %]" data-entity="[% Data.Field | html %]" data-config="[% Data.FieldConfig | html %]" title="[% Translate("Name: %s", Data.Field) | html %]">
+                                <li data-id="[% Data.Field | html %]" data-entity="[% Data.Field | html %]" data-config="[% Data.FieldConfig | html %]" title="Name: [% Translate(Data.Field) | html %]">
                                     [% Translate(Data.Field) | html %]
                                     <span class="Functions ShowOnAssignedList">
                                         <a href="#" data-id="[% Data.Field | html %]" data-entity="[% Data.Field | html %]" class="FieldDetailsOverlay Icon" title="[% Translate("Edit") | html %]">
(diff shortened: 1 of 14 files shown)


var/httpd/htdocs/js/Core.Agent.Responsive.js
```diff
@@ -41,7 +41,7 @@ Core.Agent.Responsive = (function (TargetNS) {
         // Add trigger icon for pagination
         $('span.Pagination a:first-child').parent().closest('.WidgetSimple').each(function() {
             if (!$(this).find('.ShowPagination').length) {
-                $(this).find('.WidgetAction.Close').after('<div class="WidgetAction ShowPagination"><a title="Close" href=""><i class="fa fa-angle-double-right"></i></a></div>');
+                $(this).find('.WidgetAction.Close').after('<div class="WidgetAction ShowPagination"><a title="[% Translate("Close") | JSON %]" href=""><i class="fa fa-angle-double-right"></i></a></div>');
```
mgruner: This will not work in JS files. We usually place a text in the FooterJS via Config.Set() and then fetch it via Config.Get().

Kernel/Output/HTML/Templates/Standard/AdminProcessManagementActivityDialog.tt
```diff
@@ -169,7 +169,7 @@
                             </ul>
                             <ul class="AllocationList Tablelike W90pc" id="AssignedFields">
 [% RenderBlockStart("AssignedFieldRow") %]
-                                <li data-id="[% Data.Field | html %]" data-entity="[% Data.Field | html %]" data-config="[% Data.FieldConfig | html %]" title="Name: [% Translate(Data.Field) | html %]">
+                                <li data-id="[% Data.Field | html %]" data-entity="[% Data.Field | html %]" data-config="[% Data.FieldConfig | html %]" title="[% Translate("Name: %s", Data.Field) | html %]">
```
mgruner: This looks like a logical change. Before, Data.Field was translated, now, it is not translated any more. Not sure right now that it exactly contains. You could try something like ```[% Translate("Name: %s", Translate(Data.Field)) | html %]```. Do you test this stuff?

urbalazs:
Hi @mgruner 
I removed two modifications:
1.) From Core.Agent.Responsive.js - I haven't use Config.Set() and Config.Get() before, I have to learn to use these functions.
2.) From AdminProcessManagementActivityDialog.tt - I will test, what you suggested: <tt>[% Translate("Name: %s", Translate(Data.Field)) | html %]</tt>

mgruner:
I just merged the current state and backported to rel-5_0. Thanks!

PR #1069, opened 2016-03-18
Option to exclude customer from cc list for auto responses

reneeb:
One might want to disable that the customer is added to the Cc list of auto responses automatically. Imagine you have intensive communication with a supplier, then you don't want that your customer gets all those notifications that a follow up was created etc....

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10377,6 +10377,17 @@ Thanks for your help!
             </Option>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @reneeb. Thanks for sharing this; I do see your use case. However, OTRS is already over-configurable and I would like to avoid new config options. In this case what I could imagine is to decide depending on the article type ("internal"), if the customer CC should be suppressed, not globally based on a config option.
Please also add unit tests if you would consider this.
Best regards, Martin

PR #1074, opened 2016-03-20
Changed reply template selection to not include -Reply- anymore but a label instead

nilei:
The reply option always had -Reply- as option 0, which was not very user friendly as it was selectable and unselectable without any action triggered. Also under some circumstances, e.g. when there is a template with the name "!!! some template", it happened that these were listed before the -Reply- text.

Cheers, Nils

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -2847,22 +2847,13 @@ sub _ArticleMenu {
                 # use this array twice (also for Reply All), so copy it first
                 my @StandardResponseArrayReplyAll = @StandardResponseArray;
 
-                unshift(
-                    @StandardResponseArray,
-                    {
-                        Key   => '0',
-                        Value => '- '
-                            . $LayoutObject->{LanguageObject}->Translate('Reply') . ' -',
-                        Selected => 1,
-                    }
-                );
-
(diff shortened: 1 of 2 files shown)


nilei:
If this PR is valid, I will do the same for Forward and Move.

mgruner:
Hi @mrcbnsls, can you review this please? No backport, I guess.

mrcbnsls:
Hi @nilei,

This looks somehow odd:



With the old behavior, you would be able to see that you can actually do something there. I must admit that I generally don't like the idea of having a dropdown there. It just doesn't make sense for me, as it is a form field. I would prefer a mechanism similar to the cluster we introduced in TicketZoom. What do you think?

Marc

nilei:
Hi @mrcbnsls, I think the solution okay, considering we keep the dropdown.

Cluster? Yes and no! :-)

I actually don't like the select element, but with the modernized field, it provides a lot of comfort when you have a large number of templates. I don't think that we can simply add this search functionality into the cluster solution too.

mrcbnsls:
Ok, agreed. Merged in f8f4303b992280e1b77bb9b5b3b932db5b3e8494. Thanks!

nilei:
Thanks @mrcbnsls, I will send the PRs for Forward and Move soon.

PR #1075, opened 2016-03-21
Rel-5_0 - Fixed tests

s7design:
Hi @mgruner 

I saw on test server that there are some test which are failed. 
Herby I tried to fix them:

- News.t
- RSS.t
- OverviewSmall.t

If you would like it, you could merge it.

Regards
Zoran


mgruner:
The tests have been fixed in the meantime, thanks!

PR #1076, opened 2016-03-21
Fixed bug #11954

thorsteneckel:
Can't handle double quotes in option value - http://bugs.otrs.org/show_bug.cgi?id=11954

This currently works for us.

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -442,7 +442,7 @@ Core.UI.InputFields = (function (TargetNS) {
                     .data('value', Value);
 
                 // Textual representation of selected value
-                Text = $SelectObj.find('option[value="' + Value + '"]').first().text().trim();
+                Text = $SelectObj.find('option[value="' + Value.replace(/"/g, '\\"') + '"]').first().text().trim();
                 $TextObj = $('<div />').appendTo($SelectionObj);
                 $TextObj.addClass('Text')
                     .text(Text)
--- a/var/httpd/htdocs/js/Core.UI.InputFields.js
+++ b/var/httpd/htdocs/js/Core.UI.InputFields.js
@@ -442,7 +442,7 @@ Core.UI.InputFields = (function (TargetNS) {
                     .data('value', Value);
 
                 // Textual representation of selected value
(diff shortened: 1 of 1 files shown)


zottto:
Hi @thorsteneckel,

have you tried it with Core.App.EscapeSelector() instead of the regex?

Marc

thorsteneckel:
Thanks for the hint! Works like a charm.

zottto:
Merged with 796f4b6 (master) and aecf3aa (5).

thorsteneckel:
Hi @zottto !

Our UnitTests in one of our customer packages lead me to these two lines which cause the same error as fixed above:

https://github.com/OTRS/otrs/blob/796f4b6e43b1cbef24e8099a8900521c800c60d0/var/httpd/htdocs/js/Core.UI.InputFields.js#L1667

https://github.com/OTRS/otrs/blob/796f4b6e43b1cbef24e8099a8900521c800c60d0/var/httpd/htdocs/js/Core.UI.InputFields.js#L1671

While researching for further possible cases I found there where I'm not sure if it could cause trouble, too:

https://github.com/OTRS/otrs/blob/796f4b6e43b1cbef24e8099a8900521c800c60d0/var/httpd/htdocs/js/Core.UI.InputFields.js#L1073

https://github.com/OTRS/otrs/blob/796f4b6e43b1cbef24e8099a8900521c800c60d0/var/httpd/htdocs/js/Core.UI.InputFields.js#L1340

What do you think?
- Thorsten

zottto:
Hi @thorsteneckel,

looks like these occurences must be fixed, too. Do you mind sending an PR for them including updated unit test?

Marc

thorsteneckel:
HI @zottto 

thanks for your fast reply. To be honest I have no idea how to test these lines of code. I can only reproduce an error with our package and this line of code:

https://github.com/OTRS/otrs/blob/796f4b6e43b1cbef24e8099a8900521c800c60d0/var/httpd/htdocs/js/Core.UI.InputFields.js#L1671

But without the extension I don't know of any use case where to get to this line. This is basically caused by my lack of knowledge of Core.UI.InputFields and currently my motivation to get deep into that is very low.

I'll happily create a new PR with fixes for the lines but please understand that I'm not able to provide UnitTests for that.

Please let me know how we proceed.

Greetings,
  Thorsten

zottto:
@thorsteneckel Just send the PR without UT, I'll have a look at it ;-)

Thank you very much for your work!

Marc

PR #1077, opened 2016-03-21
Rel-5_0 - fixed tests SMIME/PGP

s7design:
Hi @mgruner, @mrcbnsls,

I tried to fix these tests, I am not sure if I succeeded in that. There is not problem on my local system with the tests, but I saw errors on the test server.

Regards
Zoran


mgruner:
The tests are working now, thank you!

PR #1078, opened 2016-03-21
Ambiguous labels while adding new GenericAgent job

urbalazs:
There are ambiguous labels on AdminGenericAgent screen. The first one refers to "NewCustomerUserLogin", but the label tells "New customer". It should be "New customer user" instead. In the second one the "ID" is unnecessary, because it is a user interface. "New customer" is enough. See the change in this screenshot:
genericagent_update
Please, also apply this PR on rel-5_0!


mgruner:
Hello @urbalazs, I fixed this slightly different in f1fcfaa4a133c0a53ba51ccb07c92fcd8b601727 and backported. This is now consistent also with AgentTicketCustomer. CustomerID is the right term because there is not neccessarily a customer (company) record behind it.

PR #1079, opened 2016-03-21
Configurable out of office message

pboguslawski:
Out of office strings in user name like

    User Name *** out of office till YYYY-D-M/2d ***

should be configurable not hardcoded english string.

According to https://github.com/OTRS/otrs/pull/1037
it cannot be translated easily in current OTRS User.pm
design so new SysConfig parameter OutOfOfficeMessageTemplate
was introduced to allow out of office message customization.

Fixes: cb8834d8
Related: https://dev.ib.pl/ib/otrs/issues/24
Author-Change-Id: IB#1004603


urbalazs:
Hi @mgruner , can you check this? This would solve the "out of office message" issue.

mgruner:
Merged with a slight wording change (until instead of to) in d1bd2a38730e05bf00c19253cee6fb141d80bccc and backported. This does not solve the real issue of the messages being not translatable, but for monolingual systems it should already be an improvement. Thanks!

PR #1080, opened 2016-03-22
CustomerNotificationsDisabled option added

pboguslawski:
This mod introduces new SysConfig parameter
CustomerNotificationsDisabled that allowes one to disable sending
notifications to customers with Ticket Notifications (extra
protection against customer spamming).

Related: https://dev.ib.pl/ib/otrs/issues/31
Author-Change-Id: IB#1050079


mgruner:
Hello Pawel. Thanks for your suggestion. I fail to see an important use case though as the recipients can already be configured on a per-notification level. Please let me know if I overlooked something.
Best regards, Martin

pboguslawski:
Hello Martin,

We use this mod if OTRS application admins should not be allowed to enable stuff that automatically generates e-mails to customers (we hardcore this setting in Config.pm for example). May be useful for other OTRS users i think. Ignore it if you don't agree :)

Regards,
Pawel

PR #1081, opened 2016-03-22
NotificationEmailDefaultTemplate option added

pboguslawski:
This mod introduces new SysConfig parameter
```
NotificationEmailDefaultTemplate
```
that allowes one to specify default template (from
Kernel/Output/HTML/Templates/Standard/NotificationEvent/Email
directory) to use when sending notification e-mails.

Related: https://dev.ib.pl/ib/otrs/issues/33
Author-Change-Id: IB#1050091

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-22 NotificationEmailDefaultTemplate option added.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Pawel. What's the use case for this? There is only one global Default template per system, even with your new config option. So you could also just override the default template file in the Custom/ directory and put a customized template there, right? I don't see right now how an additional config setting could be beneficial here. Please explain.

pboguslawski:
Hi,

> So you could also just override the default template file in the Custom/ directory and put a customized template there, right?

We wanted to have own custom template with colors matching our skin and not to overwrite Default one and allow admin to choose which skin and template will be used in system as default, without messing at filesystem level. Thats why this mod.

Regards,
Pawel

mgruner:
Thanks for the explanation. Then I would not merge it at this point. Best regards, Martin

PR #1083, opened 2016-03-23
[rel-5_0] Ported back translatable messages from master - part4

urbalazs:
Hi @mgruner 
This PR contains backported translatable messages from master. In this part only Kernel/Config/Files/*.xml files are included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in 8052076af5179c8c0cac7533c59ec77195318fec, thanks!

PR #1084, opened 2016-03-24
[rel-5_0] Ported back translatable messages from master - part5

urbalazs:
Hi @mgruner
This PR contains backported translatable messages from master. In this part only Kernel/Modules/*.pm files are included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in 35e25dd3a440c405e690da5c0be95882cdb31a9b, thanks!

PR #1086, opened 2016-03-24
[rel-5_0] Ported back translatable messages from master - part6

urbalazs:
Hi @mgruner
This PR contains backported translatable messages and fixed comments from master. In this part only Kernel/Config/Defaults.pm is included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in 509287d630eff351982483720a8997c6eba14717 without the "use" statement, I also removed this from master.

PR #1087, opened 2016-03-24
[rel-5_0] Ported back translatable messages from master - part7

urbalazs:
Hi @mgruner
This PR contains backported translatable messages from master. In this part only Kernel/Output/HTML/* files are included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in db75a2d93418b416ba351aa8c3aab9905b273890. Thanks!

PR #1088, opened 2016-03-24
[rel-5_0] Ported back translatable messages from master - part8

urbalazs:
Hi @mgruner
This PR contains backported translatable messages from master. In this part only Kernel/System/* files are included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in 855925b9c07ca8acfec65472a19d8a635e2a36bf, thanks.

PR #1089, opened 2016-03-24
[rel-5_0] Ported back translatable messages from master - part9

urbalazs:
Hi @mgruner
This PR contains backported translatable messages from master. In this part only Kernel/Language.pm and scripts/database/otrs-initial_insert.xml files are included. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Merged in ee099799ed18a47db4d29b15c0e40b33a344c8ec, thanks!

PR #1091, opened 2016-03-25
OTRS Console Command "Dev::Tools::Shell" - a OTRS REPL

thorsteneckel:
Hi there,

as discussed with @mgruner I want to share our approach for a REPL / shell for the OTRS context as a OTRS Console command. It is based on the Perl module "Devel::REPL" which can be installed via CPAN if you want to use this feature. I've added it as optional to the "otrs.CheckModules.pl" script. If not installed you will get an error message telling you about it, see https://github.com/OTRS/otrs/pull/1091/files#diff-18f4874b62331c044aef188c6fd8fcb4R35.

Here are some basic example screenshots:

Initializing the TicketObject:
bildschirmfoto 2016-03-25 um 17 25 51

Getting the Ticket with TicketID 1:
bildschirmfoto 2016-03-25 um 17 26 19

The use case would be to avoid writing a "whole" test script just for executing one command in the OTRS context. Classic scenarios would be debugging or testing. Looking inside the DB or adding debug lines can cause false information because some functions like ACLs might not be executed yet. 

I had some issues writing tests for the command since it keeps in the state of waiting for user input. I was not able to "kill" the process or exit the loop without exiting the test completely. I'm open for proposals to write proper tests.

Additionally I created a custom Devel::REPL plugin for formatting purposes which is based on the default DDP plugin. It basically adds a newline before and after the output and tries to determine the last variable type for proper formatting. See https://github.com/OTRS/otrs/pull/1091/files#diff-7345d00a44c8d5e75deff2b36b73072aR28 .

That should be all. If you have any questions or feedback I'm glad to hear it.

  Thorsten

### DIFF
--- a/Kernel/System/Console/Command/Dev/Tools/Shell.pm
+++ b/Kernel/System/Console/Command/Dev/Tools/Shell.pm
@@ -28,11 +28,20 @@ sub Configure {
 sub PreRun {
     my ( $Self, %Param ) = @_;
 
-    return if $Kernel::OM->Get('Kernel::System::Main')->Require(
-        'Devel::REPL',
-    );
+    my @Dependencies = ('Devel::REPL', 'Data::Printer');
 
-    die "Required Perl module 'Devel::REPL' not found. Please install and retry.";
+    DEPENDENCY:
+    for my $Dependency ( @Dependencies ) {
+
+        next DEPENDENCY if $Kernel::OM->Get('Kernel::System::Main')->Require(
+            $Dependency,
+            Silent => 1,
+        );
+
+        die "Required Perl module '$Dependency' not found. Please make sure the following dependencies are installed: ". join(' ', @Dependencies);
+    }
+
+    return 1;
 }
 
 sub Run {
--- a/Kernel/System/Console/Command/Dev/Tools/Shell.pm
+++ b/Kernel/System/Console/Command/Dev/Tools/Shell.pm
@@ -1,5 +1,5 @@
 # --
-# Copyright (C) 2001-2015 OTRS AG, http://otrs.com/
+# Copyright (C) 2001-2016 OTRS AG, http://otrs.com/
 # --
 # This software comes with ABSOLUTELY NO WARRANTY. For details, see
 # the enclosed file COPYING for license information (AGPL). If you
@@ -22,23 +22,32 @@ sub Configure {
 
     $Self->Description('An interactive REPL shell for the OTRS context.');
 
+    $Self->AddOption(
+        Name        => 'eval',
+        Description => 'Perl code that should get evaluated and printed in the OTRS context.',
+        Required    => 0,
+        HasValue    => 1,
+        ValueRegex  => qr/.*/smx,
+    );
+
     return;
 }
 
 sub PreRun {
     my ( $Self, %Param ) = @_;
 
-    my @Dependencies = ('Devel::REPL', 'Data::Printer');
+    my @Dependencies = ( 'Devel::REPL', 'Data::Printer' );
 
     DEPENDENCY:
-    for my $Dependency ( @Dependencies ) {
+    for my $Dependency (@Dependencies) {
 
         next DEPENDENCY if $Kernel::OM->Get('Kernel::System::Main')->Require(
             $Dependency,
             Silent => 1,
         );
 
-        die "Required Perl module '$Dependency' not found. Please make sure the following dependencies are installed: ". join(' ', @Dependencies);
+        die "Required Perl module '$Dependency' not found. Please make sure the following dependencies are installed: "
+            . join( ' ', @Dependencies );
     }
 
     return 1;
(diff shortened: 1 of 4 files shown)


mgruner:
Hi @thorsteneckel, thanks so much! One question, did you also look at the CPAN module ```Reply```? Why did you favor ```Devel::REPL``` over it?

thorsteneckel:
Thanks for taking this to the framework. I appreciate it.

Yes I actually evaluated using `Reply` as you proposed. It seems to me that it might be a proper alternative sometime but is currently in an early stage. The  [cpan page](https://metacpan.org/pod/Reply) notes:
`NOTE: This is an early release, and implementation details of this module are still very much in flux. Feedback is welcome!`

`Devel::REPL` comes with a multiline plugin that I have activated. It makes it possible to support code blocks that stretch over more than one line. I could find anything similar for `Reply` which was pretty much a show stopper for me. There were some minor differences which weren't that important to me (output, fancy prompt, the time I had spend with `Devel::REPL` ;) ) 

Since the dependency structure for both modules are not that great I decided to stick with `Devel::REPL` over `Reply`.

If there are any unanswered questions let me know.

I see that the TravisCI tests are failing. I have not much experience with the output but it seems to me that some notification tests are failing. I think that has nothing to do with my pull request? Let me know if I'm wrong.

Kernel/System/Console/Command/Dev/Tools/Shell.pm
```diff
@@ -0,0 +1,72 @@
...
+    $Self->Description('An interactive REPL shell for the OTRS context.');
+
+    return;
+}
+
+sub PreRun {
+    my ( $Self, %Param ) = @_;
+
+    return if $Kernel::OM->Get('Kernel::System::Main')->Require(
+        'Devel::REPL',
```
mgruner: You can add Silent => 1 here to suppress the error messages if not installed.

bin/otrs.CheckModules.pl
```diff
@@ -455,6 +455,16 @@
             zypper => 'perl-YAML-LibYAML',
         },
     },
+    {
```
mgruner: I'm not sure if we need this. We also have other dev deps that aren't listed here (at least Locale::PO). The message from the command should be a good start, what do you think?

mgruner:
Regarding the testability, how about adding a new ```--eval``` or ```--code``` parameter that would allow passing in code from a unit test or script?

mgruner:
After installing Devel::REPL via cpanminus, I get this error:
```Error: Failed to load role: Devel::REPL::Plugin::OTRS Can't locate Data/Printer.pm in @INC (you may need to install the Data::Printer module) (@INC contains: /ws/otrs6/Custom /ws/otrs6/Kernel/cpan-lib /ws/otrs6 /Library/Perl/5.18/darwin-thread-multi-2level /Library/Perl/5.18 /Network/Library/Perl/5.18/darwin-thread-multi-2level /Network/Library/Perl/5.18 /Library/Perl/Updates/5.18.2/darwin-thread-multi-2level /Library/Perl/Updates/5.18.2 /System/Library/Perl/5.18/darwin-thread-multi-2level /System/Library/Perl/5.18 /System/Library/Perl/Extras/5.18/darwin-thread-multi-2level /System/Library/Perl/Extras/5.18 .) at /ws/otrs6/Kernel/cpan-lib/Devel/REPL/Plugin/OTRS.pm line 16.
```

thorsteneckel:
Thanks for your feedback. I will work on it tomorrow.

bin/otrs.CheckModules.pl
thorsteneckel: I'm not sure what the disadvantages of this are or is your concern that users will be confused? From my perspective as an customizer I see no reason to remove it but I'm full aware of my lack of framework knowledge.

thorsteneckel: Changed with: https://github.com/OTRS/otrs/pull/1091/commits/f1f00908bc13ef67c40a31af27bdee5fa5724c16

thorsteneckel:
As you can see I reverted the changed in `otrs.CheckModules.pl` - https://github.com/OTRS/otrs/pull/1091/commits/f1f00908bc13ef67c40a31af27bdee5fa5724c16.

Additionally I improved the dependency check with - https://github.com/OTRS/otrs/pull/1091/commits/25562e9789f0ff67be67a47eb9ea20cfc91d2e2f.

I haven't found anything to pass code from the outside scope to the REPL instance. There is no parameter as far as I can say.

mgruner:
Hi @thorsteneckel, you could try running $Repl->formatted_eval() on the code passed in to avoid entering the REPL loop. Maybe then pass the result to $Repl->print().

mgruner:
That's basically what run_once does, just without reading.

thorsteneckel:
Hi @mgruner. Done! what do you think about my changes?

mgruner:
Thanks a lot for your update, Thorsten, very nice. 

There is one thing I strongly dislike: the determination of the variable type based on history. This is ugly and error prone (try `my %Test = (1 => 2); my @List = (1, 3, 4)`).

My suggestion would be to try to guess if the list could be a hash. Please try this: https://gist.github.com/mgruner/a72d906636eb7d4c565ebb666812436a

You need to load first the original DDP plugin and then the OTRS plugin. This avoids the need to patch DDP.pm. What do you think?

The drawback is that everything that *could be* a hash will be printed as one, but I believe this is not too bad.

thorsteneckel:
Thanks for your Feedback, Martin.

I like your proposal and will work it into my PR in the next free timeslot.

mgruner:
Hi Thorsten,

that’s not neccessary. I already have a local branch that only requires cleanup and merging.

Best regards,

Martin Gruner
Team Lead R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

mgruner:
Merged in 9b579b6105a8bf05440713ec833f6c84b83cf8ea. Cool, @thorsteneckel. Please let me know if this works for you.

PR #1094, opened 2016-03-29
Ported translatable messages from rel-5_0

urbalazs:
Hi @mgruner 
This PR contains messages from rel-5_0. These are missing from language file of OTRS master. I found them with diff, so each messages exactly the same in rel-5_0. If you disagree with some changes, it should be fixed in rel-5_0, too.


mgruner:
Merged in master, thanks a lot!

PR #1095, opened 2016-03-29
Marked a notification as translatable

urbalazs:
This also affects rel-5_0.


mgruner:

PR #1096, opened 2016-03-30
Improved SysConfig edit header translation

urbalazs:
The word "in" was no translatable. This also affects rel-5_0.


mgruner:

PR #1098, opened 2016-03-30
Marked messages as translatable

urbalazs:
Hi @mgruner 
These strings are not marked as translatable neither in rel-5_0 nor in master.


mgruner:

PR #1099, opened 2016-03-30
Set X-Mailer to undef if "Secure::DisableBanner" is active

reneeb:
If "Secure::DisableBanner" is active and the X-Mailer header isn't set to undef, MIME::Tools adds its own X-Mailer header. There are some mail servers that handles mails sent by MIME::Tools as spam mails (as MIME::Tools is seen as a program that is sometimes used to send mass mails).


mgruner:
Hello Renée, merged with slight changes in f05a1665b36aef9bffa846d7045347375450e89d and backported to rel-5_0. I also added unit tests for this. Could you please consider always doing so in future when you send contributions? I guess that it's a normal practice and quality standard these days.
Thanks a lot for your contribution!
Best regards, Martin

PR #1100, opened 2016-03-30
Improved widget close icon title

urbalazs:
Hi @mgruner 

This PR has two purpose:
1.) According to <a href="http://otrs.github.io/doc/manual/developer/stable/en/html/accessibility.html#id-1.6.8.5.2">developer manual</a>, it should be used <q><em>real, understandable and precise sentences</em></q>. In the manual there is exactly the same example: <em>"Close this widget" is much better than "Close"</em>.
2.) In some languages (e. g. in Hungarian) there are different words for closing a ticket (in this case "close" is "<strong>le</strong>zárás" in Hungarian) and closing a window or widget (in this case "close" is "<strong>be</strong>zárás" in Hungarian). Because of the structure of the .po file, there is not possible to translate the two kind of "close" differently, as it contains only one instance of "close".

This also affects rel-5_0.


mgruner:
Excellent, I fully agree. Merged and backported.

PR #1093, opened 2016-03-29
[rel-5_0] Tidied, small translation updates

urbalazs:
Hi @mgruner
This is the final part of backporting messages from master. This PR includes mostly tiding, but it also contains small translation updates (renamed "Submit" buttons). I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.

PR #1101, opened 2016-03-31
[rel-5_0] Ported back translatable messages from master - part10

urbalazs:
Hi @mgruner 
I removed the non-relevant changes from https://github.com/OTRS/otrs/pull/1093. This PR contains backported translatable messages from master. I found them with diff, so each messages exactly the same in master. If you disagree with some changes, it should be fixed in master, too.


mgruner:
Partially merged in f5fdf6b. The XML changes were not really needed, the change in Layout.pm would break things, it is a bugfix made only in master. There is some code in DBUpdate-to-6.pl to make sure that existing user settings will continue to work. Thanks!

PR #1104, opened 2016-04-01
Fixed bug #11973: Mappin Simple can't get initialized multiple times.

thorsteneckel:
Hi there,

here is my proposal for fixing the following bug http://bugs.otrs.org/show_bug.cgi?id=11973

Greetings,
  Thorsten

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-03-29 Added possibility to configure default headers for outgoing emails (Sendmail::DefaultHeaders), thanks to Renée Bäcker.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
  - 2016-03-03 Fixed bug#[11872](http://bugs.otrs.org/show_bug.cgi?id=11872) - TicketGet function returns SolutionTime variable.
  - 2016-03-03 Fixed bug#[8631](http://bugs.otrs.org/show_bug.cgi?id=8631) - "ghost" tickets after merge.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
  - 2015-11-27 Improved command Maint::Ticket::InvalidUserCleanup. It can now now both unlock tickets of invalid users and also (optionally) change their state to make sure they will not be overlooked. Thanks to Moritz Lenz @ noris networks.
 
 #5.0.9 2016-??-??
+ - 2016-03-28 Fixed bug#[11957](http://bugs.otrs.org/show_bug.cgi?id=11957) - Inform agents not working, the agent who is different to owner not received notification.
  - 2016-03-22 Fixed bug#[11836](http://bugs.otrs.org/show_bug.cgi?id=11836) - Pending date selection impossible due ACL limitation.
  - 2016-03-22 Fixed bug#[11956](http://bugs.otrs.org/show_bug.cgi?id=11956) - locked tickets for invalid users are not shown in support data collector.
  - 2016-03-22 Fixed bug#[11954](http://bugs.otrs.org/show_bug.cgi?id=11954) - Can't handle double quotes as option value in modernized InputFields, thanks to Thorsten Eckel.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-01 Fixed dropdown for CustomerTicketOverview.
  - 2016-04-01 Changed reply template selection to not include -Reply- anymore but a label instead, thanks to Nils Leideck.
  - 2016-03-29 Added possibility to configure default headers for outgoing emails (Sendmail::DefaultHeaders), thanks to Renée Bäcker.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,9 @@
  - 2015-11-27 Improved Generic Agent performance at deleting old execution times. Thanks to Moritz Lenz @ noris networks.
  - 2015-11-27 Improved command Maint::Ticket::InvalidUserCleanup. It can now now both unlock tickets of invalid users and also (optionally) change their state to make sure they will not be overlooked. Thanks to Moritz Lenz @ noris networks.
 
-#5.0.9 2016-??-??
+#5.0.10 2016-??-??
+
+#5.0.9 2016-04-05
  - 2016-03-30 Updated translations, thanks to all translators.
  - 2016-03-28 Fixed bug#[11957](http://bugs.otrs.org/show_bug.cgi?id=11957) - Inform agents not working, the agent who is different to owner not received notification.
  - 2016-03-22 Fixed bug#[11836](http://bugs.otrs.org/show_bug.cgi?id=11836) - Pending date selection impossible due ACL limitation.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@
  - 2015-11-27 Improved command Maint::Ticket::InvalidUserCleanup. It can now now both unlock tickets of invalid users and also (optionally) change their state to make sure they will not be overlooked. Thanks to Moritz Lenz @ noris networks.
 
 #5.0.9 2016-??-??
+ - 2016-03-30 Updated translations, thanks to all translators.
  - 2016-03-28 Fixed bug#[11957](http://bugs.otrs.org/show_bug.cgi?id=11957) - Inform agents not working, the agent who is different to owner not received notification.
  - 2016-03-22 Fixed bug#[11836](http://bugs.otrs.org/show_bug.cgi?id=11836) - Pending date selection impossible due ACL limitation.
  - 2016-03-22 Fixed bug#[11956](http://bugs.otrs.org/show_bug.cgi?id=11956) - locked tickets for invalid users are not shown in support data collector.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@
  - 2015-11-27 Improved command Maint::Ticket::InvalidUserCleanup. It can now now both unlock tickets of invalid users and also (optionally) change their state to make sure they will not be overlooked. Thanks to Moritz Lenz @ noris networks.
 
 #5.0.10 2016-??-??
+ - 2016-03-30 Fixed bug#[11922](http://bugs.otrs.org/show_bug.cgi?id=11922) - Notification to non RealCustomer.
 
 #5.0.9 2016-04-05
  - 2016-03-30 Updated translations, thanks to all translators.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-01 Changed reply template selection to not include -Reply- anymore but a label instead, thanks to Nils Leideck.
  - 2016-03-29 Added possibility to configure default headers for outgoing emails (Sendmail::DefaultHeaders), thanks to Renée Bäcker.
  - 2016-03-04 Fixed bug#[11787](http://bugs.otrs.org/show_bug.cgi?id=11787) - No Ticket::StateAfterPending found with manual state update.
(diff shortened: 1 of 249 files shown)


carlosfrodriguez:
Hi @thorsteneckel,

Thank you for you pull request.

I'm not sure about this one, Apparently the mapping module changes the references of the mapping configuration internally, then when you instantiate for the second time, the configuration does not have the correct information. Looking at the POD the missing information is needed, so I try this:

Create a local copy of the Test scenario as

```perl
TEST:
for my $Test (@MappingTests) {

    my $OrigTest = Storable::dclone($Test);

    my $StartSeconds;

```

And then pase your new tests but using the cloned information 

```perl
    if ( !$Test->{ResultSuccess} ) {
        $Self->True(
            $MappingResult->{ErrorMessage},
            $Test->{Name} . ' error message found',
        );
    }

     my $SecondMappingObject = Kernel::GenericInterface::Mapping->new(
         DebuggerObject => $DebuggerObject,
         MappingConfig  => {
             Type   => 'Simple',
            Config => $OrigTest->{Config},
         },
     );

     $Self->Is(
         ref $SecondMappingObject,
         'Kernel::GenericInterface::Mapping',
         'SecondMappingObject was correctly instantiated',
     );
}    # end test
```

And it seams to work correctly, so it might be possible the issue you found is in deed a problem with the way you are testing this, please correct me if i'm wrong.

What do you think?

((enjoy))
cr

carlosfrodriguez:
When the constructor does not return the object (but a hash) it is catch by the requester and provider correctly (it contains the error messages about why the object was not created).

```perl
        # if mapping init failed, bail out
        if ( ref $MappingInObject ne 'Kernel::GenericInterface::Mapping' ) {
            $DebuggerObject->Error(
                Summary => 'MappingIn could not be initialized',
                Data    => $MappingInObject,
            );

            return $Self->_GenerateErrorResponse(
                DebuggerObject => $DebuggerObject,
                ErrorMessage   => $FunctionResult->{ErrorMessage},
            );
        }
```

thorsteneckel:
Hi @carlosfrodriguez , thanks for your fast and comprehensive feedback. You see it's not only the provided extended UnitTest file that's failing, we had a concrete case where something like this was done in an EventModule:

```perl

for my $Action ( @Actions ) {
    my $Result = $RequesterObject->Run(
        WebserviceID => $Webservice->{ID},
        Invoker      => 'SomeInvokerName',
        Data         => {
            TicketID => $Param{Data}->{TicketID},
            Event    => $Param{Event},
            Action   => $Action,
        },
    );

    # error handing etc.
    ...
}
```

As you will see  the error occurs that have been reproduced with the given extended UnitTest. The same Invoker of the same Webservice will get executed two times right after another therefore the Mapping backend will get initialized two times with the same config where the first Mapping backend will work fine and the second one will fail like in the unit test.

Your feedback gave me the idea to prevent this more elegantly with the usage of Storable::dclone which I did in the last commit.

Please let me know if things are clearer now.

Thanks!
  Thorsten

carlosfrodriguez:
Hi @thorsteneckel,

This update looks for sure much more elegant, I might suggest another change..., but before continue with this the question is: why would you like to create another instance of the same mapping with the same data?

What is the use case for that?

In Requester and Provider we have 2 instances of the mapping working correctly as they both have different and fresh data (inbound/outbound).

((enjoy))
cr

thorsteneckel:
Hi Carlos! Thanks for your feedback.

Like written above we have something like this in a Ticket event module:

```perl
for my $Action ( @Actions ) {
    my $Result = $RequesterObject->Run(
        WebserviceID => $Webservice->{ID},
        Invoker      => 'SomeInvokerName',
        Data         => {
            TicketID => $Param{Data}->{TicketID},
            Event    => $Param{Event},
            Action   => $Action,
        },
    );

    # error handing etc.
    ...
}
```

As you can see depending on the content of the Actions Array the Requester might get executed multiple times. The first time everything is fine. The request gets performed etc. without any issues. The second time the magic happens:

Here ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/GenericInterface/Requester.pm#L117 ):
```perl
    my $Webservice = $Kernel::OM->Get('Kernel::System::GenericInterface::Webservice')->WebserviceGet(
        ID => $WebserviceID,
    );
```

The same webservice is get the second time. As you can see we have a Hash**Reference**. The ref comes from the Cache ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/System/GenericInterface/Webservice.pm#L246
 ):

```perl
    my $Cache = $CacheObject->Get(
        Type => 'Webservice',
        Key  => $CacheKey,
    );
    return $Cache if $Cache;
````

Therefore the Content of $Webservice is the same for first request (fine) and second request (bad). 

Lets go further. Now since we are in the Requester.pm only the Requester part of the $Webservice config ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/GenericInterface/Requester.pm#L135 ) is interesting for us:

```perl
    my $RequesterConfig = $Webservice->{Config}->{Requester};
```
Attention: **Reference** again!

Now later when the first mapping backend (Simple) is initialized the gathered $RequesterConfig is passed to the mapping backend constructor ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/GenericInterface/Requester.pm#L231 ), ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/GenericInterface/Mapping.pm#L126 ), ( https://github.com/OTRS/otrs/blob/1e908159a5dbdcfb94cc35d13bf15b04ac3e6a24/Kernel/GenericInterface/Mapping/Simple.pm#L56 ):

```perl
        my $MappingOutObject = Kernel::GenericInterface::Mapping->new(
            DebuggerObject => $DebuggerObject,
            Invoker        => $Param{Invoker},
            InvokerType    => $RequesterConfig->{Invoker}->{ $Param{Invoker} }->{Type},
            MappingConfig =>
                $RequesterConfig->{Invoker}->{ $Param{Invoker} }->{MappingOutbound},
        );
```

Notice: Still our little **Reference** friend.

But now the noted keys get changed from `undef` to an empty hash ref (see above) which causes the error in the second execution run of my example Actions array loop. And as you can hopefully see now both references are the same.

This same behavior applies to the second instantiation of of the mapping object too as you already wrote. The difference is that both use different keys: `MappingOutbound` and `MappingIngoing`. Therefore they can't cause this issue in the first requester call.

I think maybe the best solution would be to use Storable::dclone in the `WebserviceGet` function? I'm looking forward to hear your proposal. If there are any questions remaining, please let me know.

carlosfrodriguez:
Hi @thorsteneckel,

I think I see your point, there is something wrong in *Simple.pm*, analyzing your commits in this PR, I was intrigued of what is exactly wrong in this file and if the configuration really needs to be manipulated. This are my findings: 

ValueMap is not marked in the POD as required or optional (normally in this cases this should be considered required) but in the code there is a place where suggest that it is not:

```perl
    my %RequiredConfigTypes = (
        KeyMapDefault   => 1,
        ValueMapDefault => 1,
    );
    CONFIGTYPE:
    for my $ConfigType (qw(KeyMapExact KeyMapRegEx KeyMapDefault ValueMap ValueMapDefault)) {

        # require some types
        if ( !defined $Config->{$ConfigType} ) {
            next CONFIGTYPE if !$RequiredConfigTypes{$ConfigType};
            return $Self->{DebuggerObject}->Error(
                Summary => "Got no $ConfigType, but it is required!",
            );
        }
```

This is also consistent with the GUI, where New Value map: can be skipped. Then the POD might need an update :-)

Unfortunately there are 2 places in the code where (I think unintentionally) the ValueMap key is created as an empty hash reference if it does not exits, and I believe this is wrong. Can you please take a look on the following patch and let me know what do you think about it? does it works for you?

```diff
diff --git a/Kernel/GenericInterface/Mapping/Simple.pm b/Kernel/GenericInterface/Mapping/Simple.pm
index 3c38210..a2286af 100644
--- a/Kernel/GenericInterface/Mapping/Simple.pm
+++ b/Kernel/GenericInterface/Mapping/Simple.pm
@@ -243,7 +243,11 @@ sub Map {
         }
 
         # check if we have a value mapping for the specific key
-        my $ValueMap = $Config->{ValueMap}->{$NewKey};
+        my $ValueMap;
+        if ( $Config->{ValueMap} && $Config->{ValueMap}->{$NewKey} ) {
+            $ValueMap = $Config->{ValueMap}->{$NewKey}
+        }
+
         if ($ValueMap) {
 
             # first check in exact (1:1) map
@@ -413,42 +417,44 @@ sub _ConfigCheck {
     }
 
     # check ValueMap
-    for my $KeyName ( sort keys %{ $Config->{ValueMap} } ) {
-
-        # require values to be hash ref
-        if ( !IsHashRefWithData( $Config->{ValueMap}->{$KeyName} ) ) {
-            return $Self->{DebuggerObject}->Error(
-                Summary => "Got $KeyName in ValueMap, but it is no hash ref with content!",
-            );
-        }
+    if ( $Config->{ValueMap} ) {
+        for my $KeyName ( sort keys %{ $Config->{ValueMap} } ) {
 
-        # possible sub-values are ValueMapExact or ValueMapRegEx and need to be hash ref if defined
-        SUBKEY:
-        for my $SubKeyName (qw(ValueMapExact ValueMapRegEx)) {
-            my $ValueMapType = $Config->{ValueMap}->{$KeyName}->{$SubKeyName};
-            next SUBKEY if !defined $ValueMapType;
-            if ( !IsHashRefWithData($ValueMapType) ) {
+            # require values to be hash ref
+            if ( !IsHashRefWithData( $Config->{ValueMap}->{$KeyName} ) ) {
                 return $Self->{DebuggerObject}->Error(
-                    Summary =>
-                        "Got $SubKeyName in $KeyName in ValueMap,"
-                        . ' but it is no hash ref with content!',
+                    Summary => "Got $KeyName in ValueMap, but it is no hash ref with content!",
                 );
             }
 
-            # key/value pairs of ValueMapExact and ValueMapRegEx must be strings
-            for my $ValueMapTypeKey ( sort keys %{$ValueMapType} ) {
-                if ( !IsString($ValueMapTypeKey) ) {
+            # possible sub-values are ValueMapExact or ValueMapRegEx and need to be hash ref if defined
+            SUBKEY:
+            for my $SubKeyName (qw(ValueMapExact ValueMapRegEx)) {
+                my $ValueMapType = $Config->{ValueMap}->{$KeyName}->{$SubKeyName};
+                next SUBKEY if !defined $ValueMapType;
+                if ( !IsHashRefWithData($ValueMapType) ) {
                     return $Self->{DebuggerObject}->Error(
                         Summary =>
-                            "Got key in $SubKeyName in $KeyName in ValueMap which is not a string!",
+                            "Got $SubKeyName in $KeyName in ValueMap,"
+                            . ' but it is no hash ref with content!',
                     );
                 }
-                if ( !IsString( $ValueMapType->{$ValueMapTypeKey} ) ) {
-                    return $Self->{DebuggerObject}->Error(
-                        Summary =>
-                            "Got value for $ValueMapTypeKey in $SubKeyName in $KeyName in ValueMap"
-                            . ' which is not a string!',
-                    );
+
+                # key/value pairs of ValueMapExact and ValueMapRegEx must be strings
+                for my $ValueMapTypeKey ( sort keys %{$ValueMapType} ) {
+                    if ( !IsString($ValueMapTypeKey) ) {
+                        return $Self->{DebuggerObject}->Error(
+                            Summary =>
+                                "Got key in $SubKeyName in $KeyName in ValueMap which is not a string!",
+                        );
+                    }
+                    if ( !IsString( $ValueMapType->{$ValueMapTypeKey} ) ) {
+                        return $Self->{DebuggerObject}->Error(
+                            Summary =>
+                                "Got value for $ValueMapTypeKey in $SubKeyName in $KeyName in ValueMap"
+                                . ' which is not a string!',
+                        );
+                    }
                 }
             }
         }
```

It looks like a lot of changes but they are really just 2 if conditions.

I would not like to go to the storable solution because I would like to keep all Mapping modules under the same basic principles so if we use storable in *Simple.pm* then we should use storable in all the others, but if the others does not have the same issue, then it might worth the effort to prevent the issue in *Simple.pm*

If you agree with this solution can I ask you to update the pull request adding this changes, the unit tests enhancements and the fix in the POD? could be possible?

((enjoy))
cr

thorsteneckel:
Sorry I broke it :) I'll create a new pull request for that.

PR #1105, opened 2016-04-03
auth: Handle Sync before TwoFactor-Auth. possible fix of #11966

prauscher:
some TwoFactor-Modules require a set UserID-field to gather their settings.
this will sync users even when their TwoFactor-Auth would have failed, but
their primary Password lookup must have been successful.


sb-otrs:
Thanks for the pull request.
I've decided not to use it directly but to modify the code with slight changes myself.

PR #1106, opened 2016-04-03
Twofactor QR

prauscher:
Adds two Buttons to GoogleAuthenticator-Pane in AgentPreferences:
* `Generate Secret` generates a new secret which is base32-compatible
* `Show QR-Code` opens a new window to show a QR-Code to be imported directly to your phone

Parts of this pr could probably be nicer, this is my first "major" pr to OTRS, so I'm open to learn :)


prauscher:
Sorry for the mess, but the actual changes are understandable I think. Generally I'd suggest to remove the input-field there fully and make changes via the JS-API only to make sure the Shared Secret does not get transfered too often.

prauscher:
I will redo this PR to provide a cleaner and better solution

PR #1108, opened 2016-04-04
Marked TimeInputFormat types as translatable

urbalazs:
Hi @mgruner 
This small PR makes translatable the types of TimeInputFormat in SysConfig. Branch rel-5_0 is also affected.


mgruner:
Merged, thanks!

PR #1110, opened 2016-04-04
Update Core.UI.Autocomplete.js

capeit-tto:
Modification fixes use of CI-attribute type "Customer" if used as sub-attribute. If not applied, auto complete does not start because of JS error on ids which contain double-colons.

PS: check if update required on HEAD/MASTER as well


mgruner:
Looks nice, thanks!

mrcbnsls:
Thanks! Merged in b5eddaa and e02e294.

PR #1111, opened 2016-04-04
Marked Installer messages as translatable

urbalazs:
Hi @mgruner 
I found some messages in the installer, which are not marked as translatable. Branch rel-5_0 is also affected.


mgruner:

PR #1114, opened 2016-04-06
Fixed translatable messages to display translations

urbalazs:
Hi @mgruner 
This messages have complex HTML structure, when they are displayed on the notification bar, so they need to translated in place. Otherwise they remains untranslated.

### DIFF
--- a/Kernel/Modules/AdminPackageManager.pm
+++ b/Kernel/Modules/AdminPackageManager.pm
@@ -527,7 +527,7 @@ sub Run {
             $Output .= $LayoutObject->Notify(
                 Priority => 'Error',
                 Data     => "$Name $Version - "
-                    . Translatable(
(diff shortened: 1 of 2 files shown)


mgruner:
Can you try switching from ```Data => ...``` to ```Info => ...```? I would think that it should work then.

mgruner:
There is actually no HTML included here so that ```Data``` would be needed, or do I miss something?

urbalazs:
Hi @mgruner 
Thanks for the suggestion, changing "Data" to "Info" in the file UIDCheck.pm makes the string translatable in the interface. But in the AdminPackageManager.pm there are string concatenations, so only Translate() works there.

mgruner:
Right! Merged and backported, thanks!

PR #1115, opened 2016-04-06
Removed dummy 'Reply All' and 'Forward' options to align with 'Reply' select.

nilei:
The "Reply All" and "Forward" selects are now updated to align with the changed "Reply" field.


mrcbnsls:
Thanks, merged in e47ae13.

PR #1118, opened 2016-04-08
[Rel-5_0] - UnitTest - CustomerRealName - bug 4640

s7design:
Hello @mrcbnsls ,

Hereby is extended unit test for CustomerRealName from the bug #795 . There was an issue when $From needs to be taken from Recipients, param RealName doesn't exists in Recipients. Corrected typo to param Realname and now it perform as expected.

In unit test added 2 more scenarios when there is customer ID and one more with recipient, Additionally included notification tag check, since it differs then auto response.

Please let me know what do you think about this proposal and if there are any questions regarding this PR, feel free to ask.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -1462,7 +1462,7 @@ sub _Replace {
                 }
 
                 # try to get the real name directly from the data
-                $From //= $Recipient{RealName};
+                $From //= $Recipient{Realname};
 
                 # get real name based on reply-to
                 if ( $Data{ReplyTo} ) {
--- a/Kernel/System/Ticket/Event/NotificationEvent.pm
+++ b/Kernel/System/Ticket/Event/NotificationEvent.pm
@@ -311,7 +311,9 @@ sub Run {
                 );
 
                 # remember to have sent
-                $AlreadySent{ $Bundle->{Recipient}->{UserID} } = 1;
+                if ( $Bundle->{Recipient}->{UserID} ) {
+                    $AlreadySent{ $Bundle->{Recipient}->{UserID} } = 1;
+                }
 
             }
 
--- a/scripts/test/TemplateGenerator/CustomerRealName.t
+++ b/scripts/test/TemplateGenerator/CustomerRealName.t
@@ -13,10 +13,14 @@ use utf8;
 use vars (qw($Self));
 
 # get needed objects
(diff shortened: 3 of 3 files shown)


mrcbnsls:
Well, great, I think that's what UT are for, right? It has already paid off that you've created it :-D

As some changes were made on the UT, could you please rebase it? Thanks!

s7design:
Hi @mrcbnsls 

I updated this PR, I believe it works now. You can see that we solved one issue that occurred in Kernel/System/Ticket/Event/NotificationEvent.pm. If Recipient is not stored in DB ( ie. there is not UserID which is used for $AlreadySent{ $Bundle->{Recipient}->{UserID} } = 1; ) there was message in log  'Use of uninitialized value in hash element at /opt/otrs//Kernel/System/Ticket/Event/NotificationEvent.pm line 303'

Please let me know if we can improve this test.

Regards
Zoran

s7design:
Hello @mrcbnsls ,

This PR is rebased and updated.

Regards,
Sanjin

mrcbnsls:
Merged, thanks, works good for me. See bugreport for merge URLs.

mrcbnsls:
Had to revert this. It seems that under certain circumstances the UT is failing. I could not observe this in my own testing environment though... could you please try to find out more about it?

s7design:
Hi @mrcbnsls 

Yes of course, we will check it

Regards
Zoran

mrcbnsls:
Any news on this, @s7design? :)

s7design:
Hi @mrcbnsls 

Sorry for late answer, we will have in that look.

Regards
Zoran

s7design:
Hi @mrcbnsls 

I updated this test, there was a problem with compering some strings. As you can see there is used index() function now. Compering with the operator ~= could make some problems. It looked like True() function doesn't have Name param

```
             $Self->True(
                $ArticleNotification{$NotificationKey} =~ /$Test->{ResultNotification}->{$NotificationKey}/,
                 "Notification: $Test->{Name}, $Test->{ResultNotification}->{$NotificationKey}, tag $NotificationKey"
             );
```
You will see I change $CustomerUser, because it was too long and makes a problem because in some notification such string went in new line so I adjusted this variable to pass the test. IMHO the behavior is the same and the test case is successfully tested.
I think it woks now, actually it works on my system.

Regards
Zoran

mrcbnsls:
Thanks, seems to work fine now. Let's see what the UT servers say. Merged in f6af7dd and d1d00e4.

PR #1119, opened 2016-04-09
Do not use hard coded action for cancel link

MariusTimmer:
Until now the action in the cancel link is hard coded. Creating it dynamicly allows a same behavior like the "Go to overview" link.


mgruner:
Hello @MariusTimmer, thanks a lot, I just applied your change to master and rel-5_0 for consistency reasons. But what is your real use case for it, why do you need it? Do you use the template with another frontend module?

mgruner:
BTW, for future contributions, please consider signing our CLA as described on https://otrs.github.io. Thanks!

MariusTimmer:
Hello @mgruner, thanks for your reaction. As final project of my apprenticeship I am writing a OTRS plugin. As you guessed I am using a template (AdminTemplate) with my own frontend module. Right now OTRS redirects me to the wrong frontend module if I click cancel. This would fix it.

Thank you for your hint. Next time I will respect your CLA. ;-)

PR #1121, opened 2016-04-12
Fixed bug #11430: (Ticket )Cache won't get updated on DynamicField Va…

thorsteneckel:
…lueDelete.

Please see: http://bugs.otrs.org/show_bug.cgi?id=11430

We ran in this issue again in one of our package UnitTests.

Basically this is what's wrong:

- Create Ticket with DF value
- Delete DF Value via DFBackendObject->ValueDelete
- Perform a TicketGet including DFs (in the same process as DF Delete)
- Notice that DF is still set in TicketGet result

A workaround for the TicketGet issue is:

```perl
# clear ticket cache
$TicketObject->_TicketCacheClear( TicketID => $TicketID );
```

Additionally there is no entry in the TicketHistory where the deletion of the value is logged mismatching the behavior of ValueSet which gets logged in TicketHistory.

If there are any questions please let me know.
  Thorsten

### DIFF
--- a/Kernel/System/DynamicField/Backend.pm
+++ b/Kernel/System/DynamicField/Backend.pm
@@ -621,6 +621,11 @@ sub ValueDelete {
         }
     }
 
+    my $OldValue = $Self->ValueGet(
+        DynamicFieldConfig => $Param{DynamicFieldConfig},
+        ObjectID           => $Param{ObjectID},
+    );
+
     # set the dynamic field specific backend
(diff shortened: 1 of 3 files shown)


mgruner:
Merged in 6dc7c336052d86c0c55fd1ea78f9e14c183c1b5f, thanks a lot Thorsten!

thorsteneckel:
Thank you!

PR #1123, opened 2016-04-14
Rel-5_0 - fixed visibility of navigation menu items

s7design:
Hi @mgruner and @mrcbnsls 

There is an issue with navigation manu items in case there are many menu items. You can see following image:

manu-items

It is my suggestion how we can solve that, but maybe it could be better if we refactoring this and add min-height and height:auto for NavigationContainer. I am not sure about that, but @mrcbnsls is definitely better with frontend and your opinion is welcomed about that.

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.js
+++ b/var/httpd/htdocs/js/Core.Agent.js
@@ -224,7 +224,7 @@ Core.Agent = (function (TargetNS) {
 
                     if ($('body').hasClass('Visible-ScreenXL')) {
 
-                        $('#NavigationContainer').css('height', '300px');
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Well, I think this would only be another workaround. Although its unlikely that you add as much items to a main menu item that the sub menu gets higher than 500px, it still COULD happen. So I decided to add a real solution here which resizes the container based on the currently expanded sub menu: 6c33eb7 and 4ae447c. Please let me know if this also works for you. Thanks anyway!

mrcbnsls:
@zottto, could you please also have a look at my solution? Thanks!

PR #1125, opened 2016-04-17
Second publish parameter should be an array.

thorsteneckel:


mrcbnsls:
Merged. There was another place in the same file where this event was published in the same way. I fixed this in a separate commit (e6d1fcd5473a852a75bcbeb4617fb178db77786c and 33ed3785ff5664fbb0551b84e4063d766459a95c).

PR #1130, opened 2016-04-20
Image pasting as data URI reenabled

pboguslawski:
This mod reenables pasting images as data URIs to signatures, templates
and AdminEmail messages. It also fixes problem with embedding inline
images using copy/paste on messages sent from AdminEmail (now such
images will be converted to attachments in e-mail which is more
compatible with poor mail readers).

This mod introduces also new SysConfig option
Frontend::RichText::PreventImagePaste (disabled by default) which
allowes one to disable pasting images as data URIs to rich text areas.

Fixes: 74fe26ecdb3be0fffe5551cce8349e0a40806193
Related: https://dev.ib.pl/ib/otrs/issues/41
Related: http://bugs.otrs.org/show_bug.cgi?id=11944
Author-Change-Id: IB#1051028

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-20 Image pasting as data URI reenabled.
  - 2016-04-15 Added possibility to use multiple named captures in Postmaster filters, thanks to Renée Bäcker.
  - 2016-04-08 Removed dummy 'Reply All' and 'Forward' options to align with 'Reply' select, thanks to Nils Leideck.
  - 2016-04-01 Fixed dropdown for CustomerTicketOverview.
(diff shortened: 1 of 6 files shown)


zottto:
Hi @pboguslawski,

thanks for your PR to re-enable pasting of base-64 encoded images for signatures, templates etc.
These modules never had image functionality, pasting of base64 encoded images was a workaround that worked by accident.
Although we agree that it should be possible to add images to content in these modules, this is not possible at the moment. But re-enabling the base64 workaround for these modules is also not an option for us at the moment.

We will work on a solution that supports real image upload in these modules, which will also allow pasting base64 encoded images again.

I'll close this PR for the moment. Please feel free to re-open, if you have more information or adjustments to your code that lead us in the correct direction.

Thanks,
Marc

PR #1131, opened 2016-04-22
Add the possibility to use variables in URL config

thorsteneckel:
Currently it's not possible to build OTRS internal links via a 'Frontend::CustomerUser::Item' SysConfig using the  Kernel/Output/HTML/CustomerUser/Generic.pm backend. It's only possible to configure absolute URLs.

This change makes it possible to make configurations like:

```xml
...
                <Item Key="Required">UserID</Item>
                <Item Key="Attributes">UserID</Item>
                <Item Key="URL">[% Env("CGIHandle") %]?Action=AgentTicketCustomAction;CustomerUserID=[% Data.UserID | uri %]
...
</Item>
```

possible.


mgruner:
Hello @thorsteneckel. Did you test this? I currently don't see that this kind of data is acutally passed to the block at all, so this would not be working. Besides, Generic.pm uses another internal mechanism to construct the URL which does support generating dynamic URLs. Please feel free to provide more information!

PR #1133, opened 2016-04-24
Added table filters to some more admin modules were easy applicable.

nilei:


nilei:
Hi @mrcbnsls, as requested I Added 'No matchs found' row to AdminACL and a match count check to AdminSelectBox to only display the filter if more than one result item is returned.

Additionally I added a 'W10pc' to AdminPostmasterFilter 'Delete' table column and a match count to the result table header in AdminSelectBox.

Cheers, Nils

mrcbnsls:
Thanks a lot, merged in 8df59db.

PR #1134, opened 2016-04-25
Fixed parsing CSV data with quoted values containing newlines

pboguslawski:
OTRS CSV parser does not handle quoted values with newlines. Parsing file like

    c1;c2;c3
    "c1 value line1
    c1 value line2";c2 value;c3 value

produces errors.

According to

* https://metacpan.org/pod/Text::CSV_XS#Embedded-newlines
* http://perlmaven.com/how-to-read-a-csv-file-using-perl

for parsing such data one should use getline().

Related: https://dev.ib.pl/ib/otrs/issues/44
Author-Change-Id: IB#1013124

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-25 Fixed parsing CSV data with quoted values containing newlines.
  - 2016-04-22 Added possibility to set the ticket title in Postmaster filters, thanks to Renée Bäcker.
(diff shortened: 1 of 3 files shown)


reneeb:
+1

mgruner:
Merged in 463d6b92fed77e92e612d6b20e274f44e8e790f4 with several changes. Thanks a lot! Please always use OTRSCodePolicy in future before sending PRs, there were several issues in this one that also Travis had reported about.

PR #1135, opened 2016-04-26
Master - Added new responsible mandatory fields

s7design:
Hi @mgruner 

As you suggested in the bug http://bugs.otrs.org/show_bug.cgi?id=12011 we implemented Responsible mandatory fields in the screen there are Responsible field as well.
There is updated AgentTicketResponsible.t, if you would like we will be able to update the other Selenium test where we can check this field ( AgentTicketFreeText, AgentTicketClose, AgentTicketNote AgentTicketOwner ...). 

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -2819,6 +2819,17 @@
             </Option>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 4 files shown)


mgruner:
@mrcbnsls can you QA and merge to master (only) please?

mrcbnsls:
Thanks, looks good. Merged in b254895.

PR #1136, opened 2016-04-26
Master - Translate strings directly in JavaScript files - part 1

s7design:
Hi @zottto 

There is first PR with refactoring JS where is used translate directly in JS. 
I started removing string from CustomerFooterJS.tt and FooterJS.tt, I suppose it is OK. 
I saw that these strings are also used in some else packages.  For example Duplicated.TitleText and Duplicated.RemoveText are used in OTRSBussines, but now they are removed in the framework. I suppose we should update these packages as well, when we finish refactoring here in the framework.

I update in AgentCustomerSearch.tt as a example that you follow-up if I have done this well. 
 
You also mentioned in the mail:

> in all other tt files there is even more translated text in JavaScript blocks
> if the translation is used directly, please ignore it for the moment 

I think you meant on something like  that:

https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/Templates/Standard/AdminACLEdit.tt#L263

Instead of DialogCloseMsg I used only Close, as I have see you used in 
https://github.com/OTRS/otrs/commit/344fdd109526357fe00a9827ca6eb0235ebb65dd#diff-b9564e1cf513f62502507520701c60b5R197

Please check this, and I will be able to continue to work on this task.

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
+++ b/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
@@ -638,7 +638,7 @@ Core.Agent.CustomerSearch = (function (TargetNS) {
     TargetNS.ShowDuplicatedDialog = function(Field){
         Core.UI.Dialog.ShowAlert(
             Core.Language.Translate('Duplicated entry'),
-            Core.Language.Translate('This address already exists on the address list. It is going to be deleted from the field, please try again.'),
+            Core.Language.Translate('This address already exists on the address list.') + ' ' + Core.Language.Translate('It is going to be deleted from the field, please try again.'),
             function () {
                 Core.UI.Dialog.CloseDialog($('.Alert'));
                 $('#' + Field).val('');
--- a/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
+++ b/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
@@ -637,8 +637,8 @@ Core.Agent.CustomerSearch = (function (TargetNS) {
      */
     TargetNS.ShowDuplicatedDialog = function(Field){
         Core.UI.Dialog.ShowAlert(
-            Core.Config.Get('Duplicated.TitleText'),
-            Core.Config.Get('Duplicated.ContentText') + ' ' + Core.Config.Get('Duplicated.RemoveText'),
+            Core.Language.Translate('Duplicated entry'),
+            Core.Language.Translate('Duplicated entry') + ' ' + Core.Config.Get('It is going to be deleted from the field, please try again.'),
             function () {
                 Core.UI.Dialog.CloseDialog($('.Alert'));
                 $('#' + Field).val('');
--- a/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
+++ b/var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
@@ -638,7 +638,7 @@ Core.Agent.CustomerSearch = (function (TargetNS) {
     TargetNS.ShowDuplicatedDialog = function(Field){
         Core.UI.Dialog.ShowAlert(
             Core.Language.Translate('Duplicated entry'),
-            Core.Language.Translate('Duplicated entry') + ' ' + Core.Config.Get('It is going to be deleted from the field, please try again.'),
+            Core.Language.Translate('This address already exists on the address list. It is going to be deleted from the field, please try again.'),
             function () {
                 Core.UI.Dialog.CloseDialog($('.Alert'));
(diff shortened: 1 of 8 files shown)


var/httpd/htdocs/js/Core.Agent.CustomerSearch.js
```diff
@@ -637,8 +637,8 @@ Core.Agent.CustomerSearch = (function (TargetNS) {
      */
     TargetNS.ShowDuplicatedDialog = function(Field){
         Core.UI.Dialog.ShowAlert(
-            Core.Config.Get('Duplicated.TitleText'),
-            Core.Config.Get('Duplicated.ContentText') + ' ' + Core.Config.Get('Duplicated.RemoveText'),
+            Core.Language.Translate('Duplicated entry'),
+            Core.Language.Translate('Duplicated entry') + ' ' + Core.Config.Get('It is going to be deleted from the field, please try again.'),
```
urbalazs: Hi @s7design 
You need to change this line to:
<code>Core.Language.Translate('This address already exists on the address list. It is going to be deleted from the field, please try again.'),</code>

var/httpd/htdocs/js/Core.Agent.TicketAction.js
```diff
@@ -209,7 +209,7 @@ Core.Agent.TicketAction = (function (TargetNS) {
                     Core.App.Publish('Event.Agent.TicketAction.NeedSpellCheck', [$('#RichText')]);
                     Core.UI.Dialog.ShowContentDialog('<p>' + Core.Config.Get('SpellCheckNeededMsg') + '</p>', '', '150px', 'Center', true, [
                         {
-                            Label: '<span>' + Core.Config.Get('DialogCloseMsg') + '</span>',
+                            Label: '<span>' + Core.Language.Translate('Close') + '</span>',
```
urbalazs: Please use the original message. See my https://github.com/OTRS/otrs/pull/1100 for the explanation.

s7design:
Hi @urbalazs 

Thanks a lot for your suggestions. I made mistake in Core.Agent.CustomerSearch.js .

Regards
Zoran

zottto:
Hi Zoran!

This looks quite good, just some comments:

I would suggest to not change the translated strings:
https://github.com/OTRS/otrs/pull/1136/commits/1734d191a6b4ccd827a81bd33ab412d5da6d15ef#diff-e77dec3e30c8143b954fc36f8ebe11d4L641
you should not merge both strings

As for the dialog close message, please keep the original "Close this dialog". I accidently had the old "Close" message in my branch when I merged this to master.


Regarding the strings from FooterJS and CustomerFooterJS: if these strings are also used in other modules (which is definetly the case as you mentioned) we should also update these strings there later.

Things like https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/Templates/Standard/AdminACLEdit.tt#L263 can be converted, too.
I thought of translation usage like https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt#L13 (where "Article filter" is translated directly). These kind of translations can be ignored in the first step.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for follow-up and useful suggestions. I will update this PR and then continue with the updating other files. 

Regards
Zoran

s7design:
Hi @zottto 

I updated PR, I hope it is ready for merging now. 
Please let me know if I can improve something else.

I will made more PRs for the rest of the framework.

Regards
Zoran

zottto:
Hi Zoran,

there is still "Close this widget" instead of "Close this dialog" in https://github.com/s7design/otrs/blob/master_translate_JS_part1/var/httpd/htdocs/js/Core.Agent.TicketAction.js#L212

Everything else looks good.

Marc

s7design:
Hi @zottto 

I forgot to do that, sorry. I hope it is good now.

Regards
Zoran

zottto:
Hi Zoran,

"close this dialog" is now also in https://github.com/s7design/otrs/blob/master_translate_JS_part1/Kernel/Output/HTML/Templates/Standard/AgentDashboardCommon.tt which is wrong, because these elements are real widgets and should keep there "close this widget" text. Besides that, the changes were made in HTML code and not in JS code.

Please correct that.

Thanks,
Marc

s7design:
Hi Marc, 

Thanks for follow-up it is my mistake. It will be done.

Regards
Zoran

s7design:
Hi @zottto,

I updated this PR, I hope it is good now.

Regards
Zroan

zottto:
Merged with d0249b4.

PR #1137, opened 2016-04-27
Fixed bug: Missing EscapeSelector.

thorsteneckel:
Hi there,

we stumbled over an issue while using the search for ITSM ConfigItems where the attribute is a subelement like: "Building::Room". The '::' are causing a jQuery selector exception. Using Core.App.EscapeSelector fixes this.


mrcbnsls:
Thanks, merged in 0da3a41 and ac9b76a.

PR #1138, opened 2016-04-27
Fixed wrong proposed default article type when replying/forwarding

pboguslawski:
When you reply to phone article (incoming or outgoing) AgentTicketCompose
form proposes email-internal as default article type if agent uses Polish
language in web frontend and email-external if agent uses English
language in web frontend.

When forwarding from article with internal type - email-external is
proposed as default in AgentTicketForward form.

With this fix OTRS proposes email-internal in AgentTicketForward and
AgentTicketCompose if sending from article with "internal"
type and email-external otherwise.

Related: https://dev.ib.pl/ib/otrs/issues/46
Author-Change-Id: IB#1014394

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-27 Fixed wrong proposed default article type when replying/forwarding.
  - 2016-04-22 Added possibility to set the ticket title in Postmaster filters, thanks to Renée Bäcker.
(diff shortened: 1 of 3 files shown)


mgruner:
The real cause was that the configured default article type was not honored in AgentTicketCompose, please see c744fb80b1190558cecbb0713f4e6f9359345228. Thanks for your proposal!

pboguslawski:
Please see https://github.com/OTRS/otrs/pull/1146 for new options that allow to preserve internal article type in AgentTicketCompose and AgentTicketForward.

PR #1139, opened 2016-04-27
Remove warning message about invalid encoding

moritz:
I don't see the value in printing it, and it potentially prints the whole
body of an email message (when called from within the postmaster), which can
be several megabytes, and completely blow up log files.

If removing this warning is not acceptable, there should at least be a length limit to the printed string.


mgruner:
Merged, thanks @moritz!

PR #1141, opened 2016-04-28
Fixed error log flooding with less important messages

pboguslawski:
In case of

* unsuccessful MX validation of e-mail address entered by agent,
* more matching LDAP entries than configured limit,

OTRS generates error log message.

Such events are not real errors and can flood error logs in bigger
systems which makes it difficult to search logs for real errors.

This mod changes such log messages priority from error to debug;
this stops error log flooding to apache error log and allow one
to disable logging it to syslog also by setting MinimumLogLevel
higher than debug.

This mod also changes order of MinimumLevelNum initialization
which may be skipped if IPC::SysV is not available.

Related: https://dev.ib.pl/ib/otrs/issues/48
Author-Change-Id: IB#1050078

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-04-28 Fixed error log flooding with less important messages.
  - 2016-04-26 Added possibility to push custom data directly from Perl code to JavaScript without having to embed it into templates.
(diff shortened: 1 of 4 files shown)


mgruner:
Already merged a while ago in 6b50b1f. Thanks a lot @pboguslawski.

PR #1142, opened 2016-04-28
Fixed capitalizations

urbalazs:
Hi @mgruner 
This PR fixes the capitalizations according to the [User Interface Design](http://otrs.github.io/doc/manual/developer/stable/en/html/ui-design.html#id-1.6.7.2) chapter of the developer manual.
Can you backported into rel-5_0, if it is possible? And after that can you sync translations with Transifex? Many strings has been changed since last sync.

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminACL.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminACL.tt
@@ -39,7 +39,7 @@
 
         <div class="WidgetSimple">
             <div class="Header">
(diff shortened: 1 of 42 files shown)


mgruner:
Impressive! Merged in b246d772be9dfd9421426d9a362931456ab152d7. I will not backport this to rel-5 to avoid unneeded retranslations which will not improve the quality of translated UIs. I will start a sync to transifex for rel-5 however. Thanks a lot @urbalazs!

PR #1144, opened 2016-04-29
Do not cache ticket data when no ticket was found

reneeb:
Use cache only when a ticket number is found otherwise a non-existent ticket is cached. That can cause errors when the cache isn't expired and e.g. postmaster creates that ticket.


mgruner:
Excellent, merged, thanks @reneeb!

PR #1146, opened 2016-05-02
Internal article type preseved in AgentTicketCompose and AgentTicketF…

pboguslawski:
…orward

When composing messages in AgentTicketCompose and AgentTicketForward OTRS
uses parameters

    Ticket::Frontend::AgentTicketForward###ArticleTypeDefault
    Ticket::Frontend::AgentTicketCompose###DefaultArticleType

for proposing default new article type. This may lead to problems
with internal information disclosure in customer web panel (i.e. when
agent replies to or forwards internal message without changing default
"email-external" article type proposed by OTRS).

This mod introduces two SysConfig options

    Ticket::Frontend::AgentTicketCompose###DefaultArticleTypeInternal
    Ticket::Frontend::AgentTicketForward###ArticleTypeDefaultInternal

which allow to define default article type when composing/forwarding
e-mails from source article with type that contains "internal"
in its name. By default both new options are set to "email-internal"
which keeps default "internal" status of messages composed/forwarded
from other "internal" articles.

Fixes: a78dfffa46edf71f4e99a35c0c6121bd4e5f0b8d
Related: https://dev.ib.pl/ib/otrs/issues/49
Author-Change-Id: IB#1051354

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-20 Internal article type preseved in AgentTicketCompose and AgentTicketForward.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
  - 2016-04-29 Fixed parsing CSV data with quoted values containing newlines, thanks to Pawel Boguslawski.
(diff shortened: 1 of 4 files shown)


mgruner:
Hello Pawel,
thanks for your suggestion. I believe that OTRS is already over-configurable and that the existing configuration settings are enough. You can set them to internal article types by default if you are concerned.
Best regards, Martin

PR #1147, opened 2016-05-02
Master - Translate strings directly in JavaScript files part 2

s7design:
Hi @zottto 

Hereby are updated some JS scripts and translated strings directly there.

Regards
Zoran


s7design:
Hi @zottto 

I updated this PR because I saw that there was fixed only DeleteMsg in the updated tt modules in Localization strings. 

I think it will be better if I complete all migration to Core.Language.Translate in these modules at once. It is easier to follow which modules are done.

Regards
Zoran

zottto:
Hi Zoran,

please keep in mind, that the translations were only put in the template files, because we did not have another possibility for translations.
This means, that code like
    Core.Agent.Admin.ACL.Localization = { ... }
in the template was only there to get the translations somehow into JS.
You now changed this code to use Core.Language.Translate(), but you kept the code block in the template file (like https://github.com/s7design/otrs/blob/master_translate_JS_part2/Kernel/Output/HTML/Templates/Standard/AdminACLEdit.tt#L262). In these cases you should check for the usage of these translations in the corresponding JavaScript files and replace the usage of
    Core.Agent.Admin.ACL.Localization.Something
directly there.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for your suggestion, it helps me a lot to understand what I need to do with Localization array which was defined in TT modules. I hope it is good now.
I removed Localization from TT but also from JS, I think it is not necessary any more.

Please let me know if it is good. I will fix other PRs tomorrow, where I did not removed Localization form the TT.

Regards
Zoran

zottto:
Hi Zoran,

looks really good, only one small issue left:
https://github.com/s7design/otrs/blob/5afae3838f43c6fdbdad82e7b29be506dc89d05d/Kernel/Output/HTML/Templates/Standard/AgentLinkObject.tt#L90

You replaces the Core.Config.get() call with a Core.Language.Translate() within a tt file. The Translate() will not work in a tt file (although it is within a JS part). Please replace this with a [% Translate %] block.

Thanks,
Marc

zottto:
Merged with 08a35fb.

PR #1148, opened 2016-05-03
Master - Translate strings directly in JavaScript files part 3

s7design:
Hi @zottto

There are updated some JS scripts and translated strings directly there.
Hereby  are updated CustomerFooterJS.tt and FooterJS.tt. In the next PR I will cover *.tt modules where script tag are used.

Regards
Zoran


zottto:
If you find translation strings that are used as a variable value (like in https://github.com/s7design/otrs/blob/master_translate_JS_part3/var/httpd/htdocs/js/Core.UI.InputFields.js#L372), please check, if you can remove the variable and use the string directly.

Marc

s7design:
Hi @zottto 

I updated this PR as you suggested, I hope it is better now.

Regards
Zoran

zottto:
Merged with b1e1c13.

PR #1149, opened 2016-05-04
[Rel-5_0 Bug 12020] - Option CaseSensitive has no impact on external customer database

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12020

Hello @mgruner ,

Hereby is our solution for fixing reporter issue. There was a small mistake with IF statement in QuerryCondition function. "$Self->GetDatabaseFunction('CaseSensitive')" and "$CaseSensitive" are carrying same value, so function never executed 'else' statement in this part.

Please let me know if you have any questions or issues regarding this fix.

Kind Regards,
Sanjin


mgruner:
Hello S7, the problem is a bit more complicated than that. Currently a colleague is working on this issue.

PR #1150, opened 2016-05-05
Master - Translate strings directly in JavaScript part 4

s7design:
Hi @zottto

Hereby are updated some tt modules where strings in JS are translated  directly.
I am not sure should I use Core.Language.Translate in confirm in AdminNotificationEvent.tt 
```
    if (window.confirm(Core.Language.Translate('Do you really want to delete this notification?'))) {
         window.location = $(this).attr('href');
         window.location = $(this).attr('href');
     }
     }

```

Also I have dilemma about EditConfirm strings n some AdminProcessManagement modules. IF there is necessary quotation marks ( EditConfirm: "[%  ...  | html %]" )

```
-    EditConfirm: "[% Translate("As soon as you use this button or link, you will leave this screen and its current state will be saved automatically. Do you want to continue?") | html %]"

+    EditConfirm: Core.Language.Translate('As soon as you use this button or link, you will leave this screen and its current state will be saved automatically. Do you want to continue?')
```

You can see how I have done, but I will be able to update if it is not good. Please let me know what do you think about that.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminDynamicField.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminDynamicField.tt
@@ -113,7 +113,7 @@ $('#ShowContextSettingsDialog').bind('click', function (Event) {
     Core.UI.Dialog.ShowContentDialog($('#ContextSettingsDialogContainer'), [% Translate("Settings") | JSON %], '20%', 'Center', true,
         [
             {
-                Label: [% Translate("Save") | JSON %],
+                Label: Core.Language.Translate('Save'),
                 Type: 'Submit',
                 Class: 'Primary'}
         ]);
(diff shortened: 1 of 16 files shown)


zottto:
As mentioned already in other PRs, please use the translation directly in the corresponding JS file instead of using it in the template in a JS hash (like in https://github.com/s7design/otrs/blob/master_translate_JS_part4/Kernel/Output/HTML/Templates/Standard/AdminGenericInterfaceDebugger.tt).

This also solves your problem in AdminProcessManagement.

The usage of Core-Language.Translate() within window.confirm() is correct.

Marc

s7design:
Hi @zottto 

I updated this PR, I believe it is much better now. Thanks for your suggestions.

Regards
Zoran

zottto:
Hi Zoran,

I also saw in this PR that you used Core.Language.Translate() within a tt file (e.g. here: https://github.com/s7design/otrs/blob/1fbe7ba6e658f6474c00f268a60c7c153c44f52e/Kernel/Output/HTML/Templates/Standard/AdminDynamicField.tt#L116). As mentioned already in the other PR, this will not work. Please use the template block instead.

Thanks,
Marc

s7design:
HI @zottto

Thanks for suggestions, I updated this PR.

Regards
Zoran

zottto:
Merged with 4ca6ebd.

PR #1153, opened 2016-05-06
Master - Translate strings directly in JavaScript part 5

s7design:
Hi @zottto ,

Hereby is next part of updated direct translation in JS.
Please if you could confirm just one thing. Using Translate() when it's already in a function.
`Core.Config.Set('Localization.Delete', Core.Language.Translate('Delete'));`

If this is not the right way to use this method, i will fix it and update PR. Please let me know what do you think about this.

Regards,
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminSysConfigEdit.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminSysConfigEdit.tt
@@ -502,7 +502,7 @@
             Length = $(this).text().length,
             ShortText,
             FullText,
-            LinkText = [% Translate("Show more") | JSON %],
+            LinkText = Core.Language.Translate('Show more'),
             OverlayTitle = $(this)
                 .closest('fieldset')
                 .find('h3')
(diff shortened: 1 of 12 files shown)


zottto:
In AdminSysConfigEdit.tt, please don't use the variable LinkText, but replace the

s7design:
Hi @zottto,

I saw all your suggestions and thank you for follow-up. It helps me a lot.  I will fix all of that as soon as possible. 

Regards
Zoran

s7design:
Hi @zottto 

I updated this PR.

Regards
Zoran

zottto:
Hi Zoran,

sorry that I perhaps was not clear enough on this, but also in this PR you used Core.Language.Translate() directly in the tt file. As long as the JS code is in the tt file, please still use the [%Translate %] element.

Thanks,
Marc

s7design:
Thanks @zottto  for your suggestion, I will fix it tomorrow. I believe all is finally clear about this task :)

regards
Zoran

s7design:
Hi @zottto 

This PR is updated as well, I hope it is good now.

Regards
Zoran

zottto:
Merged with 9f73690.

PR #1158, opened 2016-05-10
Allow X-OTRS-Loop to reenable autoresponses

pboguslawski:
Any non-empty value passed to X-OTRS-Loop header in e-mail message
or any postmaster filter disables autoresponse.

This mod forces OTRS to disable autoresponding only if X-OTRS-Loop
is not empty and not false|no. This allowes one to disable
autoresponding in one filter with X-OTRS-Loop=yes and reenable
autoresponding in other filter with X-OTRS-Loop=no
(or X-OTRS-Loop=false).

Related: https://dev.ib.pl/ib/otrs/issues/52
Author-Change-Id: IB#1017277


mgruner:
Added tests and merged in 4c1857919ba8706b0a3dc5a73d010fbe0c8b9311. Thanks a lot @pboguslawski!

PR #1161, opened 2016-05-11
Missing space from error message

urbalazs:
Hi @mgruner 
There was no space character between the two strings. The concatenation is unnecessary. Rel-5_0 is also affected.


mgruner:
Merged in ae6098e47eb1583cda83ef8c4f1513626f1105db and backported, thanks a lot @urbalazs! 👍

PR #1163, opened 2016-05-12
Added MailReprocessSpoolDir option

pboguslawski:
This mod introduces new SysConfig option MailReprocessSpoolDir
which allows one to change location of e-mail messages stored
for reprocessing.

Related: https://dev.ib.pl/ib/otrs/issues/54
Author-Change-Id: IB#1017361

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-12 Added MailReprocessSpoolDir option.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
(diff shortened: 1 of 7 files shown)


mgruner:
Hello @pboguslawski. Why would this be a requirement, what's a use case for this?

pboguslawski:
Hi,

We use it to specify custom dir for spool. If Article & tmp dirs are configurable, why not spool?

Regards,
Pawel

mgruner:
Ok, but what do you achieve with that? Article and tmp dirs can be quite big, but this one?

pboguslawski:
Hi,

We don't like to keep dirs with changing data inside /opt/otrs directory and move stuff like articles, tmp, spool away from /opt/otrs. Using simlinks is less elegant than SysConfig parameter in our opinion that why this mod.

Regards,
Pawel

carlosfrodriguez:
Hi @pboguslawski,

Thank you for your pull request, currently the backup.pl script does the backup for all var directory, with this feature this is not covered, can you add the manage of the spool directory into the backup.pl script? 

Please also rebase the this pull request
((enjoy))

mgruner:
I'll close this pull request for now.

PR #1164, opened 2016-05-13
Master - Translate strings directly in JavaScript part 6

s7design:
Hi @zottto ,

There is next part of updated direct translation in JS. Hereby all TT modules is checked, I hope I updated all strings in script tags. 

I was not sure about abort in: PreferencesNotificationEvent.tt. I changed it to `Core.Language.Translate("Sorry, but you can't disable all methods for this notification.")` but after I called Dev::Tools::TranslationsUpdate, this string is not added in JavaScriptStrings. Is it case that you mentioned that we ignored it for the moment?

The same with [% Translate("You have unanswered chat requests") | JSON %]  in CustomerNavigationBar.tt. Should I fix it and use Core.Language.Translate. I think no need.

I didn't translate directly 

```
    Core.Config.Set('RichText.Lang.SplitQuote', '[% Translate('Split Quote') | html %]');
    Core.Config.Set('RichText.Lang.RemoveQuote', '[% Translate('Remove Quote') | html %]');
```
from the following modules:
RichTextEditor.tt
CustomerRichTextEditor.tt

This strings are used in the thirdparty script:
/opt/otrs/var/httpd/htdocs/js/thirdparty/ckeditor-4.5.6/plugins/splitquote/plugin.js

This also cannot be added to JavaScriptStrings after running Dev::Tools::TranslationsUpdate.

Please let me know what do you think about this. Your opinion is welcome now. If all PRs that I made is good, this task is done. I can go on with the same work on the packages.

Regards,
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/DashboardEventsTicketCalendar.tt
+++ b/Kernel/Output/HTML/Templates/Standard/DashboardEventsTicketCalendar.tt
@@ -11,59 +11,6 @@
 
     $(document).ready(function() {
         Core.Agent.Dashboard.EventsTicketCalendarInit({
-            AllDayText: [% Translate("All-day") | JSON %],
-            MonthNames: [
-                [% Translate("January") | JSON %],
-                [% Translate("February") | JSON %],
-                [% Translate("March") | JSON %],
-                [% Translate("April") | JSON %],
-                [% Translate("May_long") | JSON %],
-                [% Translate("June") | JSON %],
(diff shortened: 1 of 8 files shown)


zottto:
Merged with small changes in a8b8367.

For PreferencesNotificationEvent.tt and CustomerNavigationBar.tt, these are strings that we keep them as they are at the moment (with [% Translate %]).

Also, it is correct to not change the settings of the rich text editor.

Thanks,
Marc

PR #1165, opened 2016-05-14
Added Hungarian stats

urbalazs:
Hi @mgruner 
Here is the localized (Hungarian) stats for OTRS. As this XML files contains only the localization, it would be good to also add to rel-5_0, because there are the same files both in rel-5_0 and master.

Suggestion:
These XML files are redundant. All of them contain the same lines. The only difference between *_en.xml, *_de.xml and now *_hu.xml is the &lt;Description&gt; and the &lt;Title&gt; element.
These XML files can be merged into one with a lang="" attribute (like in OTRS packages). For example:

ListOfOpenTicketsSortedByTimeLeftUntilEscalationDeadlineExpires.xml
&lt;Title lang="de"&gt;Ticketliste aller offenen Tickets geordnet nach der verbleibenden Zeit bis um Ablauf der Eskalationsfrist&lt;/Title&gt;
&lt;Title lang="en"&gt;List of open tickets, sorted by time left until escalation deadline expires&lt;/Title&gt;
&lt;Title lang="hu"&gt;Nyitott jegyek listája az eszkalációs határidő lejáratáig hátralévő idő szerint rendezve&lt;/Title&gt;

And the same with the &lt;Description&gt; tag.

### DIFF
--- a/var/stats/ListOfOpenTicketsSortedByTimeLeftUntilEscalationDeadlineExpires.hu.xml
+++ b/var/stats/ListOfOpenTicketsSortedByTimeLeftUntilEscalationDeadlineExpires.hu.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<otrs_stats>
+<Cache>0</Cache>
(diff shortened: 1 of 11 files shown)


mgruner:
Merged and backported without further changes at this point, thanks a lot!

PR #1166, opened 2016-05-17
Master - fixed, updated test AgentTicketZoom

s7design:
Hi @mgruner 

I saw that the test AgentTicketZoom was failed here http://opms.otrs.com/otrs/public.pl?Action=PublicUnitTest;Subaction=ViewUnitTest;XMLKey=OTRS%205%20git_ut-ubuntu-1604.otrs.com_5.22.1_mysql%205.7.12;ShowSuccess=0

To ne honest I couldn't reproduce this, and I am not sure why it failed. Hereby I updated test and maybe solved this problem, if you like this solution you can mage it.

Regards
Zoran


mgruner:
Merged, thanks.

PR #1167, opened 2016-05-17
Master - Bug 12065 - queue and state not mandatory 

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12065

Hi @mgruner 

Hereby is solved the bug and also we extended the test AgentTicketFreeText.t
We are planing to continue the same with other screens where there are not mandatory fields (Queue, Owner,Responsible,State)
I am not sure if it is necessary for other screens. 

Please let me what do you think about that.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentTicketActionCommon.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentTicketActionCommon.tt
@@ -227,14 +227,6 @@
                             [% END %]
                         </div>
                         <div class="Clear"></div>
-
-                        <label class="Mandatory" for="NewStateID"><span class="Marker">*</span>[% Translate("Next state") | html %]:</label>
-                        <div class="Field">
-                            [% Data.StateStrg %]
-                            <div id="NewStateIDError" class="TooltipErrorMessage"><p>[% Translate("This field is required.") | html %]</p></div>
-                            <div id="NewStateIDServerError" class="TooltipErrorMessage"><p>[% Translate("State invalid.") | html %]</p></div>
-                        </div>
-                        <div class="Clear"></div>
 [% RenderBlockEnd("State") %]
 
 [% RenderBlockStart("StatePending") %]
--- a/Kernel/Output/HTML/Templates/Standard/AgentTicketActionCommon.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentTicketActionCommon.tt
@@ -227,16 +227,7 @@
                             [% END %]
                         </div>
                         <div class="Clear"></div>
-[% WRAPPER JSOnDocumentComplete %]
-<script type="text/javascript">//<![CDATA[
-    $('#NewStateID').bind('change', function (Event) {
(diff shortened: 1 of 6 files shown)


s7design:
Hi @mrcbnsls 

You can see we changed in AgentTicketActionCommon.tt ServerError messages. All these messages were 'This field is required.'. Now they are changes similar to ServerError which there is for Service input field. I know that it is not maybe the best solution but it out proposal in this moment. 

There are added mandatory sysconfig items for all screens from AgentTicketActionCommon.pm and AgentTicketMove.pm as well. In my opinion there is no sense for other screens. If you would like we will be able to add for other once.

Selenium test is extended only for AgentTicketFreeText. There is checked mandatory fields. I spoke with Marin by Skype, and we agree that there is not needed extended test with such test cases for other changed screens.

Regards
Zoran

Kernel/Output/HTML/Templates/Standard/AgentTicketActionCommon.tt
```diff
@@ -115,6 +114,23 @@ $('#TypeID').bind('change', function (Event) {
...
+[% WRAPPER JSOnDocumentComplete %]
+<script type="text/javascript">//<![CDATA[
+    $('#NewQueueID').bind('change', function (Event) {
+        Core.AJAX.FormUpdate($('#Compose'), 'AJAXUpdate', 'NewQueueID', [ 'TypeID', 'ServiceID', 'NewOwnerID', 'NewResponsibleID', 'NewStateID', 'NewPriorityID' [% Data.DynamicFieldNamesStrg %] ]);
+    });
+//]]></script>
+[% END %]
+[% RenderBlockEnd("QueueMandatory") %]
+
 [% RenderBlockStart("Service") %]
```
mrcbnsls: Hi @s7design, I don't think its necessary to have two different blocks which basically do the same, except for the label marker in case the field is mandatory and some error message. Why not have only one block and decide if it should be mandatory using a parameter which is called on calling Block() in AgentTicketActionCommon.pm?

s7design: Thanks we will change that. You are definitely right.

Zoran

s7design:
Hi @mrcbnsls 

I updated modules and removed *Mandatory blocks. I provided solution with TemplateToolKit IF statements. I suppose you like it.
Maybe we should check other screens where there are such mandatory block and I will be able to update it as well in this PR or another one.

Regards
Zoran

s7design:
Hi @mrcbnsls 

It is rebased, I hope it works now

Regards
Zoran

mrcbnsls:
For me the state fields now shows up twice in AgentTicketFreeText (and probably in other screens, too).

bildschirmfoto 2016-08-05 um 14 52 57

s7design:
Hi @mrcbnsls 

Thanks, sorry it is my mistake, I've not seen that. It is fixed now, and I hope it is finally right.

Regards
Zoran

mrcbnsls:
Thanks, looks good for me. Merged in e93a39c7f38612ba008a4888f5c61449ae1121e7.

PR #1168, opened 2016-05-18
Master - Changed sysconfig name for AgenTicketPhoneCommon state default

s7design:
Hi @mgruber

Hereby is updated name for  AgentTicketPhoneOutbound###State and AgentTicketPhoneInbound###State. I saw that such sysconfig items for other screen is named StateDefault. I thought it would be better if there is consistently name for this item here as well. It is not so important but if you agree with me you can merge this.
There is updated Selenium test where this item is checked.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -3041,7 +3041,7 @@
             <TextArea></TextArea>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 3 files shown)


mgruner:
Hi @s7design, thanks for the proposal. I know we have some inconsistently named settings, but changing them would usually require a migration script, and I'm not sure that it's worth it in this case.

PR #1169, opened 2016-05-18
Fixed missing OTRS tags in AdminTemplate after Frontend::RichText change

pboguslawski:
If you have stored any plaintext (Frontend::RichText = No) templates
in AdminTemplate with `<OTRS*>` tags inside
(i.e. `<OTRS_OWNER_UserFirstname>`) and then switch to
Frontend::RichText = Yes, then all `<OTRS*>` tags will be lost
in rich text editor in AdminTemplate.

This patch solves this problem analogous to AdminAutoResponse.

Related: https://dev.ib.pl/ib/otrs/issues/55
Author-Change-Id: IB#1019597


mgruner:
Bug http://bugs.otrs.org/show_bug.cgi?id=12087 created and fix applied, thanks a lot @pboguslawski!

PR #1170, opened 2016-05-18
Follow up detection based on Message-ID e-mail header

pboguslawski:
OTRS creates separate tickets if one sends same e-mail message to
a few OTRS system addresses (i.e. using multiple OTRS system
addresses in To/Cc/Bcc in the same e-mail message) . It may be
difficult to notice such duplicates in bigger OTRS systems
and same request be processed by different agents.

This mod introduces SysConfig parameter
`PostMaster::CheckFollowUpModule###0600-MessageID`
that allowes OTRS to merge messages with the same Message-ID
into one ticket. To avoid merging too many messages into one
ticket (in case of Message-ID generation problem at sender side)
or merging new messages to very old tickets in case of
Message-ID duplicates in wider time, two module parameters
are provided:

* `MaxAge` - time (in seconds) from the first article with
   given Message-ID creation, after which new messages with
   the same Message-ID won't be merged into the same ticket
   any more,

* `MaxArticles` - number of messages with the same
   Message-ID in OTRS database after which next e-mail messages
   with the same Message-ID won't be merged into the same ticket
   any more.

For example: with MaxAge=604800 and MaxArticles=10 OTRS will merge
into one ticket up to 10 messages with the same Message-ID if all
will arrive in the same week.

Related: https://dev.ib.pl/ib/otrs/issues/56
Author-Change-Id: IB#1051946

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-18 Added follow up detection based on Message-ID e-mail header.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
  - 2016-04-29 Fixed parsing CSV data with quoted values containing newlines, thanks to Pawel Boguslawski.
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -10437,6 +10437,18 @@ Thanks for your help!
             </Hash>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="PostMaster::CheckFollowUpModule###0600-MessageID" Required="0" Valid="0">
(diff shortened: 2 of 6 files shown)


mgruner:
@jensoliver @nilei Can you please provide your opinion on this suggestion from Pawel? Would you agree?

nilei:
As it could be an intended function to create 2 tickets, at least the ticket status and ticket queue should be considered too, to not update and reopen closed tickets. The more restriction we can set (type, service, sla, dynamic fields, etc.) the better we can adjust the module to the user requirements. The module should be inactive by default.

Cheers, Nils

jensoliver:
Hi,

as the behaviour of OTRS depends on the mailsetup I also suggest to have it deactivated by default and also have the possibilitiy to restrict it to some states of the ticket

Cheers

Jens

mgruner:
Hello @pboguslawski. Thank you very much for your suggestion! Would you consider extending this PR to allow for additional configuration of the module, and adding unit tests for it? In this case we'll review and merge it.

pboguslawski:
Hi,

> Would you consider extending this PR to allow for additional configuration of the module

In our opinion this module is more obvoius than merging messages using Related headers for example, because Message-ID should be globally unique. This should work same like Related stuff if enabled (its already deactivated by default) so no extra checks should be introduced in our opinion (queues, states, etc.). Keep it simple. Without it, in large systems with many system addresses agents may have no chance to see duplicates and two same problems will be processed in two different tickets which may cause problems/waste of time.

We use this mod with success for many months now and there was no need for extra parameters. Feel free to extend it of course.

Regards,
Pawel

nilei:
@mgruner, will this be merged into master?
Cheers, Nils

mgruner:
@nilei no, as this PR is not in a mergeable state.

PR #1171, opened 2016-05-18
Change wording for better translation

urbalazs:
Hi @mgruner 
I changed wording of this sentence because in Hungarian (and maybe in some other languages) this string has to be translated with different word order.


mgruner:

PR #1174, opened 2016-05-19
7 Day Stats performance optimizations

pboguslawski:
OTRS recalculates all data for 7 Day Stats graph every 30 min which
may cause performance problems in bigger systems with lot of ticket
and agents, when graph data expires. Data for past days don't change
so it may be cached for longer time. This mod forces 7 Day Stats
data for past days to be cached for 8 days - only todays data
is cached for 30 min.

This mod adds also hint for MySQL database query optimizer to avoid
its wrong decisions and ticket_history table scans during
7 Day Stats data generation.

Related: https://dev.ib.pl/ib/otrs/issues/58
Author-Change-Id: IB#1020711

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-19 7 Day Stats performance optimizations.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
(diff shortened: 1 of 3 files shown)


pboguslawski:
Hi,

In case you still don't like DB specific optimizations in OTRS code (http://bugs.otrs.org/show_bug.cgi?id=10380) please consider merging only CacheTTL optimization in TicketStatsGeneric.pm.

Regards,
Pawel

mgruner:
Merged and backported to rel-5_0 without the mysql specific change. Thanks @pboguslawski!

PR #1097, opened 2016-03-30
Fixed link recognition

pboguslawski:
This allowes OTRS to recognize links like

    http://www.server.nl:80/%7Eguido/Python.html
    https://aa.bb.com/wiki/Obs%C5%82uga_ABC#Sekcja
    http://msdn.microsoft.com/en-us/library/windows/hardware/ff557211%28v=vs.85%29.aspx
    https://git.proxmox.com/?p=qemu-server.git;a=commit;h=445f06cdb193557c72b5aa83d347fbad1acac283

Related: https://dev.ib.pl/ib/otrs/issues/36
Author-Change-Id: IB#1008598

PR #1176, opened 2016-05-20
[ Rel-5 Bug 8761 ] - OTRS doesn't recognize URLs which contain nested URLs

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8761

Hello @mgruner ,

Hereby is proposal for fixing this bug. There was an issue where nested URL were not correctly recognised by LinkQuote() regex. It was ignoring characters that are used for nested URL  (e.g. : . / ) when '?' is matched after first level of URL.

Modified unit test to check for this behaviour. Please let me know if you have any issues or questions regarding this fix.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/HTMLUtils.pm
+++ b/Kernel/System/HTMLUtils.pm
@@ -827,11 +827,11 @@ sub LinkQuote {
             )*
             (?:                                    # param string
                 [\?]                               # if param string is there, "?" must be present
-                [a-zA-Z0-9&;=%\-_]*                # param string content, this will also catch entities like &amp;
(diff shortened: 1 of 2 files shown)


mgruner:
See also #1097 (probably merge together).

s7design:
Hello @mgruner ,

Thank you for pointing out PR https://github.com/OTRS/otrs/pull/1097 . Updated this PR with case where in URL is port number and expanded file path element part of regex. Excluded last UT case from mentioned PR, since it directly reflect issue on which this bug was open and it is originally covered in this PR.

Regards,
Sanjin

mgruner:
Merged in e68d6b96965ca71353936ee0f18e333a65a9a47c, thanks a lot @s7design.

PR #1177, opened 2016-05-20
Marked titles as translatable in column names

urbalazs:
Hi @mgruner 
This PR makes translatable the title attributes of the headers of the ticket tables on the dashboard and the headers of ticket small view on status view. These strings are only displayed, if you hover the mouse over the table headers. Now it also translates the sorted and filtered headers.
Rel-5_0 is also affected.


mgruner:
Merged in ebdc34bc10f6b3b1a87ee2773cb44e403a8f4854 and backported to OTRS 5. Thanks a lot!

PR #1179, opened 2016-05-23
[ Master Bug 8822 ] - Queue system address settings does not follow the postmaster mail account settings in some cases

s7design:
http://bugs.otrs.org/show_bug.cgi?id=8822

Hi @mgruner 

Hereby is fixed this bug, this is only our proposal. Actually, there is updated SystemAddressUpdate and there is not possible to set to invalid SystemAddress which there is as system address for a Queue.

There are updated unit test and Selenium test as well.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminSystemAddress.pm
+++ b/Kernel/Modules/AdminSystemAddress.pm
@@ -91,14 +91,14 @@ sub Run {
         # if no errors occurred
         if ( !%Errors ) {
 
+            my $Success = $SystemAddressObject->SystemAddressUpdate(
+                %GetParam,
+                UserID => $Self->{UserID},
+            );
+
             # update email system address
-            if (
-                $SystemAddressObject->SystemAddressUpdate(
-                    %GetParam,
-                    UserID => $Self->{UserID},
-                )
-                )
-            {
+            if ( $Success ){
+
                 # if the user would like to continue editing system e-mail address, just redirect to the edit screen
                 # otherwise return to overview
                 if (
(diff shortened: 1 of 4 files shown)


mrcbnsls:
Could you please rebase the PR, @s7design? Thanks!

s7design:
Hi @mrcbnsls 

Yes of course it is no problem

Regards
Zoran

s7design:
Hi @mrcbnsls 

I rebased this PR, there is added notification for update as well, because I need notification for this situation if there is set system address to some queue(s). I used approach that we use for such notification in other modules. 

I saw that you did not merge code for notification for update in commite:

https://github.com/OTRS/otrs/commit/83f950712b2c71fc13e167b84b029d884184aa75

in our PR there was this notification

https://github.com/OTRS/otrs/pull/1472/files

If you don't wont this notification please let me know I can remove it.

P.S. I have one question, and there would be nice if you have any suggestion:
https://github.com/OTRS/otrs/pull/1179/files#diff-3a9609fb8faadaa09102320a2420099cR249

What do you think that we change this code and return some  parameter if there is not possible to set system address, because if there is no any messages maybe there could be a problem if there would be error in next step if SQL query is not successful for any reasons.  

maybe it could be:

`return 'QueueAlredySet' if $Queue{SystemAddressID} eq $Param{ID};`

Regards
Zoran

mrcbnsls:
Hi Zoran,

I currently don't understand exactly the use of this changed return statement. Where would you make use of it later? Currently the code would just exit there, right? So the following SQL would not be executed at all AFAICS.

Marc

s7design:
Hi @mrcbnsls 

Yes you are right, the code just exit  here now:

`return if $Queue{SystemAddressID} eq $Param{ID};`

I suggest
`
return 'QueueAlredySet' if $Queue{SystemAddressID} eq $Param{ID};`

It could be use here:
https://github.com/OTRS/otrs/pull/1179/files#diff-be45239ff8d878af32150e4c40cf2d3cR114
Everything is OK if there is not a problem with SQL DB UPDATE query execution. But if there is no connection with DB, for example, here will be soem erro message, but the code `my $Notification = $Success ? 'Update' : 'NotPosibleSetInvalid'; `could  show wrong message.

Instead this code:
`my $Notification = $Success ? 'Update' : 'NotPosibleSetInvalid';`

could be something like:
```
if($Success == 1){
   #handle if everything OK and SystemAddressUpdate return '1'
...
}
elsif($Success eq 'QueueAlredySet'){
# handle f the system address is connected with some queue
...
}

# if there is error in DB update, the statement below will be run
# in part:
.... # something has gone wrong:

   # update system address
       # return if !$Kernel::OM->Get('Kernel::System::DB')->Do(
....


```

I tried to explane what I wanted, if it is not clear enough, please let me know. Maybe it is not so important I just mentioned if we should think about this detail to make better code. If I am wrong we always could merge the code as it is now.

Regards
Zoran

s7design:
It is fixed here

https://bugs.otrs.org/show_bug.cgi?id=6144

PR #1180, opened 2016-05-23
Master - JS refactoring - part1

s7design:
Hi @zottto and @mrcbnsls 

This is first PR for the test where we will refactoring JS code in the framework. As Marc suggested I started with removing global config params from AddConfig call.
There is added AddJSData in Layout Footer method, I think it is OK because these params are used in whole framework. If it is OK I can continue and do the same with other config items in FooterJS. I am planing to check if I can put somewhere else AddJSData if it is not necessary to be globall for all screens ( for example IncludeUnknownTicketCustomers it is used in AgentCustomerInformationCenterSearch and AgentCustomerSearch  but also in Core.Agent.Dashboard.js and Core.Agent.TableFilters.js) Should I put this one also in Layout Footer.

I also removed:
```
        // translations
-        BrowserListMsg: [% Translate("OTRS runs with a huge lists of browsers, please upgrade to one of these.") | JSON %],
-        BrowserDocumentationMsg: [% Translate("Please see the documentation or ask your admin for further information.") | JSON %],
```
These two config are not used anywhere. 

Please let me know if I can improve it or if I on good way at all :)

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -1459,9 +1459,31 @@ sub Footer {
     my $Type          = $Param{Type}           || '';
     my $HasDatepicker = $Self->{HasDatepicker} || 0;
 
-    # Generate the minified CSS and JavaScript files and the tags referencing them (see LayoutLoader)
+    # generate the minified CSS and JavaScript files and the tags referencing them (see LayoutLoader)
     $Self->LoaderCreateAgentJSCalls();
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Merged in 6694b3c, thanks, looks good for me. For things which are used in more than one JS file, I'd suggest that we also include them in the footer for now, e.g. IncludeUnknownTicketCustomers.

PR #1181, opened 2016-05-24
Whitelisted characters allowed in filenames for security

pboguslawski:
OTRS uses bad chars blacklisting in attachment filename creation. Best
practice is to use whitelisting for security (to avoid possible other
bad chars in filenames).

This mod changes OTRS behaviour in this ares - only whitelisted
characters are allowed in filenames.

This mod also moves filename cleanup logic to one place -
the FilenameCleanUp() function - and fixes leading dots removing regexp
(was removing only first leading dot in filename).

Related: https://dev.ib.pl/ib/otrs/issues/59
Related: http://bugs.otrs.org/show_bug.cgi?id=10395
Author-Change-Id: IB#1016067

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
-#6.0.0.beta1 2016-??-??
+#6.0.0.beta1 2016-??-??  
+ - 2016-09-30 Attachment name validation added.
  - 2016-05-24 Whitelisted characters allowed in filenames for security.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-24 Whitelisted characters allowed in filenames for security.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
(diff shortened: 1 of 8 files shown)


pboguslawski:
Hi,

In case you still don't like whitelisting (http://bugs.otrs.org/show_bug.cgi?id=10395) please consider merging only "strip dots from filenames" regexp fix and moving filename cleanup logic to one place /FilenameCleanUp()/.

Regards,
Pawel

pboguslawski:
OTRS does not validate attachment names passed in e-mail headers and upload requests before using this values for filenaeme creation in filesystem; this is risky (possible escaping quirks).

There is also problem with missing attachments with long names; if one try to send e-mail from Windows with attachement

    C:\tmp\0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę0123456ąćę001.txt

OTRS baked on XFS will throw an error

    Message: Can't write '[...]0123456ąćę0123456ąćę012[...]001.txt.content_type': File name too long

and attachment won't be present in ticket. Similar problem when agent tries to upload file as above from Windows workstation to OTRS baked on XFS.

Commits 4c9b901c47f234e08a083c2f8bac36eedd66b4cb (rel-5_0) and 459ae466100b66d8f264b7d52d832103d2dcce7d (master) fix both issues described above by filtering attachment names using Kernel::System::Main::FilenameCleanUp function that was modified in

https://dev.ib.pl/ib/otrs/issues/59

https://github.com/OTRS/otrs/pull/1181

to limit filename length to 100 chars and to whitelist allowed characters.

s7design:
Hi,

Thanks for PR, we use some of your suggestion to fix bug 

https://bugs.otrs.org/show_bug.cgi?id=11132

You can see commits there 

Regards
Zoran

PR #1182, opened 2016-05-24
Master - JS refactoring - part2

s7design:
Hi @mrcbnsls ,

Hereby is finished removing  Core.Config.AddConfig  from FooterJS and CustomerFooterJS.
Please let me know if it is done well or we can improve the code.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -1499,6 +1499,10 @@ sub Footer {
             = $Self->{LanguageObject}->Translate( $AutocompleteConfig->{$ConfigElement}->{ButtonText} );
     }
 
+    my $AutocompleteConfigJSON = $Self->JSONEncode(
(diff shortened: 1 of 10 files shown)


s7design:
Hi @mrcbnsls 

You can see I updated this PR, the first commit I have done because there was error, and in the last one I added JSON encode as it has been before.
It works in any cases but you will see if I made some mistake. Please let me know if it is wrong and I will be able to fix that.

Regards
Zoran

mrcbnsls:
Merged in 4cd7c52, thanks.

As you can see, there were some issues which I fixed, I'd assume due to copy & paste. I added two new UT to have a very basic check that the data transfer via Layout.pm is working correctly.

s7design:
Thanks Marc for follow-up, we will be more careful. It is rely rookie mistakes. We will try to avoid such ones.

Regards
Zoran

PR #1183, opened 2016-05-24
Master - JS refactoring - AdminDynamicFieldText

s7design:
Hello @zottto  and @mrcbnsls ,

Hereby is refactored JS for AdminDynamicFieldText.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminDynamicFieldText.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminDynamicFieldText.tt
@@ -245,25 +245,3 @@
         </form>
     </div>
 </div>
-[% WRAPPER JSOnDocumentComplete %]
-<script type="text/javascript">//<![CDATA[
-$('.ShowWarning').bind('change keyup', function (Event) {
-    $('p.Warning').removeClass('Hidden');
-});
-
-// click handler to remove regex
-$('.RemoveRegEx').bind('click', function () {
-    Core.Agent.Admin.DynamicFieldText.RemoveRegEx( $(this).attr('id') );
-    return false;
-});
-// click handler to add regex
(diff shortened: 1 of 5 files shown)


s7design:
Hi @mrcbnsls 

We are planing to add JS unit test for Core.Agent.Admin.DynamicFieldText.js because it is not exist.

Regards
Zoran

s7design:
Hi @mrcbnsls ,

Included test for Core.Agent.Admin.DynamicFieldText.js 
 
Please if you have any comment, advice or suggestion how can i improve this test let me know.

Regards,
Sanjin

mrcbnsls:
Merged in 178d08c.

Regarding the new JS-UT: Thanks for the effort, but I'd prefer to add every possible test case to the existing selenium test. The JS-UT are meant for testing the generic libraries (such as Core.Form.js, Core.Ajax.js, etc.). These libraries provide generic methods which can be used from any screen while Core.Agent.Admin.DynamicFieldText.js is more like a supplemental library which is meant for only one screen. If we successfully test the screen functions using a selenium test (which we do with AdminDynamicFieldText.t), we'll also know the JS library is working correctly.

s7design:
Hi @mrcbnsls 

Thanks a lot for your follow-up. We needed such feedback. I like your comment in general, because you help us that we know what is the best workflow and way how we should work on the task.

I totally agree with you. I spoke with my colleagues on other project as well about JS testing and how deep we should go in such test. We have Selenium test where is done complete framework testing and I could say some kind of behavior testing is some cases. So in JS is enough test global JS scripts and functions which are used in many screens.

As I mentioned in some previous comments, there will be maybe some similar cases where we had dilemmas but with your help we will be able to solve any of them :)

Regards
Zoran

PR #1184, opened 2016-05-25
Fixed premature end of script headers error when using long URL

pboguslawski:
If you use mod_perl and you paste the following URL to your web browser
(without active OTRS session)...

    https://[yout otrs domain here]/otrs/index.pl?Action=AgentTicketQueue;QueueID=12;Filter=Unlocked;View=;SortBy=Age;OrderBy=Up;StartWindow=0;StartHit=21;TicketID=111378;TicketID=111647;TicketID=111941;TicketID=111985;TicketID=234056;TicketID=234056;TicketID=234079;TicketID=234178;TicketID=234214;TicketID=234214;TicketID=234255;TicketID=234459;TicketID=234434;TicketID=234428;TicketID=234865;TicketID=234728;TicketID=234786;TicketID=234786;TicketID=234957;TicketID=410006;TicketID=410006;TicketID=410161;TicketID=410199;TicketID=410319;TicketID=410292;TicketID=410581;TicketID=410618;TicketID=410587;TicketID=410907;TicketID=411009;TicketID=411028;TicketID=411184;TicketID=411363;TicketID=411375;TicketID=411354;TicketID=411354;TicketID=411381;TicketID=411426;TicketID=411596

...OTRS will ask you for login and password. After passing correct login
data, OTRS fires "Premature end of script headers: index.pl" error
on redirection.

The problem occurs if HTTP redirection response after login contain
HTTP headers that (all together) are longer than 1024 bytes.
The problem does not occur if you rollback change introduced in
28996cf88f8a57e70d52fed7ab8963b7f1b4771b.

This workaround switches STDOUT from :encoding(utf8) to :utf8 only
for redirection after successfull login.

Fixes: 28996cf88f8a57e70d52fed7ab8963b7f1b4771b
Related: https://dev.ib.pl/ib/otrs/issues/60
Related: http://bugs.otrs.org/show_bug.cgi?id=12100
Author-Change-Id: IB#1052183

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-05-25 Fixed premature end of script headers error when using long URL.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
  - 2016-04-29 Fixed parsing CSV data with quoted values containing newlines, thanks to Pawel Boguslawski.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-06-07 Fixed premature end of script headers error when using long URL.
  - 2016-05-25 Fixed premature end of script headers error when using long URL.
  - 2016-05-04 Added the possiblity to configure the responsible field as mandatory (enabled by default for AgentTicketResponsible, if responsible feature is enabled), thanks to S7.
  - 2016-04-29 Reduced error log noise by reducing the log level of less important messages, thanks to Pawel Boguslawski.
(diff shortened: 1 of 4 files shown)


pboguslawski:
This is workaround for http://bugs.otrs.org/show_bug.cgi?id=12100 that works for us.

pboguslawski:
Similar problem may occur in other places where OTRS tries to redirect to long URL (i.e. after moving ticket to queue with move_into permission, if last overview page had long URL).

Just commited more general solution to avoid such header buffering problems for outputs.

According to http://bugs.otrs.org/show_bug.cgi?id=5158 commit 28996cf88f8a57e70d52fed7ab8963b7f1b4771b was introduced to avoid security problems connected with input encoding. This mod restores original encoding :utf8 for outputs (i.e. STDOUT and STDERR) which resolves headers buffering problems and should be safe (separate safer function SetI() with :encoding(utf8) was provided in case one needs to use it for input encoding).

mgruner:
Solved with a slightly different solution. Please let me know if you see any problems there.

mgruner:
http://bugs.otrs.org/show_bug.cgi?id=12100

PR #1185, opened 2016-05-25
Master - JS refactoring - Modify AdminACL

s7design:
Hi @mrcbnsls 

Hereby is updated AdminACl. We removed script from TT module to Core.Agent.Admin.ACL.js. There is updated Selenium test where we add test cases for the filter.

We would like to make new JS unit test for Core.Agent.Admin.ACL.js. but we are not sure how it is hard in this moment. We will try this as well, but if it is so hard we will continue with refactoring the next TT module

Regards
Zoran


mrcbnsls:
Merged in 7feb772, thanks. Would be nice if you could extend the ACL UT step by step, but it's not urgent I think.

PR #1186, opened 2016-05-26
Master - fix mistake in AJAX JS unit test

s7design:
Hi @mrcbnsls 

There is fix for Core.AJAX.UnitTest.js. I investigated  JS tests and I found this small mistake.

Regards
Zoran


mrcbnsls:
Merged in d3fdf99, thanks!

PR #1187, opened 2016-05-26
Master - JS refactoring - Modify AdminACLEdit

s7design:
Hi @mrcbnsls

Hereby is updated AdminACLEdit. We removed script from TT module to Core.Agent.Admin.ACL.js. 

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminACL.pm
+++ b/Kernel/Modules/AdminACL.pm
@@ -795,6 +795,27 @@ sub _ShowEdit {
         }
     }
 
+    # create data for JS
+    my @ACLEditPossibleActionsList;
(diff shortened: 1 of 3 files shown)


mrcbnsls:
Thanks, merged in 92bd689. I think it makes sense to have the Subaction also available via Core.Config.Get, so we can use it in ACL.js to see if we want to trigger InitACLEdit or not. Seems somehow more comprehensible to me.

s7design:
Yes I agree it is much better. We had such idea. you will see in other PRs, some things could be done on different way. I always choose one, but we can discus to find the best one. 
Thanks for feedback, I like when you told me your opinion since I can learn something new.

Regards
Zoran

PR #1191, opened 2016-05-27
Fixed Internal Server Error on attachment removal page reload

pboguslawski:
If you remove file from message form (i.e. in AgentTicketEmail) and then
refresh page (ignoring web browser warning) OTRS will throw Internal
Server Error and apache error log will contain message similar to

    Can't use an undefined value as a HASH reference at /opt/otrs/Kernel/System/Web/UploadCache/FS.pm line 177.\n

This mod fixes handling such errors in OTRS.

Related: https://dev.ib.pl/ib/otrs/issues/61
Author-Change-Id: IB#1052184


mgruner:
Merged and backported, thanks a lot @pboguslawski.

PR #1192, opened 2016-05-27
Master - JS refactoring - Modify AdminGroup

s7design:
Hi @mrcbnsls

Hereby is updated AdminGroup. We removed script from TT module to Core.Agent.Admin.Group.js, which is created. There is also JS unit test where is checked AdminGroupCheck function.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -9061,6 +9061,9 @@ via the Preferences button after logging in.
                     <Description Translatable="1">Create and manage groups.</Description>
                     <Prio>150</Prio>
                 </NavBarModule>
+                <Loader>
+                    <JavaScript>Core.Agent.Admin.Group.js</JavaScript>
+                </Loader>
             </FrontendModuleReg>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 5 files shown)


s7design:
Hi @mrcbnsls 

I the second commit there is added submit event as well as there is for the click on Save button.
In my opinion it is better behavior.

Regards
Zoran

mrcbnsls:
Merged in 0f02fbb. As you can see, I did some bigger changes on this.

- We're no longer using bind(), but on() instead
- Using on() on both the submit button and the form itself triggers the function actually two times
- The check for the group name was not working. Even when clicking "cancel" in the confirm dialog, the form would still be submitted. I changed this to use the proper function we have for this (Core.Form.Validate.SetSubmitFunction()) and to use a dialog instead of a confirm() because the latter was having issues with SetSubmitFunction.
- I added a section in the selenium UT to check for changed admin group name

s7design:
Hi @mrcbnsls 

Thanks a lot. You again helped us that we clarify something. I also wanted to update Selenium test but I have a problem to check confirmation message. I think your solution is much better. We will use it in further work.

Regards
Zoran

PR #1193, opened 2016-05-27
Master - JS refactoring - Modify AdminAutoResponse

s7design:
Hi @mrcbnsls

Hereby is updated AdminAutoResponse. We moved script from TT module to perl module with AddJSOnDocumentComplete call. I think it is better solution then creating new JS only for Init function with:
`
Core.UI.Table.InitTableFilter($("#FilterAutoResponses"), $("#AutoResponses"))`

Please let me know if you have any suggestions.  

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -9095,6 +9095,9 @@
                     <Block>Queue</Block>
                     <Prio>400</Prio>
                 </NavBarModule>
+                <Loader>
+                    <JavaScript>Core.Agent.Admin.AutoResponse.js</JavaScript>
+                </Loader>
             </FrontendModuleReg>
         </Setting>
     </ConfigItem>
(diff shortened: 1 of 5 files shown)


mrcbnsls:
Hi Zoran,

AddJSOnDocumentComplete is mainly meant for modules which don't have their own frontend module, such as the different dynamic fields which bring their own JS. Please create a separate JS file for AdminAutoResponse. Yes, it will only contain this one call for now, but its fine. Its just more consistent.

Should the same be the case for other modules, please update the according PRs, too. Thanks!

Marc

mrcbnsls:
Sorry, closed unintentionally.

s7design:
Hi @mrcbnsls 

Thanks, no problem we will be able to update all PR where it is necessary to add new JS file.

Regards
Zoran

s7design:
Hi @mrcbnsls

There is updated PR accordingly your suggestion.
JS code is moved to Init function of new JS file.
Also, new block for filter in TT module is added.

Regards
Zoran

mrcbnsls:
Perfect, merged in 3b27132. Thanks!

PR #1194, opened 2016-05-27
Master - JS refactoring - Modify AdminAttachment

s7design:
Hi @mrcbnsls

Hereby is updated AdminAttachment. We moved script from TT module to Perl module with AddJSOnDocumentComplete call. There is used the same approach as in the PR for AdminAutoResponse.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -9135,6 +9135,9 @@
                     <Block>Queue</Block>
                     <Prio>600</Prio>
                 </NavBarModule>
+                <Loader>
+                    <JavaScript>Core.Agent.Admin.Attachment.js</JavaScript>
+                </Loader>
             </FrontendModuleReg>
         </Setting>
     </ConfigItem>
--- a/Kernel/Modules/AdminAttachment.pm
+++ b/Kernel/Modules/AdminAttachment.pm
@@ -260,12 +260,6 @@ sub Run {
         $Self->_Overview();
         my $Output = $LayoutObject->Header();
         $Output .= $LayoutObject->NavigationBar();
-
-        # add JS code
-        $LayoutObject->AddJSOnDocumentComplete(
-            Code => 'Core.UI.Table.InitTableFilter($("#FilterAttachments"), $("#Attachments"))',
-        );
-
         $Output .= $LayoutObject->Output(
             TemplateFile => 'AdminAttachment',
             Data         => \%Param,
@@ -345,6 +339,9 @@ sub _Overview {
     );
     $LayoutObject->Block(
         Name => 'ActionAdd',
+    );
+    $LayoutObject->Block(
+        Name => 'Filter',
     );
     $LayoutObject->Block(
         Name => 'OverviewResult',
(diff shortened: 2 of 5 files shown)


mrcbnsls:
Hi Zoran,

Same goes here as for PR#1193. Please update.

Also, could you please explain a bit about the changes on the UT? I could see that you changed many wordings and variable names (e.g. $RandomID to $AttachmentName). Whats the purpose of it? 

Marc

s7design:
Hi @mrcbnsls 

We will update this one also. Regarding to change $RandomID to $AttachmentName, there is new variable

my $AttachmentName = 'StdAttachment' . $Helper->GetRandomNumber();

there is used random id as well. It is not so important. There is not changed behavior of the test. Actually, the main changes is to add test case where we test the filter. We add test case which filter by the extension of the attachments. If you don't like this test case we can change it.

Regards
Zoran

mrcbnsls:
Ok, thanks for the feedback.

I feel that we should avoid such cosmetical changes to the tests which are not really necessary in this turn. Let's focus on the changes regarding the JS refactoring (which includes functional updating of UT of course).

s7design:
Hi @mrcbnsls

There is updated PR accordingly your suggestion. JS code is moved to Init function of new JS file. New block for filter in TT module is added too.

Regards
Zoran

mrcbnsls:
Thanks for the updated PR. I have some very bad issues with the UT right now which I were not able to solve yet. It seems as if the delete link can't be clicked. According to the selenium log the link is enabled and displayed, but clicking it fails. I think this is related to my setup, but maybe you are experiencing this, too. I already tried changing the layout (z-index etc.), but it didn't help...

s7design:
Hi @mrcbnsls 

Thanks for follow-up. We are working on this. To be honest we are not sure what was a problem, but I hope it will be solved very soon.

Regards
Zoran

mrcbnsls:
So you're having the same issues, @s7design?

s7design:
No, it woks fine for us. I am planing to find what is the posible problem. Maybe your Selenium error log could help us. You can send me on Skype or mail if it is easier for you.

Regards
Zoran

s7design:
Hi @mrcbnsls 

There was a problem with the filter which was used before test case where Attachment should be deleted. Sometimes this sleep is not necessary, it occurs in my system after reinstalling vanilla system. It works fine for me.

Regards
Zoran

mrcbnsls:
Thanks a lot, that was indeed the problem. It's pretty obvious, but I was not able to find that it was caused by the filter because in the screenshot which was created by Selenium, all attachments were visible. It seems that this is because there is a short delay between the failing test and the creation of the screenshot which was why I didn't recognize it... big thanks, merged now in 869259e. I also added a piece of code to let the user confirm they really want to delete the attachment. I think it's a usability issue that we still have places in OTRS where clicking a delete icon/button instantly deletes the item without asking again. If you stumble upon such places, please feel free to add something similar to what I did in Core.Agent.Admin.Attachment.js :-)

s7design:
Thanks you Marc. Yes, there are always such tricky cases when we write Selenium test. Many things can influence on the behavior of the system. It is funny that this one a little sleep solved the problem :) 
I can say we are lucky, because sometimes we waste our time much longer. 
Screenshots is good feature, but sometimes it is useless.  
But the most important we managed to fix that. 

I think that this confirm dialog is very good solution, we will add it if we find such cases.
Thanks, I like cooperation with you :+1: 

Regards
Zoran

PR #1197, opened 2016-05-30
Master - JS refactoring - Modify AdminCloudServices

s7design:
Hi @mrcbnsls 

Hereby is updated AdminCloudServices.tt. There was JS snippet. We decided to create new JS   Core.UI.MasterActionLink.js and then we are planning to add this one in all modules where there is such code ( I think there is the same code in 11 TT files)
Of course it could be solved on many other ways, if you would like different approach please let me know.

We can also extend Selenium test and add JS unit test if you like it.
Your follow-up is welcome in this moment.

Regards
Zoran


mrcbnsls:
Hi Zoran,

Removing the code duplications is a very good idea, agree on that. But I'd rather suggest to have a new method InitMasterAction() in Core.UI. I don't think its necessary to have a whole new file just for this method. What do you think? If you agree, please update all affected PRs.

Marc

s7design:
Hi @mrcbnsls

There is updated PR accordingly your suggestion. JS code is moved to the new function InitMasterAction in Core.UI.

Regards
Zoran

s7design:
Hi @mrcbnsls 

There was a problem with the test CustomerUser.t so we rebased this PR. We are going to rebase other PRs with the same problem.

Regards
Zoran

mrcbnsls:
Merged in 5983311, thanks.

PR #1198, opened 2016-05-30
Master - JS refactoring - Modify AdminCustomerCompany

s7design:
Hi @mrcbnsls

Hereby is updated AdminCustomerCompany.tt. There we made new JS Core.UI.MasterActionLink.js as well as in the PR https://github.com/OTRS/otrs/pull/1197. Actually, it is enough to add in one PR, but we will add in all PRs until one of them is not merged.

If you have any other suggestion we will be able to change that.

Regards
Zoran


s7design:
Hi @mrcbnsls

There is updated PR accordingly your suggestion. JS code about disabling form is moved to the new JS file - eliminate AddJSOnDocumentComplete. The code regarding the MasterAction stuff is moved to the function InitMasterAction in Core.UI.

Regards
Zoran

Kernel/Modules/AdminCustomerCompany.pm
```diff
@@ -505,11 +514,9 @@ sub _Edit {
     );
 
     # Get needed objects.
-    my $ConfigObject              = $Kernel::OM->Get('Kernel::Config');
     my $ParamObject               = $Kernel::OM->Get('Kernel::System::Web::Request');
     my $DynamicFieldBackendObject = $Kernel::OM->Get('Kernel::System::DynamicField::Backend');
 
-    ENTRY:
```
mrcbnsls: I think this was deleted by accident, wasn't it?

mrcbnsls:
Merged in b5f882e, thanks.

I think rather than doing
    var ReadOnly = Core.Config.Get('ReadOnly');
    if (ReadOnly == 1) {
it's better to do something like
    if (parseInt(Core.Config.Get('ReadOnly'), 10) === 1) {
        Core.Form.DisableForm($("form#edit"));
    }
which you also did in other cases.

PR #1200, opened 2016-05-31
 Master - JS refactoring - Modify AdminCustomerUserGroup

s7design:
Hi @mrcbnsls

Hereby is updated AdminCustomerUserGroup.tt. As you can see there is added Core.Agent.Admin.Checkbox.js for initialization all checkboxes in such screens. If you like this solution we are planing to solve the other relation screens the same.

We are also planning to add JS unit test for this.

Regards
Zoran


mrcbnsls:
Hi Zoran,

Well, we already have Core.Form.InitSelectAllCheckboxes(), which was also used in this screen before your changes. Why did you decide to create a new lib with only this function (which is actually at least partly a code duplication with the existing function)? Maybe there is something I don't see or understand yet.

Also, as in other PRs, I'd ask you to create a separate JS lib for AdminCustomerUserGroup rathern than adding the JS code using AddJSOnDocumentComplete.

Thanks!

Marc

s7design:
Hi @mrcbnsls

There is updated PR accordingly your suggestion. JS code is moved to the new JS file.

Regards
Zoran

mrcbnsls:
Merged in 7efff9e, thanks!

PR #1201, opened 2016-05-31
Master - JS refactoring - Modify AdminCustomerUserService

s7design:
Hi @mrcbnsls 

Hereby is updated AdminCustomerUserService. There is very similar approach  as in the PR https://github.com/OTRS/otrs/pull/1200

Regards
Zoran


mrcbnsls:
Please see my comment on #1200 :-)

s7design:
Hi @mrcbnsls 

There is updated PR in accordance with your instructions. All JS code is moved to the new JS file.

Regards
Zoran

mrcbnsls:
Merged in e8acec6, thanks.

Please note my changes on the JS and PM files. Though I appreciate your effort to have a consistent solution across the screen regarding the RelationItems, I'd prefer to have the code as clean as possible. As there is only one fixed item, I think there is no need to add it via AddJSData. Do you agree on this?

PR #1203, opened 2016-05-31
Master - JS refactoring - Modify AdminSalutation

s7design:
Hi @mrcbnsls 

Hereby is updated AdminSalutation. I did not extend AdminSalutation with the test for the filter. We are also planing to cover JS unit tests for all JS and there we will test InitTableFilter as well. However, this filter is tested in other screens ( e.g. here also https://github.com/OTRS/otrs/pull/1193/files )

If you would like we will be able to add this test case in  AdminSalutation.t 

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -9175,6 +9175,9 @@
                     <Block>Queue</Block>
                     <Prio>800</Prio>
                 </NavBarModule>
+                <Loader>
+                    <JavaScript>Core.Agent.Admin.Salutation.js</JavaScript>
+                </Loader>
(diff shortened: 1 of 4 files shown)


mrcbnsls:
Please create a separate JS file here, too. Thanks!

s7design:
Hi @mrcbnsls 

This PR is updated as well.

Regards
Zoran

mrcbnsls:
Merged in f8d2c74, thanks. I also added a filter block here.

mrcbnsls:
Just saw that the namespace was not named correctly, which I fixed in b809b00. Please make sure that the namespace always matches the right frontend! Thanks!

PR #1204, opened 2016-05-31
Master - JS refactoring - Modify AdminQueueTemplates

s7design:
Hi @mrcbnsls 

There is updated AdminQueueTemplates.tt
As you can see we add two JS calls in AdminQueueTemplates.pm, I suppose it could be done with one AddJSOnDocumentComplete like:

> 
>     # add JS code
>     $LayoutObject->AddJSOnDocumentComplete(
>         Code => 'Core.Agent.Admin.Checkbox.InitSelectAllCheckboxes();
>         Core.UI.Table.InitTableFilter($("#Filter"), $("#ItemsTable"));',
>     );

What do you think about that, it is my proposal, but I will be able to change  this.

Regards
Zoran


mrcbnsls:
Hi Zoran,

As discussed in other PRs, please add the calls to the according JS file instead of using AddJSOnDocumentComplete() and also change the checkbox usage (see my remarks on other PRs for this, too).

Thanks!

Marc

s7design:
Hi @mrcbnsls 

PR is modified in accordance with your suggestion. JS code is moved to the new JS file.

Regards
Zoran

mrcbnsls:
Merged in 97f5c69, thanks. Again I removed the AddJSData because it's not really needed.

PR #1205, opened 2016-05-31
Fixed invalid UTF-8 characters stripping when article displaying or forwarding; fixed possible log flooding

pboguslawski:
OTRS does not strip invalid UTF-8 characters when displaying article content
in ticket zoom and forwarding article. In case of Convert()-ing from utf-8
to utf-8 without Check=1 OTRS sets UTF-8 flag using Encode::_utf8_on without
checking if it's valid UTF-8 data which according to
http://perldoc.perl.org/Encode.html is bad idea.

This may lead to warnings in logs in case of invalid UTF-8 data from HTML
e-mails and even problems with OTRS stability (DOS) when invalid data hits
regexps in OTRS and perl is vunerable to CVE-2015-8853.

Sometimes in case of encoding errors OTRS prints whole converted text
in error message which may cause logs flooding (i.e. in case
of converting large HTML).

This patch fixes problems described above.

Related: https://dev.ib.pl/ib/otrs/issues/62
Author-Change-Id: IB#1017588


pboguslawski:
Consider backporting also; fix for OTRS5 is slightly different - see https://dev.ib.pl/ib/otrs/issues/62

Regards,
Pawel

mgruner:
Applied and backported, thanks!
https://bugs.otrs.org/show_bug.cgi?id=12415

PR #1206, opened 2016-05-31
Added additional Checks for GenericInterface

Rooobaaat:
Added checks for REST::Client and SOAP::Lite


mgruner:
Hm. SOAP::Lite is bundled, I believe. And with REST::Client, it seems that the package names are wrong. Could you check and correct that perhaps, @Rooobaaat?

mgruner:
No feedback, and both modules are bundled in Kernel/cpan-lib. Thanks for the suggestion anyway, @Rooobaaat.

PR #1211, opened 2016-06-01
Master - JS refactoring - Modify AdminTemplateAttachment

s7design:
Hi @mrcbnsls

There is updated AdminTemplateAttachment.tt

As you can see there were initialization for filters on two places for the same screen. When we move JS code in Perl, this code is added only once. It is not necessary two times. We will be able to add some IF statement and check which one of two filters is necessary (Template and/or Attachment). Please let me know, what do you think about that. 

```
[% WRAPPER JSOnDocumentComplete %]
<script type="text/javascript">//<![CDATA[
    Core.UI.Table.InitTableFilter($('#FilterTemplates'), $('#Templates'));
    Core.UI.Table.InitTableFilter($('#FilterAttachments'), $('#Attachments'));
//]]></script>
[% END %]
```

Regards
Zoran


s7design:
Hi @mrcbnsls 

There is updated PR in accordance with your instructions. All JS code is moved to the new JS file.

Regards
Zoran

mrcbnsls:
Merged with small changes again in f419a54, thanks!

PR #1213, opened 2016-06-01
Master - JS refactoring - Modify AdminGenericInterfaceDebugger

s7design:
Hello @mrcbnsls ,

Hereby is refactored AdminGenericInterfaceDebugger. There was no Selenium test for this module, so created it where we test JS as well as some other module functionalities. 

Please if you have any additional cases you would like to see in the test or have some issues or comments regarding refactoring, let me know.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericInterfaceDebugger.pm
+++ b/Kernel/Modules/AdminGenericInterfaceDebugger.pm
@@ -50,9 +50,10 @@ sub Run {
         );
     }
 
-    # add JS code
-    $LayoutObject->AddJSOnDocumentComplete(
-        Code => "Core.Agent.Admin.GenericInterfaceDebugger.Init($WebserviceID);",
+    # send value to JS
+    $LayoutObject->AddJSData(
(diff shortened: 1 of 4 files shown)


s7design:
Hello @mrcbnsls ,

This PR is updated. Encountered issue while trying to use TargetNS.WebserviceID from Init function while reading values from perl with Core.Config.Get() to pass parsed WebServiceID to the rest of JS script. TBH i am not really sure why it was happening, changed to call WebServiceID from config in each function from perl instead using TargetNS.WebserviceID from Init and called variable from config in each function.

Perhaps if you could clarify this for me, i would be more then grateful.

Regards,
Sanjin

mrcbnsls:
Merged in d2c9528, thanks.

Regarding WebserviceID: Please have a look on how I changed it. Whats being called when the module is loaded is Init(). So I am just getting the WebserviceID there now and adding it to the namespace, so it's available for alle functions from TargetNS.

Marc

PR #1216, opened 2016-06-02
Master - JS refactoring - Modify AdminLog

s7design:
Hi @mrcbnsls 

Hereby is updated AdminLog.tt. We also add Selenium test for this module.

Regards
Zoran


mrcbnsls:
Merged without any complaints in 850c0a5, thanks :-D

PR #1217, opened 2016-06-02
Master - JS refactoring - Modify AdminMailAccount

s7design:
Hi @mrcbnsls 

Hereby is updated AdminMailAccount. There is extended the test AdminEmailAccount.t with some test. cases.

Regards
Zoran


mrcbnsls:
Merged in 8455079, thanks.

I changed several things:

- Renamed the selenium test for consistency
- Removed test for IMAPTLS because it relies on the optional perl module Mail::IMAPClient
- Moved code injected by AddJSOnDocumentComplete to JS file
- Found that the table filters are also shown in edit/add screen, which makes no sense, added a block for it

PR #1218, opened 2016-06-02
Master - JS refactoring - Modify AdminGenericInterfaceMappingSimple

s7design:
Hello @mrcbnsls and @mgruner 

Hereby is uncompleted refactoring for AdminGenericInterfaceMappingSImple. There is one issue i would to consult with you before completing this PR.

In AdminGenericInterfaceMappingSImple.pm in constructor there is value for 
`"$Self->{DeletedString}` =`'_GenericInterface_Mapping_Simple_DeletedString_Dont_Use_It_String_Please';` on line 25. This is changed in commit on July 19, 2011 ( https://github.com/OTRS/otrs/commit/f87b480c5a9c4bc39a16307003e327c90b418d79 ) Where it is removed taking value from config. IMHO there happen a mistakes, since from the looks of it, it was copy pasted value that was in config as a template where agent can modify and see how the config should be used. This parameter is used later in code to mark which string can't be used as values in mapping.

Please, correct me if my findings are wrong. Perhaps there is justified reason why this change happened and how this parameter should be used correctly. I will modify and complete this PR accordingly to your input on how to proceed with this matter.

Additionally, created Selenium test for this module as well. If you have any additional cases you would like to see in it, let me know.

Looking forward to your reply.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminGenericInterfaceMappingSimple.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminGenericInterfaceMappingSimple.tt
@@ -249,22 +249,12 @@
                             <button class="Primary CallForAction" type="submit" value="[% Translate("Save") | html %]"><span>[% Translate("Save") | html %]</span></button>
                             [% Translate("or") | html %]
                             <button class="CallForAction" type="submit" id="SaveAndFinishButton" value="[% Translate("Save and finish") | html %]"><span>[% Translate("Save and finish") | html %]</span></button>
-                            <input type="hidden" name="ReturnToAction" id="ReturnToAction" value="" />
-[% WRAPPER JSOnDocumentComplete %]
-<script type="text/javascript">//<![CDATA[
-$('#SaveAndFinishButton').bind('click', function(){
-    $('#ReturnToAction').val(1);
-});
-//]]></script>
-[% END %]
-
+                            <input type="hidden" name="ReturnToAction" id="ReturnToAction" value="" />s
                             [% Translate("or") | html %]
                             <a href="[% Env("Baselink") %]Action=[% Data.ActionFrontendModule | uri %];Subaction=Change;[% Data.ActionType | uri %]=[% Data.Action | uri %];WebserviceID=[% Data.WebserviceID | uri %]">[% Translate("Cancel") | html %]</a>
                         </div>
-
                     </fieldset>
                     <div class="Clear"></div>
-
                 </form>
             </div>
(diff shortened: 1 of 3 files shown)


mrcbnsls:
@carlosgarciac, as it was your commit back in 2011 :-D, could you probably say something about this

carlosgarciac:
Hi Marc, 

so sorry but I have not an opinion right now for this change, seems this change is not just a copy paste, similar variable is used in 'AdminDynamicFieldDropdown' and 'AdminDynamicFieldMultiselect' strings. But if you think is not necessary any more, please feel free to remove it.

Best regards,
Carlos García

s7design:
Hello @mrcbnsls and @carlosgarciac ,

If I may, would like to add some additional insight on this issue. Carlos is right when pointing out in similarity of variable use $Self->{EmptyString} in AdminDynamicField Dropdown/Multiselect module in comparison to AdminGenericInterfaceMappingSimple. They share common goal to prevent certain string from being keys in there respective fields. In DynamicFields it's the variable for 'Possible values' key while in GenericInterfaceMapping it's the value for 'Map value' for Value Mapping fields.

As it is now, this functionality is 'somewhat' working. If you try to enter string '_DynamicFields_DeletedString_Dont_Use_It_String_Please' as possible value key in DF Dropdown/Multiselect modules, you are allowed to submit form, but if you check for created DF you will see that this 'possible value' field is removed. If we allow agents to restrict creation of 'possible value' fields in such a way, IMHO then we should allow easier access to such configuration ( via SysConfig as one example ) instead of having value hard coded and not so easy to understand of configure.

Hope these comments clarified this issue a bit and gives you clearer view on how should we proceed further on. Mine suggestion is to switch these values to SysConfig and allow admin to define 'blacklisted' variables to use as key values on creation. Additionally would include some error message when such value is inputed that gives clear view why this variable is not available to use.

Looking forward to your reply.

Regards,
Sanjin

zottto:
@mrcbnsls / @s7design: Any update on this PR? What is the next step?

Thanks,
Marc

s7design:
Hi all,

As you can see in our previous comments, we have given some solution how it could be solved. Of course we can left it as it has been till now. It is no so important. It was only our suggestion what we find out working on this module. 

Please let me know what is the best solution and we will be able to fix it so.

Regards
Zoran

zottto:
Merged with a853719.

Please feel free to open another PR with the changes you recommend regarding "DeletedString".

Thanks,
Marc

PR #1221, opened 2016-06-02
Fixed error Undefined value as a SCALAR reference at UploadCache/FS.pm

pboguslawski:
OTRS may throw

    Can't use an undefined value as a SCALAR reference at [...]UploadCache/FS.pm[...]

error to apache error log if two HTTP requests call same OTRS action
(i.e. web browser failure). This patch resolves this issue.

Related: https://dev.ib.pl/ib/otrs/issues/64
Author-Change-Id: IB#1052439


mgruner:
Merged in 18266d9c15031618bbdb0cbb735ae495d2012959 for master, thanks a lot @pboguslawski!

PR #1226, opened 2016-06-04
Pass Return value to Postmaster post filter

reneeb:
That way it is far easier to check in a post filter if this runs on a new ticket or on a follow up. This can be quite hard to determine when you have ticket event modules that add notes on ArticleCreate or TicketCreate. The postmaster creates a new ticket the ticket event module runs and now you have two articles in the ticket. Now you have no chance to get to know if the postmaster post filter runs for a mail that created a new ticket


mgruner:
Applied in 77da564831bfb7ff1310b82ef73ab2b343812c7a. Thanks a lot @reneeb!

PR #1228, opened 2016-06-04
Currently it is impossible to save a 0

reneeb:
This might be ok in stock OTRS, but when you extend your table and you want to save a '0', you can't.
That's why I replaced the or-operator with the defined-or-operator.


carlosfrodriguez:
Hi @reneeb 

Thank you for your pull request it was merged in: https://github.com/OTRS/otrs/commit/b85025921c5e18b6c60a9cab9a330a02f1b4c49c and https://github.com/OTRS/otrs/commit/4837b61849ba0420d64d41f8af9f57e344defa6e.

((enjoy))
cr

PR #1229, opened 2016-06-05
Small UI changes

urbalazs:
Hi @mgruner 
This PR contains two small UI changes: it removes an unneeded dot from the end of a tooltip, and makes a button similar than the others.


mrcbnsls:
Thanks, merged in 8b73159. I found one more button which had the old style and changed it. I also introduced a class named 'Inline' which can be used for buttons which should be displayed e.g. next to an input field.

PR #1230, opened 2016-06-06
Changed ASCII arrows to Unicode

urbalazs:
Hi @mgruner 
This PR changes the ASCII arrows (&lt;-&gt;) to Unicode (↔, U+2194) in the admin area. I also changed them in the language files, so they will be still translated after commit.
Can you apply this patch also to rel-5_0, if it is possible? If it is not possible, I will make an other PR for rel-5_0, if needed.


mgruner:
Hello @urbalazs, thanks for this change. We'll only apply this to master though, and can you please remove all the language files from the commit, as the data is maintained in Transifex?

urbalazs:
Hi @mgruner 
I replaced the arrows in the language files, because these strings are directly displayed on the admin interface. If a translator don't translate them, it become untranslated, and displayed in English. This is a trivial change, we can also replace these arrows in the language files. The translators have to do nothing, it will be still translated.
I will make an other pull request for rel-5_0. In this case is very important, that the language files will be still translated.
So can I leave this changes in the language files, or only change them in the source code, and let translators to update the translations?

mgruner:
Hello @urbalazs. The language files are directly generated from Transifex, that's why they should not be modified. Changes will be lost on next sync. We won't apply these changes to the stable rel-5_0 branch, so you don't need to prepare a PR.

urbalazs:
Hi @mgruner 
I removed the changes from the language files.

mrcbnsls:
Thanks, merged in aef963c. For AdminTemplateAttachment.tt, I changed the page title to be consistent with other screens where relations are managed.

PR #1232, opened 2016-06-07
Master - JS refactoring - Modify AdminGenericInterfaceTransportHTTPSOAP

s7design:
Hello @mrcbnsls ,

Hereby is refactored AdminGenericInterfaceTransportHTTPSOAP module. Created new JS file that handles module events.

During testing found out bug in this module when selecting Sort options. You could create new levels for options and sub levels, but on submit form they were deleted and didn't record. Problem was in Core.Agent.SortedTree.js function that generates JS data, reading data from created options and inputing in hidden 'Sort' field. Modified trigger for that function and now it's performing as expected which is proved in new Selenium test.

Additionally took liberty to correct wrong label for RequestNameFreeText field.

Please let me know if you experience any issues or have any questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Config/Files/GenericInterface.xml
+++ b/Kernel/Config/Files/GenericInterface.xml
@@ -148,6 +148,7 @@
                 <Loader>
                     <CSS>Core.Agent.Admin.GenericInterface.css</CSS>
                     <CSS>Core.Agent.SortedTree.css</CSS>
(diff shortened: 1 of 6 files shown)


mrcbnsls:
Thanks, merged. I feel using 'click' directly on a form is not correct, so I changed it to be bound on the submit buttons, which is also working fine.

mrcbnsls:
Merged in 187c48b by the way.

PR #1233, opened 2016-06-07
Master - JS refactoring - Modify AdminQueue

s7design:
Hi @mrcbnsls 

Hereby is updated AdminQueue module. We moved JS code from TT file to the new JS file. Also, we added block Filter in TT file.

Regards
Zoran


zottto:
Merged with ea46e39.

PR #1239, opened 2016-06-08
Fixed performance issue in PermissionUserGet.

thorsteneckel:
Hi there,

currently the PermissionUserGet function of the Group object opens a performance leak for Agents without a role association. If the requested user is not associated to any Role (empty %RoleList) the cache will never get filled with the gathered data. Therefore on each PermissionUserGet call the whole function logic will get executed over and over again. This can lead to high response times in the AgentInterface, namely AgentTicketQueue where the function gets called a lot. We increased the response time by 0,5 sec in any of the SML views with this change.

I removed the return statement and added an if block around the loop.

Let me know if I can clarify any questions.

  Thorsten

PS: A backport for OTRS 5 would be great if possible.

### DIFF
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -1024,20 +1024,20 @@ sub PermissionUserGet {
         UserID => $Param{UserID},
     );
 
-    return %GroupList if !%RoleList;
+    if ( %RoleList ) {
+        ROLEID:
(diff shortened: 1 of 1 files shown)


mgruner:
Merged in a46867e3203f79af230f89bb99275b878b8e5a9c and backported. Thanks a lot @thorsteneckel.

PR #1199, opened 2016-05-30
Master - JS refactoring - Modify AdminCustomerUser

s7design:
Hi @mrcbnsls 

Hereby is updated AdminCustomerUser.tt.

Regards
Zoran

PR #1240, opened 2016-06-08
Master - Selenium Update - AdminCustomerUser

s7design:
Hello @mrcbnsls ,

Hereby is updated Selenium test for AdminCustomerUser as you suggested in PR https://github.com/OTRS/otrs/pull/1199 .

Please let me know if this is what you had in mind, I can do similar for AgentTicketEmail as well, but IMHO it's not necessary.

Regards,
Sanjin


mrcbnsls:
Excellent, thats exactly what I meant. Merged in c629214, thanks a lot! It should be enough do have this only for AgentTicketPhone, yes.

PR #1243, opened 2016-06-08
Master - JS refactoring - Modify AdminRole

s7design:
Hi @mrcbnsls 

Hereby is updated AdminRole module. We moved JS code from TT file to the new JS file and added block Filter in TT file.

Regards
Zoran


mrcbnsls:
Merged in c942bb5, thanks!

PR #1244, opened 2016-06-08
Refactoring of lock behavior of AgentTicketProcess.

thorsteneckel:
Hi there,

we noticed a strange behavior when opening an ActivityDialog from the AgentTicketZoom view of a ProcessTicket. Everytime the ActivityDialog pop up opens the parent window (AgentTicketZoom) reloads. (This caused an error that will be handled in another PullRequest).

After some research we found the lines that have been changed in this pull requests. The ticket lock and owner were always set even if the Agent was already the owner and the ticket locked.

Additionally the TicketNumber was only needed in the else case so we've put it right over the line.

  Thorsten

PS: A cherry-pick to OTRS 5 would be nice.

### DIFF
--- a/Kernel/Modules/AgentTicketProcess.pm
+++ b/Kernel/Modules/AgentTicketProcess.pm
@@ -142,11 +142,6 @@ sub Run {
             # check if it's already locked by somebody else
             if ( $ActivityDialogHashRef->{RequiredLock} ) {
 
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @thorsteneckel 

Thank you for your pull request, it was merged in https://github.com/OTRS/otrs/commit/82b34896cd933b423ac61491eb53c68c21fdd91b.

((enjoy))
cr

PR #1246, opened 2016-06-08
Fixed performance leak on big systems.

thorsteneckel:
Hi there,

this is a heavy one. Even though the result is cached its pretty expensive to fetch a UserList (on big systems) and filling a Hash with it for checking if a user is valid. The fetched UserList is used exclusively for checking the validity of the requested UserID. A (cached) GetUserData call is much cheaper.

A backport to OTRS 5 would be great.

  Thorsten

### DIFF
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -1383,12 +1383,11 @@ sub PermissionUserGroupGet {
 
     return if !%PermissionTypeList;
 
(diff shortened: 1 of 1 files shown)


mgruner:
Merged in 4e59ea94050b942bc14d6c0567bf1b40b1e64ba4 and backported. Thanks @thorsteneckel! 👍

PR #1247, opened 2016-06-08
Fixed bug in SOAP Username and/or Password handling

thorsteneckel:
Hi there,

the following commit should fix a possibly security issue:
https://github.com/OTRS/otrs/commit/86dd986fedf5c22f14a68fa86d4b35dd2bc6232f

I'm currently not aware of any issue that should be caused by using the plain string variables configured in the transport config. @carlosfrodriguez do you remember which issue that was? If there is one UnitTests for this would be great.

We had an issue where those line caused a wrong quoting of the char `!` in the password causing "Error in SOAP call: 401 Unauthorized" errors in the GenericInterface log. After removing it everything went as expected.

I'm glad to clarify any questions.

A cherry-pick to OTRS 5 would be great.

  Thorsten

### DIFF
--- a/Kernel/GenericInterface/Transport/HTTP/SOAP.pm
+++ b/Kernel/GenericInterface/Transport/HTTP/SOAP.pm
@@ -538,7 +538,7 @@ sub RequesterPerformRequest {
             my $User     = $Config->{Authentication}->{User};
             my $Password = $Config->{Authentication}->{Password};
             if ( IsStringWithData($User) && IsStringWithData($Password) ) {
-                $URL =~ s{ ( http s? :// ) }{$1\Q$User\E:\Q$Password\E@}xmsi;
+                $URL =~ s{ ( http s? :// ) }{$1$User:$Password@}xmsi;
             }
         }
     }


carlosfrodriguez:
Hi @thorsteneckel,

Sorry I don't remember at the moment.

((enjoy))
cr

thorsteneckel:
@carlosfrodriguez thanks for your lightning fast response 👍 I'm looking forward to see this in the framework.

mgruner:
This looks strange. Typically \Q and \E make sense in the match part, but not in the replace part?

carlosfrodriguez:
Hi @thorsteneckel,

Could it be possible for you to add a unit test?

((enjoy))
cr

thorsteneckel:
Hi @carlosfrodriguez - I'm afraid I can't do that due to the current order situation at our side. Sorry.

carlosfrodriguez:
Hi @thorsteneckel,

This issue was recently fixed for [bug#12855](https://bugs.otrs.org/show_bug.cgi?id=12855) using a different approach, please test if this works also for you.

I will close this pull request for now.

((enjoy))
cr

PR #1251, opened 2016-06-09
Master - Fixed mandatory fields

s7design:
Hi @mrcbnsls 

Hereby is updated some screens where there were duplicated blocks for a mandatory fields. I suppose now it is solved better. It the the same as I have done this in the PR https://github.com/OTRS/otrs/pull/1167

If you like this solution you can merge it.

Regards
Zoran


mrcbnsls:
Merged in 19ad14b, thanks, looks good for me!

PR #1252, opened 2016-06-09
Master - JS refactoring - Modify AdminPostMasterFilter

s7design:
Hi @mrcbnsls 

Hereby is updated AdminPostMasterFilter module. We moved JS code from TT to the new JS file, added Filter block in TT file and confirm window for delete action in JS file. Also, we included this confirmation in Selenium test.

Regards
Zoran


mrcbnsls:
Merged with some layout improvements, thanks b94f3d3 :)

PR #1253, opened 2016-06-09
Master - JS refactoring - Modify AdminPriority

s7design:
Hi @mrcbnsls 

Hereby is updated AdminPriority module. We moved JS code to the new JS file and added Filter block.

Regards
Zoran


mrcbnsls:
3391e02 👍

PR #1255, opened 2016-06-10
Master - Selenium Update - AgentTicketSearch

s7design:
Hello @jepf , @carlosfrodriguez and @mgruner ,

Hereby is updated Selenium test AgentTicketSearch regarding introduced changed with bug [#10649](http://bugs.otrs.org/show_bug.cgi?id=10649). Updated to do search with StaticDB with minimum, maximum and stop word search where we check for expected error.

Please let me know if you want any additional cases to see in Selenium test regarding this issue.

Regards,
Sanjin

### DIFF
--- a/scripts/test/Selenium/Agent/AgentTicketSearch.t
+++ b/scripts/test/Selenium/Agent/AgentTicketSearch.t
@@ -19,7 +19,22 @@ $Selenium->RunTest(
     sub {
 
         # get helper object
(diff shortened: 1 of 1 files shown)


mgruner:
Merged with small changes. I only recalculate the article index for this one article now, added the required mod_perl sleep for fast systems, and also added a new WaitFor method that waits until a dialog is open (delayed opening via AJAX, as in this case). Let me know if you have any problems with this.

PR #1257, opened 2016-06-13
Master - JS refactoring - Modify AgentBook

s7design:
Hi @mrcbnsls,

There is updated AgentBook screen for the Address option. As is has been done before, there is added new JS for the script from the frontend module. There is also new Selenium test with the test cases which test Address option feature.

Regards
Zoran


mrcbnsls:
Merged in e98f60e, thanks. I took the opportunity to improve the layout a bit. :)

PR #1258, opened 2016-06-13
Master - JS refactoring - Modify AgentCustomerInformationCenter

s7design:
Hi @mrcbnsls 

Hereby is updated AgentCustomerInformationCenter module. We moved click event to the new JS file and Init call replaced with RegisterNamespace call in Core.Agent.Dashboard.js.
Also, we added case in Selenium test which tests JS modal dialog opening.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -8819,6 +8819,7 @@ via the Preferences button after logging in.
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
                     <JavaScript>Core.Agent.Dashboard.js</JavaScript>
                     <JavaScript>Core.Agent.TableFilters.js</JavaScript>
+                    <JavaScript>Core.Agent.CustomerInformationCenter.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
--- a/Kernel/Output/HTML/Templates/Standard/AgentCustomerInformationCenter.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentCustomerInformationCenter.tt
@@ -12,15 +12,5 @@
 
 [% InsertTemplate("AgentDashboardCommon.tt") %]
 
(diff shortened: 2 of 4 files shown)


zottto:
Hi Zoran,

the new Selenium tests are failign for me:

 not ok 45 - Title 'Search' exists (is '0' should be '1')
.
 not ok 47 - Field for search by customer user exists (is '0' should be '1')

 not ok 48 - Exception in Selenium': An element could not be located on the page using the given search parameters: #AgentCustomerInformationCenterSearchCustomerID,css selector at (eval 122) line 165.

Can you have a look again, please?

Thanks,
Marc

s7design:
Hi @ zotto

Thanks for follow-up.
I also have some errors today. It is so strange because it worked yesterday. We are trying to solve this.

Regards
Zoran

s7design:
Hi @zottto 

We are tried to solve the error in the test. There is some strange and tricky situation. It works sometimes on some our system, but it is not complete stable. There is problem because  the modal dialog is opened and all fields are visible but autocomplete items is not visible/clickable in Selenium, however manually it works.  I decided to change this test a little bit, and now all test cases work fine.

If you like we can merge it in this moment. If we find some other test cases which we can test as well we will be able to extend the test. Please let me know what do you think about that.

Regards
Zoran

zottto:
I get the following Selenium UT errors:

 not ok 43 - There is link to customer information center search modal dialog. (is 'Kernel::System::UnitTest::Selenium::WebElement=HASH(0xc83410c)' should be '1')
................
 not ok 60 - Exception in Selenium': An element could not be located on the page using the given search parameters: //a[contains(@href, 'Subaction=Search;StateType=Open;CustomerID=test14671073050' )],xpath at (eval 122) line 156.

s7design:
Hi @zottto 

I fixed it I think it was my mistake.

Regards
Zoran

zottto:
Merged with 2dd27a7.

Thanks,
Marc

PR #1259, opened 2016-06-13
Master - JS refactoring - Modify AgentDashboard

s7design:
Hi @mrcbnsls

Hereby is updated AgentDashboard module. We replaced Init call from TT file with RegisterNamespace call in Core.Agent.Dashboard.js.

Regards
Zoran


zottto:
Merged with d5fa982.

PR #1261, opened 2016-06-13
Master - JS refactoring - Modify AgentCustomerInformationCenterBlank

s7design:
Hi @mrcbnsls

Hereby is updated AgentCustomerInformationCenterBlank module. TT file contains just one JS function call which is moved to PM module.

Regards
Zoran


zottto:
Merged with 4c8489c.

PR #1268, opened 2016-06-15
Master - JS refactoring - AgentCustomerInformationCenterSearch

s7design:
Hi @zottto 

Herby is updated AgentCustomerInformationCenterSearch.tt. There was only Core.Agent.CustomerInformationCenterSearch call in the screen TT module. There is specific case because it is a dialog and RegisterNamespace and ExecuteInit do not work as we have done in other modules ( with the hook APP_MODULE). We also tried add new JS and add it in syssconfig module ragidtration for AgentCustomerInformationCenterSearch, if would work it there was Core.Init.ExecuteInit('APP_MODULE').  So Init() function is called in CustomerInformationCenterSearch.js.  I think you like this solution, if you have any other options which will work please let me know.

There is also new Selenium test where we test all possible test cases how CIC search could be called.

Please let me know if we can improve that.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentCustomerInformationCenterSearch.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentCustomerInformationCenterSearch.tt
@@ -24,6 +24,3 @@
         </fieldset>
     </form>
 </div>
-<script type="text/javascript">//<![CDATA[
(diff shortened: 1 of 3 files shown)


zottto:
Merged with 61cc31e.

PR #1270, opened 2016-06-15
Master - JS refactoring - ActivityDialogFooter

s7design:
Hi @zottto 

Herby is my proposal for refactoring ActivityDialogFooter, acctually there is not moved JS script because it is Footer TT and I left the scripts as it is with FooterJS.tt.

I only saw that there is duplicated code for FieldHelpContainer, dynamic fields help tool tip description.  There is the same snippet in the Core.Agent.TicketProcess JS:
https://github.com/OTRS/otrs/blob/master/var/httpd/htdocs/js/Core.Agent.TicketProcess.js#L122

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/ProcessManagement.xml
+++ b/Kernel/Config/Files/ProcessManagement.xml
@@ -62,6 +62,7 @@
                     <CSS>Core.AllocationList.css</CSS>
                     <JavaScript>Core.Agent.Admin.ProcessManagement.js</JavaScript>
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
+                    <JavaScript>Core.ProcessManagement.ActivityDialog.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
@@ -154,6 +155,7 @@
                     <JavaScript>Core.Agent.CustomerSearch.js</JavaScript>
                     <JavaScript>Core.Agent.TicketAction.js</JavaScript>
                     <JavaScript>Core.Agent.TicketProcess.js</JavaScript>
+                    <JavaScript>Core.ProcessManagement.ActivityDialog.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
@@ -289,6 +291,7 @@
                 </NavBar>
                 <Loader>
                     <CSS>Core.Customer.TicketProcess.css</CSS>
+                    <JavaScript>Core.ProcessManagement.ActivityDialog.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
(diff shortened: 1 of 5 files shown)


zottto:
Hi Zoran / @s7design ,

the ActivityDialogFooter.tt is loaded both via Ajax or with a normal HTTP request. If loaded via AJAX, the FooterJS block is rendered, otherwise the normal footer is loaded (including the normal FooterJS.tt).
That means, in case of a normal HTTP request the normal init hooks are working. Only in case of an ajax request an init hook is missing here.
That's why I propose the following solution:
Please add the init hooks to the FooterJS block of ActivityDialogFooterJS.tt (in a similar way and order as in FooterJS.tt).
This way, you can remove the other JS in this file and move it to external files as well.

Please update this PR accordingly.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for follow-up. I modified ActivityDialogFooter.tt as you suggested, or I think you meant that I should do something like that. There is also updated CustomerActivityDialogFooter.tt.

I tested it and I hope it works as it did before. In my opinion it is better if we solve all this in one PR. Please let me know if you have any other suggestion.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

thanks for the update.
You should update the FooterJs block to not only init the APP_MODULE hook, but also all other hooks available in FooterJS.tt, to make sure, that also (possible future) JS code is correctly executed in this case.

In your JS code the following line:
```javascript
if(Core.Config.Get('ParentReload') && Core.Config.Get('ParentReload') == 1){
```
should better be
```javascript
if (typeof Core.Config.Get('ParentReload') !== 'undefined' && parseInt(Core.Config.Get('ParentReload'), 10) === 1) {
```

Please update this PR like this.

Thanks,
Marc

s7design:
Hi @zottto 

I updated PR, thanks for your suggestions. I add all other hooks. Please let me know if you thin that I should do that or something else 

Regards
Zoran

zottto:
Merged with 913d4c2.

I had to change this PR because of some problems with the ajax loading and JS initializing:

- removed the new JS file completely because most of the code was already in Core.Agent.TicketProcess.js and in CustomerTicketProcess.tt (equivalent JS file is missing here)
- added missing JS code to Core.Agent.TicketProcess.js and updated it to work also in responsive environments
- removed unnecessary call to ParentReload block from CustomerTicketprocess.pm

Marc

s7design:
Hi @zottto 

I totally agree with you, it is much better now.
To be honest I was not sure if Core.UI.TreeSelection.InitTreeSelection(); or Core.Form.Validate.Init(); are necessary there.
Thanks.

Regards
Zoran

PR #1271, opened 2016-06-16
Master - JS refactoring - AdminProcessManagement

s7design:
Hello @mrcbnsls ,

Hereby is refactored AdminProcessManagement module together with Activity/ActivityDialog/Transition/TransitionAction/ modules. Closed previous relevant PR's regarding these changes and felt the need to push these together for easier overview and merging. If you think they should go separate, please do not hesitate to request it.

Additionally add "No matches found" for invalid filter input, it is just more consistent with remaining of system behaviour. Modified Selenium tests to check for additional JS behaviour for Transition and Transition action modules.

Please if you have any comments or issues regarding these changes, let me know, i will be glad to improve it.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminProcessManagement.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminProcessManagement.tt
@@ -171,16 +171,13 @@ SET OTRSBusinessLinkLabel = '<strong><a href="#" class="OTRSBusinessRequired">OT
                             </td>
                         </tr>
 [% RenderBlockEnd("ProcessNoDataRow") %]
+                        <tr class="FilterMessage Hidden">
+                            <td colspan="4">[% Translate("No matches found.") | html %]</td>
+                        </tr>
                     </tbody>
                 </table>
             </div>
-
         </div>
     </div>
     <div class="Clear"></div>
 </div>
-[% WRAPPER JSOnDocumentComplete %]
-<script type="text/javascript">//<![CDATA[
-Core.Agent.Admin.ProcessManagement.InitProcessOverview();
-//]]></script>
-[% END %]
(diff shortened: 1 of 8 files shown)


zottto:
Hi Sanjin / @s7design ,

you removed the InitProcessOverview function from Core.Agent.Admin.ProcessManagement.js. From my point of view, now the InitProcessPopups() method is not called anymore for this case (only for the other Subactions which get their init call in the new Init() function). Isn't that missing now?

Can you please check and explain?

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for your feedback, reverted deleted InitProcessOverview function. Overlooked possibility of opening popup window besides once that are in Init function. 

Please let me know if there are any further issues regarding this PR.

Regards,
Sanjin

zottto:
HI Sanjin / @s7design,

now the InitProcessOverview() function is back in the code, but it is still not executed. You removed the JS call from the template but you did not integrate it into the JS file.
Please check that again.

Thanks,
Marc

s7design:
Hello @zottto ,

Reverted changes that I introduced initially in this PR where InitProcessOverview function is removed. To be honest there is really no need for this function anymore. Process filter is initialised in Init, and popups are executed in there respective functions. I tested manually all cases that came to my mind and with these changes everything works as far as my testing goes.

If we leave InitProcessOverview function and initialise it in Init, on opening Activity add for example, double popup is opened. Tried also removing InitProcessPopups from other functions to avoid this problem, but then there is a case where on first click popup is opened as expected yet on second click on same button new window is loaded.

Please, correct me if my findings are wrong and if you have any comments regarding this issue. Any feedback from your side is welcomed and i will revisit this PR again.

Regards,
Sanjin

zottto:
Thanks for clarification, Sanjin.
Merged with 7d7e79d.

Thanks,
Marc

PR #1273, opened 2016-06-16
Master - JS refactoring - AdminSystemAddress

s7design:
Hello @zottto ,

Hereby is refactored AdminSystemAddress module. Moved JS from TT to a new JS file. Included changes so filter only shows in overview screen. There was redundant JS function for MasterAction class click which is handled via Core.UI.js.

Regards,
Sanjin


zottto:
Merged with 9a86f37.

PR #1274, opened 2016-06-16
Fixed plain e-mails saving with unfolded headers

pboguslawski:
OTRS stores plain copy of every e-mail in plain.txt file with headers
unfolded (i.e. Received header is saved as one line even if in original
message it was multiline). This copy should keep original message
formatting, without unfolding (i.e. for debugging purposes and not
to mess up linebreaks if message arrive with DOS line endings).

Related: https://dev.ib.pl/ib/otrs/issues/70
Author-Change-Id: IB#1022256

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-06-16 Fixed plain e-mails saving with unfolded headers.
  - 2016-06-06 Added checks for length of words in search terms when using search index/StaticDB backend.
  - 2016-06-06 Fixed bug#[12020](http://bugs.otrs.org/show_bug.cgi?id=12020) - Option CaseSensitive has no impact on external customer database.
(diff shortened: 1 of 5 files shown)


mgruner:
Merged with a slight change: use the existing function GetPlainEmail() instead of adding a new one in 7fce3ac2a315d9baac2d2d94de471f9946c29243. Please let me know if this is ok for you. No backport at this time.

pboguslawski:
> Please let me know if this is ok for you.

Seems to be ok and simpler.  Passing raw e-mail without unfolding should probably not hurt badly other places that use GetPlainEmail():

    Kernel/System/PostMaster/FollowUpCheck/RawEmail.pm
    Kernel/System/PostMaster/Filter/CMD.pm

Even in case of problems, GetPlainEmail() should return original message without changes probably so above places will have to be adapted to this assumption.

Regards,
Pawel

PR #1276, opened 2016-06-16
Fixed errors when processing incomplete session data

pboguslawski:
OTRS does not check if session data is complete when counting sessions
and may throw errors like

    Use of uninitialized value $Data{"UserType"} in concatenation (.) or string at /opt/otrs//Kernel/Modules/AdminSession.pm line 170.
    Use of uninitialized value $Data{"UserLogin"} in string at /opt/otrs//Kernel/Modules/AdminSession.pm line 171.

Incomplete sessions may occur afters race conditions; i.e. open two
OTRS tabs (same session) and run longer query in one tab (i.e. report
or some other longer stuff like entering AdminSysConfig) and logout
immediately on second tab; after querty finish in the first tab,
session table will contain incomplete session, containing only two
data rows, i.e.:

    SessionID AgG2BEUhx1ZqsbZxobcD4pKIz5LkzKE
    UserLastRequest 1466080902

This patch fixes this issue.

Related: https://dev.ib.pl/ib/otrs/issues/71
Author-Change-Id: IB#1016015

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-06-16 Fixed errors when processing incomplete session data.
  - 2016-06-06 Added checks for length of words in search terms when using search index/StaticDB backend.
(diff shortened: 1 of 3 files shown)


mgruner:
Merged in 51ea89c3f47999d2f011e716b7c1004e4514183d, thanks @pboguslawski!

PR #1277, opened 2016-06-16
Master - JS refactoring - AgentCustomerSearch

s7design:
Hi @zottto 

Hereby is updated AgentCustomerSearch module. Due to many TT modules use JS from AgentCustomerSearch.tt, we changed almost all of these modules. We didn't finish refactoring ProcessManagement which also uses JS code from AgentCustomerSearch.tt because it has different behavior and because of this we didn't delete AgentCustomerSearch.tt yet.
Please if you can provide any feedback regarding this PR, it will be more than welcomed.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -8946,6 +8946,7 @@ via the Preferences button after logging in.
                 <NavBarName></NavBarName>
                 <Loader>
                     <JavaScript>Core.Agent.CustomerSearch.js</JavaScript>
+                    <JavaScript>Core.Agent.CustomerSearchAutoComplete.js</JavaScript>
                     <JavaScript>Core.Agent.TicketAction.js</JavaScript>
                     <JavaScript>Core.Agent.Book.js</JavaScript>
                 </Loader>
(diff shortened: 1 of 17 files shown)


zottto:
Hi Zoran / @s7design ,

the JS function Core.Config.Get() is not a substitute for a [% Config %] call in the template, it just can store internal data that is stored via a Core.Config.Set() call.
Because of that your code is not working in Core.Agent.CustomerSearchAutoComplete.js.

To remove the template file and still get the needed sysconfig settings in JS you have to send them there from Perl with the new $Self->{LayoutObject}->AddJSData() call. Than, these settings can also be used in JS.

Perhaps you find a way to add the two sysconfig settings globally, so that you do not have to specify them again and again in every module that wants to use the autocompletion.

Thanks,
Marc

s7design:
Hi @zottto 

Thank you for your feedback. We added sending of config settings in PM modules. Please let me know if you have any other suggestions.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

this looks quite good, but there is more room for improvement. ;-)
If you name the keys of the JSData on perl side correctly, than you don't have to move the valued from one config setting to the other in the Init function.
So, instead of ShowCustomerTickets name it CustomerSearch.ShowCustomerTickets directly on perl side. 

Thanks,
Marc

s7design:
Hi @zottto 

Thank you for your follow up. PR is updated regarding your suggestions.

Regards
Zoran

zottto:
Merged with db958f8.

After all templates are JavaScript-free, we will do a second round of refactoring to eliminate these almost useless JS files with only one function call in it.

Thanks,
Marc

PR #1278, opened 2016-06-17
Performance: CustomerGroup GroupMemberList

thorsteneckel:
Hi there,

we found out that the CustomerGroup Object function 'GroupMemberList' has a little performance hole. Even if a cache is set for a call there is always a GroupList call performed if CustomerGroupSupport is disabled and not used afterwards. So we moved the cache check before the GroupList call.
Additionally we noticed that the DB query is always performed even if the CustomerGroupSupport is disabled. This is definitive a performance loss and might even be a bug.

Thorsten

PS: A backport to OTRS 5 would be great.

### DIFF
--- a/Kernel/System/CustomerGroup.pm
+++ b/Kernel/System/CustomerGroup.pm
@@ -186,20 +186,6 @@ sub GroupMemberList {
         );
         return;
     }
-    my %Data;
-    my @Name;
-    my @ID;
-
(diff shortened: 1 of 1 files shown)


mgruner:
Applied and backported in e7206cb1e414e0574539b40688b4c87b1c92548f. I had to manually backport, can you please check if this was ok? Thanks a lot @thorsteneckel!

PR #1279, opened 2016-06-17
Fixed uninitialized counters after TicketSearch calls without permissions

pboguslawski:
When agent without permissions (no groups associated) opens
AgentTicketStatusView, OTRS throws to apache error log many messages
similar to

    Use of uninitialized value $ViewableTicketCount in numeric gt (>) at [...]Kernel/Modules/AgentTicketStatusView.pm line 286.

This mod fixes such issues by setting return value to 0 if TicketSearch()
called with Result => 'COUNT' returns undef (i.e. agent has no permissions
to any queues in system); this is similar solution to one found in
AdminGenericAgent.pm.

Related: https://dev.ib.pl/ib/otrs/issues/72
Author-Change-Id: IB#1019544

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-06-17 Fixed uninitialized counters after TicketSearch calls without permissions.
  - 2016-06-06 Added checks for length of words in search terms when using search index/StaticDB backend.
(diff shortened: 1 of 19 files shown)


mgruner:
Merged in 507446ff87ac2ae6530ed86cca50b94c90eecf6c, thanks a lot @pboguslawski!

PR #1281, opened 2016-06-17
Master - JS refactoring - AdminType

s7design:
Hi @zottto,

Hereby is updated AdminType.tt view.

Regards
Zoran


zottto:
Merged with 30cdb06.

PR #1282, opened 2016-06-20
Master - JS refactoring - AdminLinkObject

s7design:
Hi @zottto 

Hereby is updated AgentLinkObject view. We added new JS file Core.Agent.TicketLinkObject.js because Core.Agent.LinkObject.js still exists and it is much simpler to separate these two things.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -8962,7 +8962,7 @@ via the Preferences button after logging in.
                 <Title Translatable="1">Link Object</Title>
                 <NavBarName></NavBarName>
                 <Loader>
-                    <JavaScript>Core.Agent.TicketLinkObject.js</JavaScript>
+                    <JavaScript>Core.Agent.LinkObjectSearchForm.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -8962,7 +8962,7 @@ via the Preferences button after logging in.
                 <Title Translatable="1">Link Object</Title>
                 <NavBarName></NavBarName>
                 <Loader>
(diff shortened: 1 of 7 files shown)


zottto:
Hi Zoran / @s7design ,

in my opinion, Core.Agent.TicketLinkObject.js is the worng name, because the code is also used to link other type of objects.

And at the moment I don't see a reason why the JS schould not be added to the existing Core.Agent.LinkObject.js. Can you explain, why you want to separate this?
It is OK for me, if there are reasons for that, but than the file should be renamed.

Thanks,
Marc

s7design:
Hi @zottto 

Initially we want to separate files because we planned moving JS code from LinkObject.tt to Core.Agent.LinkObject.js in the future.
Second reason is that console error message appears when Core.Agent.LinkObject.js is loaded for AgentLinkObject.tt ('TypeError: Core.Agent.TableFilters is undefined'), in the case when all code is in Core.Agent.LinkObject.js due to call Core.Agent.TableFilters.SetAllocationList().
We extend Selenium test for opening popup when we search with empty form.
Please let me know if you have any comments regarding these changes, I will be grateful.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

can you please rename the file and namespace to be a subset of the LinkObject namespace like Core.Agent.LinkObject.SearchForm?

And please keep in mind, that when using JavaScript and reading data with Core.Config.Get() you have to check the value differently:
```javascript
if (TemporaryLink && TemporaryLink == 1) {
```
must be:
```javascript
if (typeof TemporaryLink === 'undefined' && parseInt(TemporaryLink, 10) === 1) {
```
1. check if a value is defined with typeof
2. To compare a value against an integer always parse the value before with parseInt
3. Always compare with '==='

Thanks,
Marc

s7design:
Hi @zottto

Thank you for your follow up. PR is updated.

Regards
Zoran

zottto:
Merged with bcb4dd624a31be349df5246a2193678f44c7952f.

Thanks,
Marc

PR #1283, opened 2016-06-20
Fixed risky status passing and redundant TicketGet calls

pboguslawski:
TicketStateSet() calls HistoryAdd() without specyfing new StateID
as argument which may cause problems in case of cache cleanup
issues. Such arguments should be passed using arguments not
cache to be safer.

HistoryAdd() calls TicketGet() many times which is not necessary.

This mod fixes both issues.

Related: https://dev.ib.pl/ib/otrs/issues/73
Author-Change-Id: IB#1018102


dvuckovic:
Merged in fc37f7d, thanks to @pboguslawski!

PR #1286, opened 2016-06-21
Fixed unnecessary merge errors when link already exists

pboguslawski:
OTRS throws error messages like

    Link already exists between these two objects in opposite direction
    Another Link already exists within the same type group!

when merging tickets with existing links between it.

This mod changes such message priority from error to notice.

Related: https://dev.ib.pl/ib/otrs/issues/75
Author-Change-Id: IB#1016061


mgruner:
Hello @pboguslawski. Thanks a lot for your contribution. I don't like the added complexity in the API, and currently I also fail to see the use case. On an API level this should probably fail and error. Such errors would better be prevented from the UI, so that the agent cannot create such invalid links. 

@UdoBretz, do you agree?

I'll close this PR for now. Please feel free to reopen with more information!

PR #1287, opened 2016-06-21
Fixed error removing shm for log on OTRS startup

pboguslawski:
Sometimes during OTRS system startup shm errors occur, i.e.

    Can't remove shm for log: Invalid argument

This may be caused by concurrent OTRS cron jobs execution on
OS startup and races with calling CleanUp() in Log.pm.

This mod changes priority of such messages from error to notice.

Related: https://dev.ib.pl/ib/otrs/issues/76
Author-Change-Id: IB#1023103


pboguslawski:
See also http://bugs.otrs.org/show_bug.cgi?id=9000

mgruner:
Merged, thanks a lot!

PR #1288, opened 2016-06-21
Guard against TicketSearch for non-existing dynamic fields

moritz:
Previously, such search filter were ignored. This means that
searching for a dynamic field that has been deleted from
or deactivated in the database might return way too many tickets,
putting undue pressure on downstream system.

Includes a test.

### DIFF
--- a/Kernel/System/Ticket/TicketSearch.pm
+++ b/Kernel/System/Ticket/TicketSearch.pm
@@ -1383,16 +1383,18 @@ sub TicketSearch {
     }
 
     # catch searches for non-existing dynamic fields
+    PARAMS:
     for my $Key ( sort keys %Param ) {
-        if ( $Key =~ /^DynamicField_(.*)$/ && $Param{$Key} ) {
-            my $DynamicFieldName = $1;
-            $Kernel::OM->Get('Kernel::System::Log')->Log(
-                Priority => 'Error',
-                Message  => qq[No such dynamic field "$DynamicFieldName" (or it is inactive)],
-            );
+        next PARAMS if !$Param{$Key};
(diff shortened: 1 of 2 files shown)


Kernel/System/Ticket/TicketSearch.pm
```diff
@@ -1382,6 +1382,19 @@ sub TicketSearch {
         }
     }
 
+    # catch searches for non-existing dynamic fields
+    for my $Key ( sort %Param ) {
```
thorsteneckel: Shouldn't that be a `sort keys`?

Kernel/System/Ticket/TicketSearch.pm
```diff
@@ -1382,6 +1382,19 @@ sub TicketSearch {
...
+        if ( $Key =~ /^DynamicField_(.*)$/ && $Param{$Key} ) {
```
thorsteneckel: Here would be a GuardClause sweet:

```perl
next *** if !$Param{$Key};
next *** if $Key !~ m{\ADynamicField_}xms;
```

moritz: @thorsteneckel  changed in 45df725c5c3292031d1d89e2d5729a9cd4f1d7a9

thorsteneckel: NICE!

mgruner:
Merged in c1a5159a45a08a12b58b4ec7ba3ecbbd6486a01f, very nice, thanks @moritz!

mgruner:
And also to @thorsteneckel for QA 😀

thorsteneckel:
@mgruner you're welcome! love to read @moritz code

PR #1290, opened 2016-06-21
[ Rel-5_0 Bug 12106 ] - Use of uninitialized value $FirstElementIndex in splice at /var/www/html/otrsnext/Kernel/Output/HTML/Layout

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12106

Hello @carlosfrodriguez ,

Hereby is proposal for fixing reporters issue. To be honest, this bug was not reproducible in our different test environments. Value $FirstElementIndex carry value of 0, and perhaps this can cause some different perl version to interpret it as undef value resulting in reporters log warning.

Please let me know if you disagree with this change.

Regards,
Sanjin


dvuckovic:
Merged in 4c42e2c9e940d9c1b5da6e907d138822325db715, 4e22d6b3c44c34a383965423216c172cf8d1fcc2 and 45b2c6b121bc290c3ae7b186a7ecaa6133ace217. Thanks @s7design!

PR #1291, opened 2016-06-22
Master - JS refactoring - AgentTicketBounce

s7design:
Hi @zottto 

Hereby is updated AgentTicketBounce module. We moved JS code from TT to the new JS file and extended Selenium test for JS functionality.

Rregards
Zoran


mrcbnsls:
Merged in 1bd69e0, thanks!

PR #1293, opened 2016-06-23
Marked Titles as translatable

urbalazs:
Hi @mgruner 
This PR makes it possible to translate the title of CustomerTicketOverview screen.
Please also apply this patch to rel-5_0, as that is also affected.


mgruner:
Applied in 1db0d46a222dcbd746b4ac74be0242e6217f46c0 and backported, thanks a lot!

PR #1294, opened 2016-06-23
Master - JS refactoring - AgentNavigationBar

s7design:
Hello @zottto ,

Hereby is refactored JS for AgentNavigationBar. Moved JS from TT file into PM and removed block creation. Selenium test for this would be a bit tricky since we can't use drag and drop functionality, but manual testing can be done easily to verify refactoring.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -2913,11 +2913,8 @@ sub NavigationBar {
     );
 
     my $NavbarOrderItems = $UserPreferences{'UserNavBarItemsOrder'} || '';
-    $Self->Block(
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Merged in a6946f6bbdb6ffb81db7cda29c593c5a3916e5de.

As discussed in other PRs, adding JS code with AddJSOnDocumentComplete should be avoided if possible. As we're using a function defined in Core.Agent.js, it seems reasonable to add the call also there. Please see my changes on the commit mentioned above.

As you can see in 144fad99fb37ba39e7cfc9082d449a8a83f90b81, I added a simple test for the drag & drop feature.

PR #1295, opened 2016-06-23
Master - JS refactoring - AgentTicketMerge

s7design:
Hi @zottto 

Hereby is updated AgentTicketMerge module. We replaced Core.Agent.TicketMerge.Init() call from TT file with hook 'APP_MODULE' in Core.Agent.TicketMerge.js. We also extended Selenium test for JS functionality.

Regards
Zoran


mrcbnsls:
Merged in 1331616, thanks!

PR #1297, opened 2016-06-23
Master - JS refactoring - AgentSpelling

s7design:
Hello @zottto ,

Hereby is refactored JS AgentSpelling. Moved JS from TT to Core.Agent.TicketAction.js. Created Selenium test for this module as well.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentSpelling.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentSpelling.tt
@@ -63,9 +63,4 @@
         </form>
     </div>
 </div>
-[% WRAPPER JSOnDocumentComplete %]
-<script type="text/javascript">//<![CDATA[
-    Core.Agent.TicketAction.InitSpellCheck();
-//]]></script>
-[% END %]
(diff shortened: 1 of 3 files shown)


mrcbnsls:
Merged in b8d88fe.

I removed the new UT for now, although it's a good idea to add one in general. I'd ask you to add it again in a separate PR (low priority!) but with these enhancements:

- Make sure that the spellchecker binary is actually present. There could be different locations (e.g. /usr/bin/aspell, /usr/bin/ispell, /opt/local/bin/aspell). You could try to !-e the path to see if its there.
- Make sure the english dictionary is present. For this, you'll probably need to make a call on the binary with the dictionary and see if its output is as expected.
- If the checks mentioned above fail, the UT should not execute any further.

But, as mentioned, this doesn't have a high priority. So please focus on the JS refactoring :-)

Thanks!

s7design:
Hello @mrcbnsls ,

Agree with you regarding comments for Selenium test. While working on it it, saw it could raise some problems depending on configuration. Will keep in mind your addition and create PR with it when times allow it.

Regards,
Sanjin

PR #1298, opened 2016-06-23

urbalazs:
Hi @mgruner 
Since OTRS 4 doesn't support Transifex, this is the only way to update the translation. I send this, because I would like to finish the Hungarian translation for the upcoming 4.0.18 (if it will be released in the future - I hope).


mgruner:
Merged in c69f222f749da65e093091fd8b4f929382a531ed, thanks a lot!

PR #1300, opened 2016-06-24
Master - JS refactoring - AgentTicketForward

s7design:
Hi @zottto 

Hereby is updated AgentTicketForward module. We moved JS code from TT to the new JS file.
Part of code from AgentTicketForward.tt:

```
<script type="text/javascript">//<![CDATA[
    $('#AttachmentDeleteButton[% Data.FileID | html %]').bind('click', function () {
        var $Form = $('#AttachmentDeleteButton[% Data.FileID | html %]').closest('form');
        $('#AttachmentDelete[% Data.FileID | html %]').val(1);
        Core.Form.Validate.DisableValidation($Form);
        $Form.trigger('submit');
    });
//]]></script>

```
exists too in e.g. AgentTicketActionCommon.tt, AgentTicketCompose.tt etc. When we refactor all of these TT files, we would create one function and call from all appropriate JS files.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -8909,6 +8909,7 @@
                 <Loader>
                     <JavaScript>Core.Agent.CustomerSearch.js</JavaScript>
                     <JavaScript>Core.Agent.TicketAction.js</JavaScript>
+                    <JavaScript>Core.Agent.TicketForward.js</JavaScript>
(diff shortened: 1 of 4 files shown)


zottto:
Merged with e1c9642.

After moving all JS from TT to JS files we'll have a second round of refactoring to streamline the JS files and merge functions as you suggested.

Thanks,
Marc

PR #1301, opened 2016-06-26
Removed double colon from label

urbalazs:
Hi @mgruner 
I found an unnecessary colon in a label. Branch rel-5_0 is also affected.


mgruner:
Merged in c696a20b969117130023434885879932ed86012f and backported, thanks @urbalazs.

PR #1304, opened 2016-06-27
Master - JS refactoring - AgentTicketEmailOutbound

s7design:
Hi @zottto 

Hereby is updated AgentTicketEmailOutbound view. We moved JS code from TT to the new JS file.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -8928,6 +8928,7 @@
                 <Loader>
                     <JavaScript>Core.Agent.CustomerSearch.js</JavaScript>
                     <JavaScript>Core.Agent.TicketAction.js</JavaScript>
+                    <JavaScript>Core.Agent.TicketEmailOutbound.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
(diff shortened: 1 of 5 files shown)


zottto:
Hi Zoran / @s7design ,

please check again the usage of "$DynamicFieldNames" vs. "DynamicFieldNameStrg".
Does DynamicFieldNames really deliver an Array in JavaScript without any further converting?

If so, than DynamicFieldNameStrg in the perl file is not needed anymore and can be removed.

Is it possible to add tests for this use case?

Thanks,
Marc

s7design:
Hi @zottto 

Thank you for your feedback. DynamicFieldNames always delivers an Array in JS and DynamicFieldNameStrg is removed because it is not needed anymore.
We extended Selenium test to check for a JS refactored code.  First we wanted to do check with an ACL but that raised some problems so we decided to check for appearance of AJAX loader field before and after change state. 
If you have any suggestions how to improve test further on or additional comments let us know.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

the Selenium test is now failing for me:
 not ok 45 - AJAX Loader for 'DFtest14677055251' exists - JS function was run (is '0' should be '1')

Can you please check that again?

Thanks,
Marc

s7design:
Hi @zottto 

PR is updated regarding your suggestion.

Regards
Zoran

zottto:
Merged with 37fdddd.

Thanks,
Marc

PR #1305, opened 2016-06-27
Master - JS refactoring - Error, Warning and NoPermission

s7design:
Hello @zoto,

Hereby is refactored JS for Error, Warning and NoPermission screen. Included these three in same PR since they all share same JS which bind function on button 'Go to previous screen'. Created new JS file Core.Warn.js , please take this name with reservation and feel free to change it into anything you think suits better to reflect all three in one string.

If you have any issues or questions regarding this PR, do not hesitate to ask.

Regards,
Sanjin


zottto:
Hi Sanjin / @s7design,

the UTs are failing because the newly created file Core.Warn.js was not added to the Defaults.pm.

In this case I would suggest to not add the file there, but to remove it completly again and to move the function into another already existing JS file, e.g. Core.App.js. You can create a new function their and call that function in the Core.App.Init() (that does not exist yet).

To make this more robust, it is probably a good idea to rename the id of the link from "GoBack" to something that explains more, that this link is part of an error or warning.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for follow up. Used Core.Exceptions.js to insert new PreviousPage function which handles return to previous page. Included changes to field ID to represent in a better way button itself as you suggested. Corrected some comments in this JS file since they were mistakes from copy paste I assume. Included small change in AgentTicketMove Selenium test that is entering NoPermission page.

Please let me know if you agree with these changes.

Regards,
Sanjin

zottto:
Hi Sanjin / @s7design,

the function "PreviousPage()" makes no sense within Core.Exception.js because the content of the function is in no way related to an Exception (it is called on warning and error pages but the functionality itself is not related).

Please move it to Core.App and also rename it to a proper function name like "ReturnToPreviousPage()". The link should also have a more clear identifier instead of "PreviousPage". And the identifier should better be a class instead of an id to allow multiple links of this type in one page.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for your feedback. This PR is updated accordingly.

Regards,
Sanjin

zottto:
Merged with 01c76d2.

Thanks,
Marc

PR #1306, opened 2016-06-27
Master - JS refactoring - AgentTicketEmail

s7design:
Hi @zottto 

Hereby is updated AgentTicketEmail module. We moved all JS code from TT to the new JS file.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -8461,6 +8461,7 @@
                 <Loader>
                     <JavaScript>Core.Agent.CustomerSearch.js</JavaScript>
                     <JavaScript>Core.Agent.TicketAction.js</JavaScript>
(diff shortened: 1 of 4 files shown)


zottto:
Hi Zoran / @s7design,

instead of looping through the dynamic fields and pushing them to the fields array, you should use the concat() function of JavaScript.

To avoid multiple new variables you can also do this inline:
```[Element1, Element2, Element3].concat(AnotherArray);```

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for this suggestion, we discussed abut the best solution her and I think it helps us a lot.
You will see there are more PR where we have used similar approach for the arrays

Regards
Zoran

s7design:
Hi @zottto 

PR is updated in regard to your suggestion.

Regards
Zoran

zottto:
Merged with 5f03db5.

Thanks,
Marc

PR #1309, opened 2016-06-28
Master - JS refactoring - AgentTicketCompose

s7design:
Hi @zottto 

Hereby is updated AgentTicketCompose module. We moved all JS from TT file to the new JS file.

Regards
Zoran


zottto:
Merged with ce667a7.

PR #1311, opened 2016-06-28
Master - JS refactoring - AgentStatistics

s7design:
Hello @zottto ,

Hereby is refactored JS for AgentStatistics modules. Included all relevant stats module in this one PR for easier review and merging. If this pose a problem, I can split them in separate PR's.

Mine start idea was to control initialisation of Add and Edit JS via Subaction, but it was not possible since on Edit screen, after inputing x-asis URL is gone and handled with AJAX. Perhaps if you know better solution for handling this issue let me know. Existing Selenium tests ran without issues with current refactoring, so perhaps we can go with this way.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentStatisticsAdd.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentStatisticsAdd.tt
@@ -84,9 +84,3 @@
         </form>
     </div>
 </div>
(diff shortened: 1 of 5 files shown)


zottto:
Merged with ee4b5f4.

Thanks,
Marc

PR #1313, opened 2016-06-28
Improved buttons

urbalazs:
Hi @mgruner 
I found some old-style buttons. I think, this can also be backported into rel-5_0.


mrcbnsls:
Thanks, looks nice. Merged into master in eef5ea8 and backported in 70153aa.

PR #1314, opened 2016-06-28
Marked privacy text as translatable

urbalazs:
Hi @mgruner 
I found some hard-coded German and English privacy texts. I removed the German one, and I marked the English as translatable. The German translations need to put into the German language file (`Kernel/Language/de.pm`) in order to avoid the double work for translate this again. Branch rel-5_0 is also affected. If you sync rel-5_0 with Transifex after commit, you can add the existing German translation to language file.


mgruner:
This was done on purpose as it is a legally binding text that we cannot have translated by volunteers, sorry.

PR #1316, opened 2016-06-29
Both types of move possible using MoveType parameter

pboguslawski:
This mod extends `Ticket::Frontend::MoveType` SysConfig parameter
with new value Dropdown and new window that allows one to use
both move operation types (dropdown and dialog) together.

With Dropdown and new window setting agents can easily choose
if they want to move ticket with comment using dialog or
to do just fast move using dropdown.

Related: https://dev.ib.pl/ib/otrs/issues/82
Author-Change-Id: IB#1053415

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-06-29 Both types of move possible using MoveType parameter.
  - 2016-06-24 Improved TicketSearch() to return error on a search for inexistent dynamic fields instead of ignoring them, thanks to Moritz Lenz.
  - 2016-06-06 Added checks for length of words in search terms when using search index/StaticDB backend.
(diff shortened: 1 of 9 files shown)


mrcbnsls:
Hi @pboguslawski,

Nice idea. Some annotations:

(1) I would suggest something like "both" instead of "Dropdown and new window" for the config value name.
(2) There should definitely be unit tests (or the existing ones be extended) for this.

Marc

pboguslawski:
Hi,

(1) "Dropdown and new window" is easier to understand in this context I think but feel free to adapt to your needs.

(2) Seems there already is unit test AgentTicketMove.t for testing ticket moving and this mod does not more than switching on both existing move controls so I don't see what should be tested more in UT. Feel free to adapt/extend UT to your needs.

Regards,
Pawel

mrcbnsls:
Closing. Could be re-opened once the new feature is fully covered by the corresponding UT.

PR #1317, opened 2016-06-29
Master - JS refactoring - AgentTicketActionPopupClose

s7design:
Hi @zottto 

Hereby is updated AgentTicketActionPopupClose.tt. We moved all JS code to existed Core.UI.Popup.js because it is included in every screen and send appropriate parameters in Popup.pm.
Please let me know if you have some comments regarding this PR.

Regards
Zoran


zottto:
Merged with 1484260.

PR #1318, opened 2016-06-29
Master - JS refactoring - AgentTicketPhone

s7design:
Hello @zottto ,

Hereby is refactored AgentTicketPhone JS. Moved JS from TT into separate Core.Agent.Phone.js file. Removed DynamicFieldNamesStrg param, since it is no longer need to use. DynamicFields array is passed into JS where we push values into array of other updatable fields. Perhaps if you are familiar with 'cleaner' solution for handling this transition, I will updated PR accordingly.

Regards,
Sanjin


zottto:
Merged with dc651a5.

Thanks,
Marc

PR #1322, opened 2016-06-30
Master - JS refactoring - AgentTicketProcess

s7design:
Hi @zottto 

Hereby are updated AgentTicketProcess and AgentTicketProcessSmall views. We moved JS code to existed Core.Agent.TicketProcess.js.

Regards
Zoran


zottto:
Merged with 50404b5.

Thanks,
Marc

PR #1324, opened 2016-06-30
Master - JS refactoring - CustomerTicketSearchResultShort

s7design:
Hello @zottto ,

Hereby is PR with removed CustomerTicketSearchResultShort JS. There is no need for this JS here anymore since Core.UI.js (InitMasterAction function) is handling this event.

Please let me know if you disagree with this change.

Regards,
Sanjin


zottto:
Merged with 0665f61.

Thanks,
Marc

PR #1328, opened 2016-07-01
Master - JS refactoring - AgentTicketActionCommon

s7design:
Hello @zottto ,

Hereby is refactored AgentTicketActionCommon JS. Moved code from TT to a new JS file. Created Core.Agent.TicketActionCommon.js instead of using existing Core.Agent.TicketAction.js since this JS is used in other modules apart from ones covered in ActionCommon. It could raise some unintentional duplication of JS if we use same file.

Please let me know if disagree with this PR or having any issues or comments.

Regards,
Sanjin


zottto:
Merged with 53961e2.

Thanks,
Marc

PR #1330, opened 2016-07-04
Master - JS refactoring - CustomerNavigationBar

s7design:
Hi @zottto 

Hereby is updated CustomerNavigationBar.tt. We moved JS code to the new JS file which is included in customer framework, because it is loaded in multiple modules.
Please let me know if you agree with these changes.

Regards
Zoran


zottto:
Hi Zoran / @s7design,

the code you moved to a new JS file is not really related to the NavigationBar, it is code to check for new chat requests. Therefore, please move the code to the generic Core.Customer.js and create an extra function for this code there (e.g. InitCheckChatRequests() or similar).

Thanks,
Marc

s7design:
Hi @zottto 

PR is updated regarding your suggestions.

Regards
Zoran

zottto:
Merged with 72ccff9.

Thanks,
Marc

PR #1331, opened 2016-07-04

urbalazs:
Hi @mgruner 
I found some strings related to stats, that are not marked for translation. This can backport into rel-5_0.


mgruner:
Merged and backported, thanks a lot!

PR #1332, opened 2016-07-05
Master - JS refactoring - AgentTicketSearch

s7design:
Hi @zottto 

Hereby is updated AgentTicketSearch module. We made changes as you suggested.
When we sent an array from PM to TT, it did not work and because of that, we decided to send a string which we use in JS.
We removed second block in TT because 'Core.Agent.Search.OpenSearchDialog( '[% Env("Action") %]' )' function call is placed in Ticket.xml at "Frontend::Module###AgentTicketSearch" config.
Please let me know if you have any comment regarding this PR.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketSearch.pm
+++ b/Kernel/Modules/AgentTicketSearch.pm
@@ -2175,6 +2175,9 @@ sub Run {
             }
         }
 
+        # attributes for search
+        my @SearchAttributes;
+
         # show attributes
         my @ShownAttributes;
         if ( $GetParamBackup{ShownAttributes} ) {
@@ -2216,12 +2219,8 @@ sub Run {
                 # show attribute
                 next ITEM if $AlreadyShown{$Key};
                 $AlreadyShown{$Key} = 1;
-                $LayoutObject->Block(
-                    Name => 'SearchAJAXShow',
-                    Data => {
-                        Attribute => $Key,
-                    },
-                );
+
+                push @SearchAttributes, $Key;
             }
         }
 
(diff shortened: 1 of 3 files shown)


zottto:
Hi Zoran / @s7design,

- you removed the bloch "Search", but now I can't find a place where the "OpenSearchDialog" function is called.

- the new function should not be called "AgentTicketSearch()" but should have a more useful name like "AddSearchAttributes" or similar

- try to call this function at the point where the HTML is injected to the DOM to only leave the Core.Config.Set() call in the tt file.

Thanks,
Marc

s7design:
Hi @zottto 

PR is updated regarding you suggestions.
We could not load JS file in modal dialog and because of that, Core.Config.Set() and OpenSearchDialog calls leave in TT file and sending parameters from PM to JS is impossible. What do you think about making mechanism of loading JS to modal dialog footer, like loading in FooterJS.tt? It is important for us because there are similar situations in other modules, e.g. ProcessManagement, Statistics, etc.
Please let me know if you have any comments about changes in PR and this proposal.

Regards
Zoran

zottto:
Hi Zoran,

I don't see a need to generate a comma-separated string for the search attributes on perl side and after that split this string again on JS side. I think, you can remove this part of the code.

I hope I can introduce a new functionality to add JS data also to AJAX calls soon, to remove the code from the template also for the search (and other templates). I will inform you, when this will be done.

Thanks,
Marc

zottto:
Hi Zoran / @s7design,

you can now use AddJSData() also for JS data from templates that are loaded via Ajax. The needed JS block is appended automatically to the end of the output.
With this improvement, you can now remove the JS block from the template which sets the Core.Config value.

The other block (with the OpenSearchDialog() call) is only used / needed, if the search is called as non-ajax to open the search dialog on an empty page. In this case, you can use again an init funtion in a JS file (and perhaps a check of subactions) to call the function outside of the template.

Thanks,
Marc

s7design:
Hi @zottto 

Thank you for your feedback. This PR is updated and rebased in regard to your suggestions.

Regards
Zoran

zottto:
Merged with ec8a2a3.

Thanks,
Marc

PR #1333, opened 2016-07-05
Master - JS refactoring - CustomerTicketZoom

s7design:
Hello @zottto,

Hereby is refactored CustomerTicketZoom module. Additional to JS refactoring there was an issue when uploading attachment on reply, return server error because $Param{AclAction} is undefined and couldn't do hash reference on undefined variable. Fixed it as well together with this PR.

Regards,
Sanjin


zottto:
Merged with f0e214a.

Thanks,
Marc

PR #1339, opened 2016-07-08
Master - JS refactoring - AdminGenericInterfaceWebservice

s7design:
Hello @zottto ,

Hereby is refactored AdminGenericInterfaceWebservice JS. Additionally there was missing 'fa fa-trash-o' class from TrashCan on delete action, added it.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericInterfaceWebservice.pm
+++ b/Kernel/Modules/AdminGenericInterfaceWebservice.pm
@@ -1145,6 +1145,17 @@ sub _ShowEdit {
                             %ActionData,
                         },
                     );
+
+                    # send data to JS
+                    $LayoutObject->AddJSData(
+                        Key   => 'ActionType',
+                        Value => $ActionData{ActionType}
+                    );
+                    $LayoutObject->AddJSData(
+                        Key   => 'Name',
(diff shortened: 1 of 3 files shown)


zottto:
Hi Sanjin / @s7design,

please name the config option for the AddJSData() call correctly to avoid calling the Core.Config.Set() in JS again with the correct config name.

Thanks,
Marc

s7design:
Hi @zottto 

I fixed this PR. Honestly it was our oversight, sorry for that. As I mentioned in one of the previous PRs, we had a problem with adding data with the key in format like "Config.SomaName", and so we use this weird approach to solve this solution. It was wrong definitely. I hope now it is good, if you would like another naming for web-service data please let me know. 

Regards
Sanjin

zottto:
Hi Sanjin / @s7design,

as discussed with Zoran: please update the AddJSData() call to only be called once with a Webservice data structure and keep the original naming on JS site.

Thanks,
Marc

s7design:
Hi @zottto 

As we have spoken via Skype, I updated this PR and renamed the JS data.

Regards
Zoran

zottto:
Merged with 1a85110.

Thanks,
Marc

PR #1340, opened 2016-07-08
Master - JS refactoring - AdminGenericInterfaceInvokerDefault

s7design:
Hello @zottto ,

Hereby is refactored AdminGenericInterfaceInvokerDefault JS.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminGenericInterfaceInvokerDefault.pm
+++ b/Kernel/Modules/AdminGenericInterfaceInvokerDefault.pm
@@ -39,6 +39,12 @@ sub Run {
         );
     }
 
+    # send data to JS
+    $LayoutObject->AddJSData(
+        Key   => 'WebserviceID',
+        Value => $WebserviceID
+    );
+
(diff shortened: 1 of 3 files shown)


zottto:
Hi Sanjin / @s7design,

the JS file does not match our coding standards (see OTRSCodePolicy checks), because the Init() function has more than one var statement.

Thanks,
Marc

s7design:
Hi @zottto 

I updated this PR. I removed insufficient "var" statements and also changed "for loop" with the "each" statement, it looks better in my opinion. Please check all of that because it is hard to test manually for me; we are not so familiar with this web-service features. 

Regards
Zoran

zottto:
Hi Zoran / @s7design,

var statementsa within an inner function  must be kept, otherwise the variable is active in the outer scope. In this case that means, that the "var URL" must be defined within the event functions.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks, I will rollback to the previous solution

Regards
Zoran

s7design:
Hi @zottto 

I updated this PR as you suggested 

Regards
Zoran

zottto:
Merged with b1ca2aa.

Thanks,
Marc

PR #1341, opened 2016-07-11
Master - JS refactoring - AdminProcessManagementPath

s7design:
Hello @zottto ,

Hereby is refactored AdminProcessManagementPath JS. Consider this PR after we resolve PR ( #1271 ), since there will be some merge conflicts and i will rebase this branch once we merge it.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminProcessManagementPath.pm
+++ b/Kernel/Modules/AdminProcessManagementPath.pm
@@ -327,7 +327,7 @@ sub _ShowEdit {
     # send data to JS
     for my $AddJSData (qw(TransitionEntityID ProcessEntityID StartActivityID)) {
         $LayoutObject->AddJSData(
-            Key   => 'Config.' . $AddJSData,
+            Key   => $AddJSData,
             Value => $Param{$AddJSData}
         );
     }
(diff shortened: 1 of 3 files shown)


zottto:
Hi Sanjin / @s7design,

if you name the config vars in the AddJSData() call correctly, you don't have to set them in the JS file again. Please use the correct config names used in the JS file directly in teh AddJSData() call.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for follow up, PR is rebased and updated.

Regards,
Sanjin

zottto:
Hi Sanjin / @s7design,

thanks for the rebase. Is there a reason, why the call for InitPathEdit() is delayed for a second?

Thanks,
Marc

s7design:
Hi @zottto 

I am working on this, in my opinion it doesn't work properly. I think this delay is not necessary, there we should find another approach.  

Regards
Zoran

s7design:
Hi @zottto 

I updated thi PR, and I hope I fxed it. After testing long time, I saw there is a problem if JS data is named "Config.SomeName". I renamed these data, actually I removed Config and there are the following ones:
- TransitionEntityID 
- ProcessEntityID 
- StartActivityID

 As I mentioned in the previous comment delay of 1 second was wrong idea. In that case there was not logged JS error but there also were not loaded right data on the screen.
Now  AdminProcessManagementPath  screen works fine. Please let me know what do you think about that.

Regards
Zoran

zottto:
Merged with e9f7580.

Thanks,
Marc

PR #1342, opened 2016-07-11
Master - JS refactoring - ArticleActionMenu

s7design:
Hi @zottto 

Hereby is updated ArticleActionMenu.tt. This is our proposal for refactoring this type of structure. Because this TT file does not have frontend module registration we could not move all JS code to the new JS file. Also, in TT file check up of 'ItemType' is executed and it is not needed to pass through all array of menu items in JS.

Regards
Zoran


zottto:
Hi Zoran / @s7design 

thanks for this PR. Although there is no matching frontend module like ArticleActionMenu.pm or so, this template is only used in AgentTicketZoom and is processed there. The needed JS can be put there.

Please move the JS completely to the AgentTicketZoom JS file.
I also think, that the remaining JS in the template can also be moved there:
The if-statement checks if the article action menu is loaded via AJAX or not. If not, the event bindings can happen in the Init of AgentTicketzoom. If loaded via AJAX, the bindings can be updated in the AJAX success callback function.

Thanks,
Marc

s7design:
Hi @zottto 

We will follow you suggestion and I thin it will be improved with your advice.

Regards
Zoran

s7design:
Hi @zottto 

PR is updated according to your suggestions. We inspected timeline view in OTRSBusiness and we think there will be some modifications when we refactor AgentTicketZoom with OTRSBusiness package.
Please let me know if you have any feedback about these changes.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

please do not change generic functionality to fix functionalitry in one frontend module! You changed ContentUpdate() of Core.AJAX with special code only for AgentTicketZoom.
Please revert that.
That part of the code can be executed in the callback of the ContentUpdate function in Core.Agent.TicketZoom.js.

Thanks,
Marc

s7design:
Hi @zottto 

Thank you for feedback. Callback of the ContentUpdate function is modified according to your comment.
Please let me know if you have any suggestion about these changes.

Regards
Zoran

zottto:
Merged with f0a299a.

Thanks,
Marc

PR #1345, opened 2016-07-12
Removed translation from log entry

urbalazs:
Hi @mgruner 
This is a log entry, and it shouldn't translate. Branch rel-5_0 is also affected.


mgruner:
Merged and backported, thanks Ur!

PR #1349, opened 2016-07-13
Master - JS refactoring - AgentOverview

s7design:
Hi @zottto 

Hereby are refactored TT files which are used in ticket overview screens. Every screen, e.g. AgentTicketQueue, AgentTicketService, AgentTicketStatusView etc., uses PM module in regard to selected view. Because of that, all JS code is put in one, new JS file. Please let me know if you have any suggestion about this PR.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -8348,6 +8348,7 @@
                     <CSS>Core.AllocationList.css</CSS>
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
                     <JavaScript>Core.Agent.TableFilters.js</JavaScript>
+                    <JavaScript>Core.Agent.Overview.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
@@ -8377,6 +8378,7 @@
                     <CSS>Core.AllocationList.css</CSS>
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
                     <JavaScript>Core.Agent.TableFilters.js</JavaScript>
+                    <JavaScript>Core.Agent.Overview.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
@@ -8597,6 +8599,7 @@
                     <CSS>Core.AllocationList.css</CSS>
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
                     <JavaScript>Core.Agent.TableFilters.js</JavaScript>
+                    <JavaScript>Core.Agent.Overview.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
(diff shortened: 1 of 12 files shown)


s7design:
Hi @zottto  

Yes it is so big and to be honest we was very happy when we had completed this refactoring for overview screens :)

Thanks for your suggestion. Yes we had a problem to solve this interpolation with Template data. I think we will be able to solve that on the way you suggested.

It will be done soon.

Regards
Zoran

s7design:
Hi @zottto 

In regard to your suggestions, PR is updated. Please let me know if you have some comment about changes.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

I can't find in the code where "UserAutocomplete.QueryDelay" is now set (after the setting was removed from the template. I thought that it must be in the perl file with AddJSData, but I can't find it. Can you enlighten me?

Thanks,
Marc

s7design:
Hi @zottto 

I will check it, it could be my mistake. I will update this PR if I find it is necessary 

Regards
Zoran

s7design:
Hi @zottto 

PR is updated following your comment. Please let me know if you have any feedback.

Regards
Zoran

s7design:
Hi @zottto 

PR is modified because access to "hash variables" was not possible like
`Core.Config.Get('CustomerUserAutocomplete.MaxResultsDisplayed')`
but
`Core.Config.Get('CustomerUserAutocomplete').MaxResultsDisplayed`

Regards
Zoran

zottto:
Merged with c8e9d36.

Thanks,
Marc

mgruner:
I had to revert this because the auto complete change mentioned by Zoran causes JS errors for example in the small queue view:

```
Cannot read property 'MinQueryLength' of undefined

Object.Core.Agent.TableFilters.TargetNS.InitCustomerIDAutocomplete@http://localhost/otrs6-web/js/Core.Agent.TableFilters.js:40:65
Object.Core.Agent.Overview.TargetNS.InitViewSmall@http://localhost/otrs6-web/js/Core.Agent.Overview.js:118:33
Object.Core.Agent.Overview.TargetNS.Init@http://localhost/otrs6-web/js/Core.Agent.Overview.js:88:22
```

A more robust solution will be needed as this variable is not always defined? Please clarify with MN how to proceed.

s7design:
Hello @mgruner ,

Thank you for spotting this issue. Will check it and update this PR accordingly.

Regards,
Sanjin

s7design:
Hello @mgruner & @zottto ,

Updated this PR with check in JS if AutoComplete config values are defined. This is one way of approaching this issue. To be honest these variables are repeating and carrying same values. I can see it is left for person with knowledge to customise to suit desired needs. 

Perhaps if you have some better solution for this issue, let me know. Additionally found small mistakes that was left from refactoring which is fixed now as well.

Regards,
Sanjin

zottto:
Merged (again) with c1e2125 (with some changes).

Marc

PR #1353, opened 2016-07-15
Master - JS refactoring - CustomerHTMLHead

s7design:
Hi @zottto 

Hereby is updated CustomerHTMLHead.tt. All JS code is moved to existed Core.Customer.js.

Regards
Zoran


zottto:
Hi Zoran / @s7design,

as already mentioned for the agent HTMLHead template, we keep the JS for the responsive functionality in the template to make sure it is executed very early.

Thanks,
Marc

PR #1354, opened 2016-07-15
Master - JS refactoring - HTMLHead

s7design:
Hi @zottto 

Hereby is updated HTMLHead.tt. JS code is moved to existed JS file. Please let me know if you have some feedback about this PR.

Regards
Zoran


zottto:
Hi Zoran / @s7design,

this is one of the rare conditions where I really want to keep the JS in the template file. As you can see in the original file, the JS is executed in the HTML head very early.
All the JS files are loaded later and therefore all JS execution hooks are executed even later which is too late.

So we keep this JS regarding the responsive CSS in the template.

Thanks,
Marc

PR #1363, opened 2016-07-22
Master - JS refactoring - PreviewWidget

s7design:
Hello @zottto ,

Hereby is refactored PreviewWidget in Statistic edit screen. There was in  PreviewResult $('.SwitchPreviewFormat').on('click', function(). It was sent by value Core.UI.AdvancedChart.Init. But now when we send reference to an array, there was a problem because of 
removing the first two elements from RowData
https://github.com/OTRS/otrs/blob/master/var/httpd/htdocs/js/Core.UI.AdvancedChart.js#L94

and I found another problem with this line in DrawStackedAreaChart stats:
https://github.com/OTRS/otrs/blob/master/var/httpd/htdocs/js/Core.UI.AdvancedChart.js#L624

So I add for ignoring Sum column: 
`if (typeof HeadingElement === 'undefined' ||  HeadingElement === 'Sum') {`

It is our proposal for fixing that, we can use also JSON.stringify and JSON.parse to avoid problem with changing data in an array when they are sent by reference.

Regards,
Zoran

### DIFF
--- a/Kernel/Output/HTML/Statistics/View.pm
+++ b/Kernel/Output/HTML/Statistics/View.pm
@@ -1107,6 +1107,12 @@ sub PreviewWidget {
         );
     }
 
+    # send data to JS
+    $LayoutObject->AddJSData(
+        Key   => 'PreviewResult',
+        Value => $Frontend{PreviewResult},
+    );
+
     my $Output .= $LayoutObject->Output(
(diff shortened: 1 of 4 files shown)


zottto:
Merged with 616daa9.

Thanks,
Marc

PR #1368, opened 2016-07-26
Master - JS refactoring -  RestrictionWidget

s7design:
Hi @zottto 

There is updated RestrictionWidget. JS code is moved from TT in JS Core.Agent.Statistics.js
I also remove some commented lines in View.pm, I think they are not necessary any more.

Regards
Zoran


zottto:
Merged with 3078019.

Thanks,
Marc

PR #1369, opened 2016-07-27
Marked JS alert messages as translatable

urbalazs:
Hi @zottto 
I found some JS alert messages, which is not marked as translatable.


zottto:
Merged with e31cbfa.

Thanks,
Marc

PR #1370, opened 2016-07-27
Master - JS refactoring - AgentCustomerSearch

s7design:
Hi @zottto 

There is finally refactored AgentCustomerSearch.tt. Actually this is second round of refactoring as we mentioned in the PR https://github.com/OTRS/otrs/pull/1277#event-706085052

To be honest, I looked into this in a detail and there were aome problem in the first round of refactoring  AgentCustomerSearch. I believe all of them are fixed now. 
I changed naming of JS data, as we have disscussed reccently, it is not possibly to set data with the key like 'CustomerSearch.ShowCustomerTickets'.

In the previous PR we set JS data in all screen where AgentCustomerSearch was included, however in my opinion AllowMultipleFrom is needed only for AgentTicketPhone, and ShowCustomerTickets is needed for three screen where Customer History is shown customer history tickets in AgentTicketPhone, AgentTicketEmail and AgentTicketCustomer. 

I hope the code is more better, if you have any suggestion please let me know.

Regards
Zoran


zottto:
Merged with e60accc.

Thanks,
Marc

PR #1371, opened 2016-07-28
Master - JS refactoring - Pagination

s7design:
Hi @zottto 

Hereby is updated Pagination.tt. All JS code is moved to new JS file Core.Pagination.js which is loaded on every screen.

Regards
Zoran


zottto:
Hi Zoran,

in this case I think we should refactor the whole code.

First of all, I think we do not need an extra JS file for the pagination code. It should be integrated in a generic JS file (Core.Agent or Core.App).

I also think, that it is not needed to bind events individually for every pagination link, e.g. for every page number there is an extra event binding, although the code is exactly the same (besides the number itself).

For all page numbers, define a class for the element (e.g. "PaginationNumber") which you can use for the event binding. Additionally, you can add the needed data for the event into data attributes of the HTML element.

You can do the same with all other back and forward links.

The goal is to have one (!) event binding function for all pagination links. If you need much data for every individual link you can send the data via AddJSData() in general and only put the most relevant data as data attribute to the HTML element (e.g. page number). All other code can than be grabbed from the data structutr provides through AddJSData() based on the page number retrieved from the data attribute.


As an example:
```
<a class="Pagination PaginationNumber" data-pagination-pagenumber="[% Data.PageNumber %]" href="#"[% Data.SelectedPage %] title="[% Translate("Show page %s", Data.PageNumber) | html %]">[% Data.PageNumber %]</a>
```
In JS:
```javascript
var ServerData = Core.Config.Get(PaginationData);
$('.PaginationNumber').off('click.Pagination').on('click.Pagination', function () {
    var PageNumber = Core.Data.Get($(this), 'pagination-pagenumber');
    var PaginationDataNumber = ServerData['Number'][PageNumber];
    // use PaginationDataNumber as base for all needed elements like
   // PaginationDataNumber.AjaxReplace, PaginationDataNumber.BaselinkAll, ...
    // ....
});
```

This is only one event binding for all pagination number links. If you refactor like this for all pagination links, that should reduce the number of events and the amount of code.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for suggestion we will update this. I will contact you if we need additional info.

Regards
Zoran

s7design:
Hello @zottto ,

This PR is updated with new pagination JS events. Please bear in mind that there are also some functions from AgentDashboardCommon refactoring. Had to include those as well in this PR since needed to check how pagination is working when certain events are triggered on dashboard it self. For example when preference is updated or when you refresh user online list, to make sure that pagination events are loaded.

Again this would be best to just to look for now if this way of approaching this refactoring is acceptable. Created events based on classes which are dynamically created depending from widget to widget. On those classes are events bind. Pagination is now performing very good from my way of testing. Only thing that does not work at this moment is when widget automatically refresh or when you sort certain columns. This will work as well once AgentDashboardTicketGeneric is refactored and those events are loaded appropriately.

Once agent dashboard common is merged i will rebase this PR for easier merging and solving conflicts. 

Please let me know if you think there should be some additional changes regarding passing certain variables from perl to JS. Do not hesitate to ask for any questions regarding this PR.

Regards,
Sanjin

s7design:
Hello @zottto ,

This PR is rebased and ready for review.

Regard,
Sanjin

zottto:
Hi @s7design / Sanjin,

the Pagination.tt is used in the LayoutObject (PageNavBar) which is a generic function. So it is not possible to onlyput the JS in the Core.Agent.Dashboard.js file.
As I said in a previous comment, please move the JS to a generic file (like Core.Agent).

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for pointing this out. Moved pagination JS to Core.Agent.js. 

Following on your other feedback regarding using Sub/Pub to call certain events when they are needed included those changes for pagination as well. Created Subscribe event for each widget separately which is called on initial executing and also in Core.AJAX.ContentUpdate() function on success. 

Now to be honest I am not really an expert on what's the downside of publishing non-existing channels but this is working really good now. Pagination event's are published specifically for widget on which action are taken.

If you have any questions regarding channel naming or some other feedback how can we improve this functionality further on, please do not hesitate to contact.

Regards,
Sanjin

s7design:
Hello @zottto ,

Updated this PR with your suggestion to subscribe on ContentUpdate channel. Included also check to see which widget we updated and to initialise pagination accordingly just for that widget.

Want to thank you again for your help regarding this issue, this has been very interesting and helped me to learn/improve a lot.

Please let me know what do you think about this PR now, and don't hesitate to contact for any clarifications or issues. Looking forward to your reply.

Regards,
Sanjin

zottto:
Hi Sanjin / @s7design,

there is still a "PaginationDataDashboard" in the JavaScript file. Is this intended?

Marc

s7design:
Hello @zottto ,

Yes this is intended, these configs are set in Layout.pm (line 2774):

`# send data to JS
    if ($Param{AJAXReplace}) {
        $Self->AddJSData(
            Key   => 'PaginationData' . $WidgetName,
            Value => \%PaginationData
        );
    }`

These configs contains different widgets that require pagination. PaginationEvent() is initiated on each of those widgets separately. Please let me know if this clarifies your questions.

Regards,
Sanjin

zottto:
Sanjin / @s7design,

I can't see how this should work outside of the Dashboard with a hardcoded "Dashboard" string in the JS file.

Marc

s7design:
Hello @zottto ,

My apologise, now I see what you mean with that comment. Removed 'Dashboard' from config string. Rebased and updated this PR.

Please let me know if you see any more issues.

Regards,
Sanjin

zottto:
Merged with af33f56.

Thanks,
Marc

PR #1372, opened 2016-07-29
Fixed error Base class package "Kernel::System::EventHandler" is empty

pboguslawski:
When running module-tools/DatabaseInstall.pl in Debian 8 - probably after
perl update https://www.debian.org/security/2016/dsa-3628 - an error
occurs

    Base class package "Kernel::System::EventHandler" is empty.
        (Perhaps you need to 'use' the module which defines that package first,
        or make that module available in @INC (@INC contains: /opt/Custom /opt/Kernel/cpan-lib /opt Kernel/cpan-lib [...] .).
     at Kernel/System/Package.pm line 25.
    BEGIN failed--compilation aborted at Kernel/System/Package.pm line 25.
    Compilation failed in require at /opt/module-tools/DatabaseInstall.pl line 54.
    BEGIN failed--compilation aborted at /opt/module-tools/DatabaseInstall.pl line 54.

This mod fixes this issue.

Related: https://dev.ib.pl/ib/otrs/issues/83
Author-Change-Id: IB#1055118

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-07-29 Fixed error Base class package "Kernel::System::EventHandler" is empty.
  - 2016-07-26 Added a new postmaster filter to decrypt and handle encrypted mails.
  - 2016-07-18 Fixed bug#[7860](http://bugs.otrs.org/show_bug.cgi?id=7860) - AgentTicketSearch and Statistics are missing TicketPending option.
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @pboguslawski, this looks really weird. Why should Kernel::System::Package be found, but not Kernel::System::EventHandler? We use this valid Perl feature (use base to also load the class) in several places, so fixing it only in one makes not so much sense. Maybe a fix must be applied in DatabaseInstall instead? I'll close this PR for now, feel free to follow-up on this please.

pboguslawski:
Seems that changing Package,pm is not necessary. Please see http://bugs.otrs.org/show_bug.cgi?id=12265 for more details.

PR #1373, opened 2016-07-29
Master - JS refactoring - RichTextEditor 

s7design:
Hi @zottto 

Hereby are updated RichTextEditor.tt and CustomerRichTextEditor.tt. All content from 'JSOnDocumentComplete' TT block is moved to new PM functions where AddJSData() is used for set up rich text editor properties.
In AgentTicketProcess and CustomerTicketProcess modules, there are different cases because of AJAX call. Core.UI.RichTextEditor.Init() function call in Core.TicketProcess.js is used for initialization rich text editor for both cases.
Please let me know if you have any feedback for this PR.

Regards
Zoran


zottto:
Hi Zoran / @s7design,

I changed the handling of Core.Config to make AddJSData easier: you don't have to put 'Config' as prefix anymore in deep data structures. Please change that to make the data structure more readable.

Second thing to change would be the function naming. RichTestSetBlock is not a very speaking name, please rename it to SetRichTextParameters or similar.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks for follow-up, it will done soon

Regards
Zoran

s7design:
Hi @zottto 

This PR is updated, I hope it is better now.

Regards
Zoran

zottto:
Merged with ab870fd.

Thanks,
Marc

PR #1374, opened 2016-08-01
Master - JS refactoring - StatsParmsWidget

s7design:
Hi @zottto

Hereby is updated StatsParmsWidget.tt. All JS code is moved to JS file Core.Agent.Statistics.js which is also loaded on Dashboard screen, it is needed for Statistic widget on Dashboard.

Regards
Zoran

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -8804,7 +8804,7 @@ via the Preferences button after logging in.
                     <JavaScript>Core.UI.AllocationList.js</JavaScript>
                     <JavaScript>Core.Agent.TableFilters.js</JavaScript>
                     <JavaScript>Core.Agent.Dashboard.js</JavaScript>
-                    <JavaScript>Core.Agent.Statistics.js</JavaScript>
+                    <JavaScript>Core.Agent.Statistics.ParamsWidget.js</JavaScript>
                 </Loader>
             </FrontendModuleReg>
         </Setting>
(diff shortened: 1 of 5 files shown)


zottto:
Hi Zoran,

although the StatsParamsWidget is used in the statistic screen I would not recomment to put the JS into the Core.Agent.Statistics.js file.
The code is also used in the dashboard and it is not a good idea to load the complete statistics JS also in the dashboard, especially because the init function of Core.Agent.Statistics initializes a lot more, which is not available in the dashboard or at least could interfere with dashboard elements.
What about an own Core.Agent.Statistics.ParamsWidget.js file that can be easily included in the dashboard and also in the statistic screens?

Thanks,
Marc

s7design:
Hi @zottto 

The PR is updated, there is new JS Core.Agent.Statistics.ParamsWidget.js who is called in dashboard and also in the statistic screens.

Regards
Zoran

zottto:
Hi Zoran,

I see one more problem in this PR:

In the original tt file, the input field initialization is not wrapped within an JSOnDOcumentComplete block. That means, this code is not only executed once the page is loaded but everytime the according HTML is injected into the page (which happens, if the widget code is loaded via AJAX).

Executing the inputfield code only in the init method of the new JS file is not enough, because this is not executed, if the widget content is loaded via AJAX.

Thanks,
Marc

s7design:
Hi @zottto 

Thanks we will see what is possible in that case. I had a look on that, but there is a problem with StatsParam in Dashboard screen. It could be solved with some PRs which there are still opened. I mark this PR for updating when Dashboard refactored PRs will be closed. 
Then we are planing to update and provide better solution here.

To be honest I did not find any situation when $Param{AJAX} in StatsParamsWidget, in that case 

-[% IF Data.AJAX %]

and refactored one:
+        if (StatsParamData.Ajax) {
+            Core.UI.InputFields.Activate();
+        }

is not needed any more. Could you help about that.

Regards
Zoran

zottto:
HI Zoran / @s7design,

dashboard widget can be refreshed every x minutes. If thsi settins is enabled, the content of the widgets is loaded via AJAX.

Can you please chack this PR again and update it?

Thanks,
Marc

s7design:
Hello @zottto ,

This PR is updated. To be honest, searched everywhere how dashboard statistics can be updated via AJAX and it stills puzzles me. 
Anyhow enabled Core.UI.InputFields.Activate() if $Param{AJAX} if passed in StatsParamsWidget function, so looking at the code it will be executed. PR is rebased as well and ready for review.

Let me know if there are further issues with this refactoring.

Regards
Zoran

zottto:
Merged with 5a13458.

Thanks,
Marc

PR #1379, opened 2016-08-02
Used File::Spec to construct paths in restore.pl

fvictorio:
Use the catfile subroutine from File::Spec to build the paths, instead
of doing "$dir/File.ext". This way, the built paths won't contain double
slashes if the given $dir ends with a slash.

Multiple slashes are a valid separator in Unix-based systems, as far as
I know, but at the very least this is cleaner and the output messages
are less confusing.

### DIFF
--- a/scripts/restore.pl
+++ b/scripts/restore.pl
@@ -23,6 +23,7 @@ use warnings;
 
 # use ../ as lib location
 use File::Basename;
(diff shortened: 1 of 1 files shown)


mgruner:
Applied in fbb864709a3d8dc443d49e1acea884a468d8ede7, thanks a lot @fvictorio!

PR #1380, opened 2016-08-02
Master - JS refactoring - LinkObject

s7design:
Hi @zottto 

Hereby is refactored LinkObject.tt. Part of JS code for 'Simple' table is moved to Core.Agent.LinkObject.SearchForms.js and part for 'Complex' table is moved to Core.Agent.LinkObject.js because 'Complex' view mode has to be set and some JS and CSS files have to be included in SysConfig.
Everything is ok when complex table is shown for AgentTicketZoom screen, but there were errors when we tried to make the same situation in AgentLinkObject screen (function Core.Agent.TableFilters.SetAllocationList reported error because columns data arrays are not defined).

Regards
Zoran


zottto:
Merged with a2aa458.

Thanks,
Marc

PR #1385, opened 2016-08-05
Master - JS refactoring - AgentTicketZoom

s7design:
Hi @zottto 

Hereby is refactored AgentTicketZoom view. JS code is moved to existed Core.Agent.TicketZoom JS file.
Instead of Core.Agent.TicketZoom.Init() call (which is removed), 'ArticleTableHeight' is sent from PM to JS and hook is added.
JS code from block 'TicketMenuPhoneAsDropdown' is removed because this is added to Core.Agent.TicketZoom JS file when ArticleActionMenu.tt was refactored.
Please let me know if you have any suggestion related to these changes.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketZoom.pm
+++ b/Kernel/Modules/AgentTicketZoom.pm
@@ -265,6 +265,12 @@ sub Run {
         return $LayoutObject->NoPermission( WithHeader => 'yes' );
     }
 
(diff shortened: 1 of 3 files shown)


zottto:
Merged with abdea53.

Thanks,
Marc

PR #1386, opened 2016-08-05
Master - JS code improvement - FieldUpdate

s7design:
Hello @zottto ,

Hereby is suggestion to improve JS code for Fields that get updated on change. Idea behind this is to reduce amount of code in JS and allow easier future updated if and when needed.

Creation of field events is moved into private function for each JS that felt had value doing it. Perhaps if you think we can create global function of this type and then use it in each file separate instead of this way, I will updated PR. Your input on this is welcomed. 

Please take this idea into consideration and let me know if you have any questions or suggestion regarding this PR. 

Regards,
Sanjin

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.TicketActionCommon.js
+++ b/var/httpd/htdocs/js/Core.Agent.TicketActionCommon.js
@@ -35,22 +35,6 @@ Core.Agent.TicketActionCommon = (function (TargetNS) {
             Fields = ['TypeID', 'ServiceID', 'SLAID', 'NewOwnerID', 'NewResponsibleID', 'NewStateID', 'NewPriorityID'],
             ModifiedFields;
 
-        /**
-         * @private
-         * @name FieldUpdate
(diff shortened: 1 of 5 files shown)


zottto:
Hi Sanjin / @s7design,

the looks like a nice optimization of the code. Some recommendations:

1. private functions can be moved outside of public functions

2. Why do you concat other fields to a variable for DynamicFields:

```
DynamicFieldNames = Core.Config.Get('DynamicFieldNames').concat(['SignKeyID', 'CryptKeyID', 'StandardTemplateID']);
```

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for feedback. PR is updated, moved private functions outside public functions.

Regarding your second comment, initially those were fields that could trigger change on other fields, so i left them also to be included. After some more research, mine findings are that those fields are not able to update or be updated by other fields, so i excluded them.

Regards,
Sanjin

zottto:
Merged with cbf5d15.

Thanks,
Marc

PR #1387, opened 2016-08-08
JS Unit test - Core.Data

s7design:
Hi @zottto,

Hereby is updated Core.Data.UnitTest.js. There are new test cases which test Core.Data.CompareObject and Core.Data.CopyObject function from Core.Data.js

Whit this PR we are starting updating JS unit tests if you have any suggestion please let me know.

Regards
Zoran


zottto:
Merged with 72d28e9.

If you add more unit tests, please add more speaking test descriptions than "okay".

Thanks,
Marc

PR #1388, opened 2016-08-08
JS Unit test - Core.Form

s7design:
Hi @zottto,

Hereby is updated Core.Form.UnitTest.js. There are new test cases which test Core.Form.SelectAllCheckboxes and Core.Form.InitSelectAllCheckboxes function from Core.Form.js

Please let me know if we can improve this.

Regards
Zoran


zottto:
Merged with 5c3cb80.

Thanks,
Marc

PR #1390, opened 2016-08-09
JS Unit test - Core.Exception

s7design:
Hi @zottto,

Hereby is updated Core.Exception.js. There is fixed function IsErrorOfType, which did not work properly.
There is only one test where is tested function from the JS:

- ApplicationError
-    GetType
- HandleFinalError
- IsErrorOfType
- Thrown

Please let me know if we can improve this.

Regards
Zoran


zottto:
Merged with 0846ac9.

Thanks,
Marc

PR #1391, opened 2016-08-10
Master - JS refactoring - AgentDashboardUserOnline

s7design:
Hello @zottto ,

Hereby is refactored AgentDashboardUserOnline module. Please take into consideration that this is closely connected with AgentDashboardCommon module. Therefore some functionality like refresh user list now remove events, which will work when AgentDashboardCommon refactoring is merged. I will rebase this PR and modify it accordingly once we have that refactoring in master.

Please let  me know if you have any issues or questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Dashboard/UserOnline.pm
+++ b/Kernel/Output/HTML/Dashboard/UserOnline.pm
@@ -421,6 +421,16 @@ sub Run {
         AJAX => $Param{AJAX},
     );
 
+    # send data to JS
+    $LayoutObject->AddJSData(
+        Key   => 'UserOnline',
+        Value => {
+            Name        => $Self->{Name},
+            NameHTML    => $NameHTML,
+            RefreshTime => $Refresh,
+        },
+    );
+
     return $Content;
 }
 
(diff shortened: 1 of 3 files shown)


s7design:
Hello @zottto ,

This PR is rebased and ready for review.

Regards,
Sanjin

zottto:
Hi Sanin / @s7design,

some recommendations:
- don't use unbind / bind but only off/on functions
- if you check for an undefined value with typeof to call some larger part of code it would be nicer to exit early like

```
if (typeof variable === undefined) {
    return;
}
```

to avoid large indents.

- to avoid calls of InitUserOnline() in multiple callback functions all over the dashboard, please use PubSub events (Core.App.Subscribe and Core.App.Publish) to manage when to call InitUserOnline() again after a ContentUpdate. Please check to only call InitUserOnline if the widget itself is updated, not with every (other) ContentUpdate call in the dashboard.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank for all the feedback regarding this PR. Updated and implemented your suggestions.

Please let me know if you have any further issues or questions.

Regards,
Sanjin

zottto:
Merged with some changed with 0f15d1e.

Thanks,
Marc

PR #1394, opened 2016-08-11
Extended searching for customers and customer users

pboguslawski:
OTRS does not allow to search for customer users using first name
and surname separated with space (natural way of searching) -
if one search for "john black", fields specified
in CustomerUserSearchFields will be searched for this whole phrase
individually  and customer user with firstname = John
and surname = Black won't match. To be able to search using
firstname and surname, one need to use "john+black" syntax which
is not obvious.

This mod introduces new parameters for CustomerUser
and CustomerCompany searching maps (to be configured in Config.pm):

```
CustomerUserSearchExtended => 0,   # 0 - normal search (space is part of search phrase), 1 - extended search (space works like +)
CustomerCompanySearchExtended  => 0,   # 0 - normal search (space is part of search phrase), 1 - extended search (space works like +)
```

When set to 1, OTRS treats whitespaces in searched string same
like +; when searching for "john black" all searched fields will
be searched separate for john and black.

Related: https://dev.ib.pl/ib/otrs/issues/84
Author-Change-Id: IB#1055633

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-08-11 Extended searching for customers and customer users.
  - 2015-08-01 Updated CPAN module CGI to version 4.32.
  - 2016-07-26 Added a new postmaster filter to decrypt and handle encrypted mails.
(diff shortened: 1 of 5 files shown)


mgruner:
Hello @pboguslawski, OTRS master now has a rewritten address book that includes a customer and customer user detail search, where every field can be searched separately. I guess that this would solve the issue you describe. Also, there is a new API to use this search in custom modules.
Best regards, Martin

PR #1395, opened 2016-08-12
Improved wording

urbalazs:
Hi @mgruner,

I think, the word "preferred" is better here in this case. Agents have to select queues, because they work with them, not because they love them.
And there is "preferred" in the caption of OTRS Business Solution chat module, but in the description is still "favorite":
http://otrs.github.io/doc/manual/otrs-business-solution/stable/en/html/otrs-business-solution-features.html#id-1.4.2.4.4
If you apply this modification, can you change this also in OTRS Business Solution?


mgruner:
Thanks, updated, also in OTRS Business Solution™, where I also added hu, id and th_TH language translations. Thanks @urbalazs!

urbalazs:
Hi @mgruner, I updated the Hungarian translation of OTRS Business Solution™. Now it is 100% again!

PR #1397, opened 2016-08-12
Master frontend usability - AdminAttachment

s7design:
Hi @zottto,

Hereby we started refactoring  frontend usability. There is added breadcrumb link for AdminAttachment screens (Overview, Add, Edit). As you can see I changed code a little bit. This is first PR and would like to see with this PR how deep we should go with this task. 
For example I removed block which are not necessary, because IF directives are enough for selecting HTML element in TT module. I am not sure if you like such approach. In any case I will be able to provide solution that is the best one. In this moment it is only our proposal.

You can see that I did not use simple redirect in ContinueAfterSave action as it has been done in AdminProcessManagement and similar modules. I wanted to add Notify - Attachment updated!

There is also extended Selenium test with new test ceases for form actions: 

- 'Save' and 'Save and finish'
- breadcrumb links 
Would you like extending  Selenium tests.

We have some opened task about refactoring JS and JS unit test, whit this PR I would like provide solution for new task in order you have time to give us feedback and after that we will be able to continue refactoring other modules on the same way.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminAttachment.pm
+++ b/Kernel/Modules/AdminAttachment.pm
@@ -91,16 +91,34 @@ sub Run {
                 UserID => $Self->{UserID},
             );
             if ($Update) {
-                $Self->_Overview();
-                my $Output = $LayoutObject->Header();
-                $Output .= $LayoutObject->NavigationBar();
-                $Output .= $LayoutObject->Notify( Info => Translatable('Attachment updated!') );
-                $Output .= $LayoutObject->Output(
-                    TemplateFile => 'AdminAttachment',
-                    Data         => \%Param,
-                );
-                $Output .= $LayoutObject->Footer();
-                return $Output;
+                if ( $ParamObject->GetParam( Param => 'ContinueAfterSave' ) eq '1' ) {
+
+                    # if the user would like to continue editing the attachment, just redirect to the edit screen
+                    my $ID = $ParamObject->GetParam( Param => 'ID' ) || '';
+                    my %Data = $StdAttachmentObject->StdAttachmentGet(
+                        ID => $ID,
+                    );
+
+                    my $Output = $LayoutObject->Header();
+                    $Output .= $LayoutObject->NavigationBar();
+                    $Output .= $LayoutObject->Notify( Info => Translatable('Attachment updated!') );
+                    $Self->_Edit(
+                        Action => 'Change',
+                        %Data,
+                    );
+
+                    $Output .= $LayoutObject->Output(
+                        TemplateFile => 'AdminAttachment',
+                        Data         => \%Param,
+                    );
+                    $Output .= $LayoutObject->Footer();
+                    return $Output;
+                }
+                else {
+
+                    # otherwise return to overview
+                    return $LayoutObject->Redirect( OP => "Action=$Self->{Action}" );
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


mgruner:
Just a side note from me: I’m very much in favor of replacing blocks with TT conditionals where it makes sense.

Martin Gruner
Team Lead R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

s7design:
Hi @zottto 

Working on PR https://github.com/OTRS/otrs/pull/1401/files I decided to add event for #SubmitAndContinue button in Core.Agent.js
As you can see in the previous commit on this PR,  I put this event in Core.Agent.Admin.Attachment.js at first ( https://github.com/OTRS/otrs/pull/1397/commits/a7ac1879b50028da16e3ad5c18b8807956f0d941 )
Also I changed Change action a little bit in controller. I think it is better now. Please let me know what do you think about that.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

thanks for this PR. 

There are many things that can be improved in the usability of the agent interface. Although I understand that you are also motivated to change many things, please reduce the amount of changes within a PR, so that we can still handle them and are able to check them correctly. :-)

For this PR I would recommend to not change or add notifications as we will improve the notification system later for OTRS 6.

"Save and finish": I saw yesterday that there is module-specific JS for this button in some other modules. I recommend to create a private function in  Core.App or Core.Agent which initializes the click binding for this button globally as you did it in this PR already. But please create an extra function for this to seperate it.

Breadcrumbs: we talked aboput that yesterday and would love to see another optimiztaion here. We want to avoid to add the same HTML structure to every admin module, so it would be great to have the breadcrumb separated in an extra snippet and to be included to every admin module.

```
[%- INCLUDE "Includes/Breadcrumb.tt" 
    Path = [
        {
            Name => 'abcd',
            Link => '...'
        },
        {
            Name => 'abcd',
            Link => '...'
        },
    ]
-%]
```

Along with the include, the needed data is defined (where "Link" is optional).

Thanks,
Marc

s7design:
Hi @zottto 

I updated this PR with the solution which you suggested. As you can see there is two possible options i the two previous commit. 
With this solution, if there is only one item in breadcrumb it is also link (Subaction= Overview):
https://github.com/OTRS/otrs/pull/1397/commits/48dd4b3e669be2258f03e571be35fad2dc2dba27

In this commit it is solved, but there ia complicated code, because Path is defined for each option:
https://github.com/OTRS/otrs/pull/1397/commits/367e322dcc9d6027c7b2e30cf38daa6b24318f53

Please let me know what do you think about these solution, I will be able to update PR if you have any advice how we can improve it. I would like that there is some generic solution that cover as many as modules at the best way.

Regards
Zoran

s7design:
Hi all,

I updated the PR and implemented solution that Martin suggested. I hope it is much better the it was before. We are closer to the final solution, of course if you have any other suggestion please let me know.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

this looks really nice, but I have one recommendation:

The name of the first breadcrumb element is now "AdminAttachment" (not translatable), but the screen name (from the heading above) is:

```
<h1>[% Translate("Attachment Management") | html %]</h1>
```

So I recommend to name the breadcrumb element like the heading.

If you add an attachment, the breadcrumb says "Add New Attachment", while the widget below says "Add Attachment". This should also be the same.

And if you edit an existing attachment, the only output in the breadcrumb is the attachment name, but the widget below is named "Edit Attachment". So my recommendation would be "Edit Attachment: Name".

Please update the naming in this and all other breadcrumb-related PRs.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for feedback. Will update this PR and other related ASAP.

Regards
Zoran

s7design:
Hi @zottto 

This PR is updated in accordance with your feedback.

Regards
Zoran

zottto:
Merged with c430521.

Thanks,
Marc

PR #1405, opened 2016-08-18
Master frontend usability - AdminCloudServices

s7design:
Hello @zottto ,

Hereby is added breadcrumb for AdminCloudService module. Working on in this module, few questions were raised regarding how we should continue. Please if you could give your input:

- This module is 'somewhat' simple, meaning there is no further action then AdminCloudService screen. For the consistency reason, think we should add also breadcrumb on these type of modules.

- There are few link which are redirecting into other modules (e.g. AdminRegistration, AdminCloudServiceSupportDataCollector and AdminOTRSBusiness). Once any of these links are clicked, and screen is loaded breadcrumb for that module will be initiated. How do you think we should handle breadcrumb path from where we arrived. This will be also more impacting on e.g. AdminUser. We can arrive on this screen from different modules, AdminRoleUser, AdminUserGroup, ect...

Your input on these points is more then welcome.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminCloudServiceSupportDataCollector.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminCloudServiceSupportDataCollector.tt
@@ -9,6 +9,19 @@
 <div class="MainBox ARIARoleMain LayoutFixedSidebar SidebarFirst">
     <h1>[% Translate("Cloud Service Management") | html %] » [% Translate("Support Data Collector") | html %]</h1>
 
+    [% BreadcrumbPath = [
+            {
+                Name => Translate('Cloud Service Management'),
+                Link => 'AdminCloudServices',
+            },
+            {
+                Name => Translate('Support data collector'),
+            },
+        ]
+    %]
+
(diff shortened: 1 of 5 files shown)


s7design:
Hi @zottto 

This PR is modified in accordance with your suggestion.

Regards
Zoran

zottto:
Hi Zoran / @s7design,

for the link Action=AdminCloudServiceSupportDataCollector;CloudServiceName=Support%20data%20collector
which is part of this module, there is no breadcrumb...

Thanks,
Marc

s7design:
Hello @zottto ,

Thanks for feedback. PR is updated. Please if you could just clarify points mentions in first comment. Thanks 

Regards,
Zoran

zottto:
Merged with 1b95be0.

Breadcrumbs are for navigation within one module. If the users leaves the module with a link into another module, the other module shows its normal breadcrumb, based on the structure of this other module. So, the breadcrumb does not need an adjustement if the user is coming from another module with another breadcrumb.

Thanks,
Marc

s7design:
Hello @zottto ,

Thanks for clarification. We will continue working in that fashion then.

Regards
Zoran

PR #1408, opened 2016-08-19
Fixed description, improved wording

urbalazs:
Hi @mgruner 
I changed the description of module registrations, as this string is already in the Ticket.xml.
I also fixed a wrong description, the setting is not for agents, but for customers.

Can you also backport this to rel-5_0 and sync with Transifex? Let's give translator a chance to translate the modified strings before the 5.0.13 will be released.
My problem is, that now there is only one translation sync within a release period, when @carlosfrodriguez announces it as a "Prepare for release" commit. With this method OTRS never will be fully translated, as new strings are syncing right before release. If a translator finish them on Transifex, the translated strings will be added only to the next release, not to the current one. So new and modified strings remain untranslated in every OTRS release. Would it be possible to introduce a "string freeze" period? I mean, that Carlos usually closes the code on Thursday, the new tarball is available usually on the next Tuesday. In this few days no commits should be applied, which affect the language files, so Carlos can do an other Transifex sync some days later, but some days before the actual release. Can you discuss my "string freeze" suggestion?

### DIFF
--- a/Kernel/Config/Files/ProcessManagement.xml
+++ b/Kernel/Config/Files/ProcessManagement.xml
@@ -7,7 +7,7 @@
         <Setting>
             <FrontendModuleReg>
                 <Group>admin</Group>
-                <Description>Admin</Description>
+                <Description Translatable="1">This module is part of the admin area of OTRS.</Description>
                 <Title Translatable="1">Process Management</Title>
                 <NavBarName>Admin</NavBarName>
                 <Loader>
@@ -37,7 +37,7 @@
         <Setting>
             <FrontendModuleReg>
                 <Group>admin</Group>
-                <Description>Admin</Description>
+                <Description Translatable="1">This module is part of the admin area of OTRS.</Description>
                 <Title Translatable="1">Process Management Activity GUI</Title>
                 <Loader>
                     <CSS>Core.Agent.Admin.ProcessManagement.css</CSS>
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @urbalazs 

Sorry to jump in, but I saw my name...

The problem with your proposal is that the release is actually build on Wednesday / Thursday for internal testing and public published on next Tuesday, but they are the same files, no rebuild is done unless something really bad is detected internally.

On the other hand I understand your point that OTRS never gets all the translations at the correct time for each release.

@mgruner I believe that this is a valid topic to discuss.

((enjoy))
cr

urbalazs:
@carlosfrodriguez what if, that
- string freeze on Tuesday
- code closing and prepare for release on Thursday
- release on next Tuesday
From Tuesday to Thursday no commits should be applied, which affect the language files. I think 2 days is enough to do the translation work.

carlosfrodriguez:
Hi @urbalazs,

Thank you for your suggestion, I will evaluate this proposal together with others that I have in mind with Martin.

((enjoy))
cr

urbalazs:
Hi @carlosfrodriguez and @mgruner 
Can you discuss my "string freeze" suggestion? Are you planning to release OTRS 5.0.13 in the next week? If so, can you apply this PR and sync translations with Trasifex in order to translators be able to translate all the missing strings?

urbalazs:
Hi @mgruner 
This is an old pull request, but it contains an important wording change. Can you merge it?

mgruner:
Merged and backported to OTRS 5, thanks a lot @urbalazs!

PR #1415, opened 2016-08-24
Master - JS refactoring - Datepicker

s7design:
Hi @zottto 

Hereby is updated Datepicker.tt. Core.Config.Set() is replaced with AddJSData function in Layout.pm.
Please let me know if you have any comment on this.

Regards
Zoran


zottto:
Merged with 035cbd9.

Thanks,
Marc

PR #1416, opened 2016-08-24
Fixed charset handling in content type of SOAP messages

jepf:
Hi,

I encountered a problem regarding the content type of SOAP messages.

When sending SOAP messages with a content type like the following:
`Content-Type: application/soap+xml;charset=UTF-8;action="urn:MyService/MyAction"`

Kernel::System::Encode will throw warnings like this:
`Not supported charset 'utf-8;action="urn:MyService/MyAction', fallback to 'iso-8859-1'!`

The reason is the regex in Kernel::GenericInterface::Transport::HTTP::SOAP which is not prepared for a content type that does not end after the charset information.

Please have a look at my proposal for the regex improvement.

Regards,
   Jens Pfeifer

### DIFF
--- a/Kernel/GenericInterface/Transport/HTTP/SOAP.pm
+++ b/Kernel/GenericInterface/Transport/HTTP/SOAP.pm
@@ -206,7 +206,7 @@ sub ProviderProcessRequest {
 
     # convert charset if necessary
     my $ContentCharset;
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @jepf, very nice, thanks a lot! I merged this and backported to OTRS 5 and 4 (see http://bugs.otrs.org/show_bug.cgi?id=12305).

For the future you would help us if you (and your colleagues) could create a bug entry for such cases as well, as that is what it is really, a bug.

Again, thanks a lot! 👍

PR #1420, opened 2016-08-25
Master frontend usability - AdminGenericAgent

s7design:
Hello @zottto ,

Hereby is added breadcrumb and form action buttons for AdminGenericAgent screens. Used already translatable string 'Add job', 'Edit job' and 'Run job' as last breadcrumb strings. IMO having 'Job Settings' is not quite valid since you can't distinguish when you add or edit job. 

If you disagree on these names or have any issues regaridng this PR please let me know.

Regards,
Sanjin


zottto:
Merged with a5b93ff.

Thanks,
Marc

PR #1421, opened 2016-08-25
JS Unit test - Core.UI.Dialog.UnitTest.js

s7design:
Hi @zottto 

Hereby is UT for Core.UI.Dialog.js. Dialogs are created and tested its properties.
Please let me know if you have any comment on this PR.

Regards
Zoran


zottto:
Merged with 45e9886.

Thanks,
Marc

PR #1424, opened 2016-08-26
Master frontend - AdminMailAccount delete button confirmation

s7design:
Hello @mgruner & @zottto ,

Hereby is our proposal for delete confirmation dialog appearance in AdminMailAccount screen. If you agree on this, there is AdminGenericAgent module in which could be used same functionality.

Delete function is appended in Core.Agent.Admin.MailAccount.js. This is similar function which is used for AdminDynamicField delete events, so perhaps we could refactor these functions and place them in some of JS which is loaded in framework.

Selenium test is modified to check for these changes.

Please let me know what do you think about this proposal, and if you have any issues or questions do not hesitate to ask.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminMailAccount.pm
+++ b/Kernel/Modules/AdminMailAccount.pm
@@ -76,7 +76,12 @@ sub Run {
         if ( !$Delete ) {
             return $LayoutObject->ErrorScreen();
         }
-        return $LayoutObject->Redirect( OP => 'Action=AdminMailAccount' );
+        return $LayoutObject->Attachment(
+            ContentType => 'text/html',
+            Content     => $Delete,
+            Type        => 'inline',
+            NoCache     => 1,
+        );
     }
 
     # ------------------------------------------------------------ #
--- a/Kernel/Output/HTML/Templates/Standard/AdminMailAccount.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminMailAccount.tt
@@ -131,6 +131,13 @@
                     </tbody>
                 </table>
             </div>
+            <div class="Hidden" id="DeleteMailAccountDialogContainer">
+                <div id ="DeleteMailAccountDialog" class="InnerContent GenericInterfaceDialog">
+                    <span class="WarningDialog ">&nbsp;</span>
+                    <p class="Center Warning">[% Translate("Do you really want to delete this mail account?") | html %]</p>
+                    <div class="SpacingTop"></div>
+                </div>
+            </div>
 [% RenderBlockEnd("OverviewResult") %]
 
 [% RenderBlockStart("OverviewAdd") %]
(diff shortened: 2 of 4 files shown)


zottto:
Hi Sanjin / @s7design,

having a confirmation before deleting something is very useful. IN this case I would prefer to also have the first dialog ("Do you really....") as a Core.UI.Dialog and not a browser-based confirmation (window.confirm).
There are several examples of delete confirmation dialogs with Core.UI.Dialog already in the core files.

Thanks,
Marc

s7design:
Hello @zottto ,

Thanks for your feedback. Removed browser-based confirmation and included dialog confirmation window. Please let me know if you would like to change some of the text or buttons.

Regards,
Sanjin

zottto:
Hi Sanjin / @s7design ,

two more adjustements:
- it is not necessary to open a second dialog, if the deletion is confirmed. You can dynamically remove the buttons and change the text in the dialog for the second message
- the text in the dialog can contain HTML, so it should be easy to make the message more pretty (padding e.g.) - please check other dialogs for reference

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for follow up. PR is updated and now has only one dialog which dynamically removes buttons and changes message on confirm click. Message is also styled now.

Button 'Confirm' is on left side and 'Cancel' is on right. To be honest there is mixed behaviour through out application in regards to button positioning. If you think we should switch order, it will be no problem to update this PR. We think this is more natural.

Regards,
Sanjin

zottto:
Hi Sanjin / @s7design,

- please use the text in black color, not orange (class Warning).
- please don't use the empty span in the dialog
- please add CallToAction design to the confirm button
- if I click on the delete icon, and click cancel in the dialog and after that click again on the delete icon and confirm the deletion, the entry is not deleted.

Thanks,
Marc

s7design:
Hello @zottto ,

Thank you for feedback. PR is updated:

- Removed orange color, text is default black now

- Removed empty span

- Added CallToAction on the confirm and removed Primary for cancel button. Two buttons have clear distinguish style now.

- This happens if you delete MailAccount right after you add one. After add screen URL just show baselink and hitting refresh button is adding another same MailAcount. I didn't want to change this in current PR, since there are couple of other screens that behave in same fashion. Executing Selenium test AdminMailAccount.t show you expected behaviour.

Please let me know if you have any additional feedback regarding this PR.

Regards,
Sanjin

s7design:
Hi @zottto 

Thanks Marc for follow-up it will be done soon. Sanjin will do that

Regards
Zoran

s7design:
Hello @zottto ,

PR is updated.

- Added classes to buttons.
- Used Core.App.InternalRedirect() function to avoid reloading page and causing duplicate creation of MailAccount after initial one on delete confirmation.

Please let me know if you have any additional feedback.

Regards,
Sanjin

zottto:
Merged with 07773dd.

Thanks,
Marc

PR #1426, opened 2016-08-29
Configurable shortcut icon

pboguslawski:
This mod introduces two new SysConfig parameters
- AgentShortcutIcon
- CustomerShortcutIcon

that allow one to specify shortcut icons to use in otrs agent
and customer panel.

Related: https://dev.ib.pl/ib/otrs/issues/85
Author-Change-Id: IB#1046319

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-08-29 Configurable shortcut icon.
  - 2016-06-07 Added index for searching dynamic field text values.
  - 2016-08-18 Added per-address email loop protection setting (PostmasterMaxEmailsPerAddress), thanks to Moritz Lenz.
  - 2016-08-12 Fixed bug#[12229](http://bugs.otrs.org/show_bug.cgi?id=12229) - Queue is not selectable if the name contains "<" or ">" characters.
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -281,6 +281,14 @@
             <String Regex="">Example Company</String>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="CustomerShortcutIcon" Required="0" Valid="0">
+        <Description Translatable="1">The shortcut icon for the customer interface.</Description>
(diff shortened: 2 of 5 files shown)


mburggraf:
Instead of configuring a fixed Shortcut-Icon, wouldn't it be more versatile to use make it relate to a selected skin? For example: The skin can be changed dynamically with Loader::Agent::DefaultSelectedSkin::HostBased or Loader::Customer::SelectedSkin::HostBased, while a fixed Icon is still fixed.

mrcbnsls:
Hi @pboguslawski and @mburggraf,

In general I think it's a nice idea to have this icon configurable, but I agree with @mburggraf that this configuration should be connected to the selected skin. I would suggest a configuration similar to the logo configuration (see AgentLogoCustom###default for reference). There could be default configuration which could then be overruled by skin configurations.

What do you think?

Marc

pboguslawski:
Hi,

OTRS allows different agents to choose skin they like most but favicon problably should stay the same in all skins (same like product name, application URL, company logo, etc.) so allowing different favicons is not optimal in my opinion.

Regards,
Pawel

mrcbnsls:
Well, with the solution I suggested it would still depend on the configuration if there are different icons for different skins. So users would not be able to change the icon by selecting a different skin if the admin doesn't allow it. Also, only this way it would be possible at all to use different icons for different skins (which makes sense if you're for example using host-based skins). So I still think this is the better way.

mrcbnsls:
Would you be willing to update this PR accordingly, @pboguslawski?

pboguslawski:
Hi,

In our opinion shortcut icon is not part of skin - its part of branding like logo and switching between skins should not change branding to keep things simple to maintain; two options introduced in our PR are enough for us. 

Feel free to extend our PR to your taste of course before merging.

Regards,
Pawel

mrcbnsls:
Hi Pawel,

Though I understand that it fits *your* needs, the way it is implemented right now just doesn't fit into the way how skins and things connected to them work in OTRS in *general*, as I already explained. Thanks a lot for your effort anyway!

Please feel free to re-submit a PR which works as expected anytime you wish.

Cheers,
Marc

PR #1432, opened 2016-08-31
 Master - JS refactoring - FooterJS

s7design:
Hello @zottto ,

Hereby is last refactored JS FooterJS.tt. First would just like to express gratitude for your cooperation and patience while working on this task. Personally, I learned and improved a lot during this journey and it gave me further insight in how OTRS handle tasks from a different perspective :)

Placed functions from FooterJS into Core.Agent.js. If you think some other JS is more suitable for carrying these events, I will update this PR accordingly.

Additionally, added two more things. First is 'New Process Ticket' from navigation menu is now also affected by TicketInNewWindow configuration. Second added classes in AgentDashboardCustomerUserList widget, so click on 'Create' for Phone and Email ticket will open in new window as well if configuration is enabled.

Please let me know if you agree with these changes or if there are any issues or questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -1506,10 +1506,11 @@ sub Footer {
     # get config object
     my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
-    # NewTicketInNewWindow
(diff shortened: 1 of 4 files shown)


zottto:
Merged with f998682.

Thanks,
Marc

PR #1440, opened 2016-09-01
Time Input Dropdown flexible configuration

skoszewski:
In the current OTRS framework Date / Time selection input control may be choosen from text boxes or dropdown control. The dropdown allows selection of any hour in the 0 - 11 range and any minute from 0 - 59. For most common uses there is no need to set the time to exact minutes. Many organizations may choose to limit choices to 4, 6 or 12 for example. This will significantly speed up chosing specific dates because the list will be much shorter.

The proposed modification adds SysConfig property which defines how many items will be in the dropdown by specifying time step in minutes. Possible values are divisions of 60:
- Every minute - values from 0 to 59 - this options is the current behavior
- Every second minute - values from 0 to 58 with step 2
- Every five minutes (I have choosen this as a reasonable default) - values from 0 to 55 with step 5
- Every ten minutes - values from 0 to 50 with step 10
- Every fiveteen minutes - values from 0 to 45 with step 15
- Every half an hour - values 0 and 30


mgruner:
Hello @skoszewski, looks nice, thank you! I don't believe that we need this also in the `Defaults.pm`. 
@mrcbnsls: could you review this please?

skoszewski:
It can be removed from the `Defaults.pm`, but then initialization code should be hardcoded into the `Layout.pm` to avoid `undef` values:

```perl
my $MinuteStep = $ConfigObject->Get('TimeInputMinutesStep') // 1;
```

skoszewski:
Guys, any chance for it to be reviewed and merged to the upstream. I was actually hoping the patch will make it to 5.0.14, so our Business Solution Instance could benefit from the change that many my users would welcome. I wanted to avoid making this as Custom Module.

frennkie:
I find the current dropdown containing 60 values from 00 to 59 to be very annoying.
I just merge these changes into my test system and think this is really helpful.

frennkie:
I still find this pretty useful... would be nice to see this in an upcoming release (5.0.19?!).

keram79:
+1 for including this in the next release!

skoszewski:
Hi @mrcbnsls and @mgruner !

It was a long time ago I posted this small feature request patch and I have a feeling it was forgotten and some people have found it helpful.

Any chances to review this patch?

NeverMin:
@skoszewski +1

mgruner:
@mrcbnsls I actually requested a review from you on 16 Sep 2016. From my side this is good to go with some added tests. Can you handle it please?

mgruner:
@skoszewski my apologies. We weren't able to spend sufficient time on PRs recently.

mrcbnsls:
Nice! Thanks a lot! I took the freedom to make some wording changes and to add a small extension to AgentTicketPending.t in order to check for the new feature. I now merged this in 245de825d5. Please understand though that we don't add such changes to patch level releases, which means that this will only be available in OTRS 6.

PR #1441, opened 2016-09-01
Master - JS and Selenium update - AdminGenericAgent

s7design:
Hello @zottto ,

Hereby is update for AdminGenricAgent JS function AddSelectClearButton(). This function was not working as it is suppose to. 

`// Only handle select fields with a size > 1, leave all single-dropdown fields untouched
  if (isNaN(Size) || Size <= 1) {
      return false;
 }`

This was causing .each() to leave event propagation on first field that satisfy condition. In this case it was validity field. Removed 'false', this now work as it is expected. Additionally included condition that this function only trigger if 'Modernize' fields are disabled. IMO there is no need for it with modernize fields, since there is option to deselect all values from the field.

Improved Selenium test to check for AddEvent(), ShowDuplicatedDialog(), ShowDeleteEventDialog() and AddSelectClearButton() JS functions. Additionally test was not clearing all dynamic fields that were created, fixed it as well.

Regards,
Sanjin

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
+++ b/var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
@@ -42,7 +42,7 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
 
             // Only handle select fields with a size > 1, leave all single-dropdown fields untouched
             if (isNaN(Size) || Size <= 1) {
-                return false;
+                return;
             }
 
             // If select field has a tree selection icon already,
@@ -82,25 +82,28 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
      */
     TargetNS.Init = function () {
 
-        $('.DeleteEvent').bind('click', function (Event) {
+        $('.DeleteEvent').on('click', function (Event) {
             TargetNS.ShowDeleteEventDialog(Event, $(this));
             return false;
         });
 
-        $('#AddEvent').bind('click', function (){
+        $('#AddEvent').on('click', function (){
             if ($('#EventType').val() !== null) {
                 TargetNS.AddEvent($('#EventType').val());
                 return;
             }
         });
 
-        $('#EventType').bind('change', function (){
+        $('#EventType').on('change', function (){
             TargetNS.ToggleEventSelect($(this).val());
         });
 
         Core.UI.Table.InitTableFilter($("#FilterGenericAgentJobs"), $("#GenericAgentJobs"));
 
-        AddSelectClearButton();
+        // Add select clear button if ModernizeFormFields is disabled
+        if (parseInt(Core.Config.Get('InputFieldsActivated'), 10) === 0) {
+            AddSelectClearButton();
+        }
     };
 
     /**
@@ -152,7 +155,7 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
         $Clone.find('.EventValue').attr('name', 'EventValues').val(EventName);
 
         // bind delete function
-        $Clone.find('#DeleteEvent').bind('click', function (Event) {
+        $Clone.find('#DeleteEvent').on('click', function (Event) {
             // remove row
             TargetNS.ShowDeleteEventDialog(Event, $(this));
             return false;
--- a/var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
+++ b/var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
@@ -36,13 +36,25 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
         // Loop over all select fields available on the page
         $SelectFields.each(function () {
             var Size = parseInt($(this).attr('size'), 10),
+                ModernizeField = $(this).hasClass('Modernize'),
+                ModernizeActive = parseInt(Core.Config.Get('InputFieldsActivated'), 10),
                 $SelectField = $(this),
                 SelectID = this.id,
                 ButtonHTML = '<a href="#" title="' + Core.Language.Translate('Remove selection') + '" class="GenericAgentClearSelect" data-select="' + SelectID + '"><span>' + Core.Language.Translate('Remove selection') + '</span><i class="fa fa-undo"></i></a>';
 
-            // Only handle select fields with a size > 1, leave all single-dropdown fields untouched
-            if (isNaN(Size) || Size <= 1) {
-                return;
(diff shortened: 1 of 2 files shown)


s7design:
Hello @zottto ,

Thanks for feedback. PR is updated.

Regards,
Sanjin

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
```diff
@@ -36,13 +36,25 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
         // Loop over all select fields available on the page
         $SelectFields.each(function () {
             var Size = parseInt($(this).attr('size'), 10),
+                ModernizeField = $(this).hasClass('Modernize'),
+                ModernizeActive = parseInt(Core.Config.Get('InputFieldsActivated'), 10),
```
zottto: This line can be moved outside the loop.

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
```diff
@@ -36,13 +36,25 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
...
                 $SelectField = $(this),
                 SelectID = this.id,
                 ButtonHTML = '<a href="#" title="' + Core.Language.Translate('Remove selection') + '" class="GenericAgentClearSelect" data-select="' + SelectID + '"><span>' + Core.Language.Translate('Remove selection') + '</span><i class="fa fa-undo"></i></a>';
 
-            // Only handle select fields with a size > 1, leave all single-dropdown fields untouched
-            if (isNaN(Size) || Size <= 1) {
-                return;
+            if (SelectID.indexOf('DynamicField') !== -1) {
```
zottto: Why is there a special handling for DynamicFields?

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
```diff
@@ -36,13 +36,25 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
...
+
+                // Only handle dynamic fields that are not modernized
+                if (ModernizeActive && ModernizeField) {
```
zottto: In my test case I have some date/time dynamic fields with one-size-dropdowns that now have a reset button, while other date/time select fields (not dynamic fields) do not have that.

s7design:
Hi @zottto 

Thanks, we will check everything that you mentioned in the comments above.

Regards
Zoran

s7design:
Hello @zottto ,

Thanks for feedback. PR is updated with different logic on add button for clear field. Since only dynamic fields which are multiselect and dropdown needs this functionality, replaced logic to check for there class 'DynamicFieldMultiSelectClass'. Now we should exclude dynamic fields like date, datetime, checkbox and other type's that don't require this button.

To answer your question why there is included special handling for dynamic fields, reason behind that is dynamic fields don't have attribute 'size' like other fields. Had to distinguish them in a way and treat them separately from remaining of fields. This could be approach from different side and add attribute size depending on possible option selection of dynamic field itself. This is much bigger approach and more complicated solution since attribute 'size' is hardcoded value for each field and doesn't represent actual number of options for appropriate field.

Please if you disagree with this logic let me know what ways could we implement to best reflect back to code functionality and user experience.

Regards,
Sanjin

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
s7design: Thanks for this feedback. PR is updated.

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
s7design: Please check comment below.

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
s7design: Please check updated PR and let me know if you still experience this issue.

var/httpd/htdocs/js/Core.Agent.Admin.GenericAgent.js
```diff
@@ -31,18 +31,31 @@ Core.Agent.Admin.GenericAgent = (function (TargetNS) {
...
+                    return;
+                }
+            }
+            else {
+
+                // Only handle select fields with a size > 1, leave all single-dropdown fields untouched
+                if ((ModernizeActive && ModernizeField) || isNaN(Size) || Size <= 1) {
+                    return;
+                }
```
zottto: In my opinion, you don't need the if-statement anymore to distinguish between DF and other fields.
The code in the else-part also handles DF correctly, because all DF that are not MultiSelects are already handled by this code.

s7design: Hello @zottto ,

Thank you for feedback, although I must disagree with you on this part. Problem is with DynamicField DropDown and MultiSelect fields. They both don't have 'Size' attribute. 

When 'If' statement is removed, those fields trigger 'isNaN(Size)' and return, resulting in no added GenericAgentClearSelect button. Perhaps this could be done a bit differently where we add to all fields properly 'Size' attribute that reflect actual size of selectable options, please check my comment on 26th September.

Reconsider this again please and let me know if you have some further questions.

Regards,
Sanjin

zottto: Hi @s7design / Sanjin,
I now understand the problem behind the code, but I don't like the code itself. :-)

The code is not really readable so I suggest to improve this part a little bit.

In the first step you can check for ModernizeActive && ModernizeField and return (for all fields).
In the second step you can check for (!DynamicFieldMultiSelectClass && (isNaN(Size) || Size <= 1)) and return.

This makes the code better readable and I think that is does the same. ;-)

Please check that again.

Thanks,
Marc

s7design: Hello @zottto ,

Yes, I agree with you that code isn't pretty. Will check it ASAP and try to come with nicer solution :) 

Thanks for feedback, I am glad we can come together to the acceptable solution thats best for OTRS :)

Regards,
Sanjin

s7design:
Hello @zottto ,

Thank you once again on latest feedback. Done some testing and your suggestion does work and looks nicer.

Please let me know if there are any other issues regarding this PR.

Regards,
Sanjin

zottto:
Merged with some changes with 3728da7.

PR #1334, opened 2016-07-05
Followup pull request 929: Added new functionality to ticket search to search for empty dynamic fields

rolfschmidt:
Hi @mgruner 

yesterday i had some problems with my old branch and rage quitted it...
Tonight i tried to get it working again and i would like to continue my pull request:

https://github.com/OTRS/otrs/pull/929

- ConnectByOr functionality removed
- Added unit tests for DynamicField/SearchSQLGet()
- The POD of Operation/Ticket/TicketSearch.pm added
- The POD of development/webservices/GenericTicketConnectorSOAP.wsdl added

Best regards,
Rolf

PR #1444, opened 2016-09-02
RFC: Allow ticket search for presence or absence of dynamic field values.

moritz:
I've only implemented it for text and datetime fields so far; if feedback for this is positive, I can add support for other dynamic field types too.


thorsteneckel:
Hi @moritz - we've implemented something similar which is already in an advanced state of review: https://github.com/OTRS/otrs/pull/1334

mgruner:
Hello @moritz. Thanks for your suggestion! I also believe that merging this with the mentioned PR would be a good idea. I'll close this entry for now.

PR #1448, opened 2016-09-05
[rel-5_0] Added confirmation for delete buttons

urbalazs:
Hi @zottto,
I added confirmations for delete buttons, because now these buttons delete items immediately. If a user clicks them accidentally, there is no way to cancel and undo.

@s7design, you are working on JS refactoring in master. Can you also add these confirmations? My solution uses the standard JavaScript confirm dialog, but OTRS has own popup dialog for confirmation (like for deleting a web service).

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminAttachment.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminAttachment.tt
@@ -79,7 +79,7 @@
                             <td>[% Data.ChangeTime | Localize("TimeShort") %]</td>
                             <td>[% Data.CreateTime | Localize("TimeShort") %]</td>
                             <td class="Center">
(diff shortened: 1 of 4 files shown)


s7design:
Hello @urbalazs ,

Thanks for bringing this up. We created PR on this topic https://github.com/OTRS/otrs/pull/1424 and it's in review process. Used Core.UI.Dialog.js OTRS functions to activate dialog on delete action, like how it used in web service, which you mentioned. 

Once we have final solution, will create PR for other screens that have delete actions.

Regards,
Sanjin 
s7design

urbalazs:
Hi @zottto 
The https://github.com/OTRS/otrs/pull/1424 is for branch master, but my patch is for branch rel-5_0. The JavaScript refactorization is only for master, so I can not use the new modal dialogs here. There are many browser confirmation is OTRS 5, only these are missing.

zottto:
Merged with c5e9c6f.
I also updates the selenium tests.

Thanks,
Marc

PR #1454, opened 2016-09-07
Master frontend usability - AdminRoleGroup

s7design:
Hi @zottto 

Hereby is refactored AdminRoleGroup. 
Blocks 'Overview' and 'Change' are separated and they are created for different screens which means that BreadcrumbPath has to be define for both blocks. Because of that, AdminRoleGroup TT is modified in order to BreadcrumbPath is used only once.
'Save and Finish' is added on the standard way and Selenium test is modified accordingly.
Please let me know if you have any comment for this PR.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminRoleGroup.pm
+++ b/Kernel/Modules/AdminRoleGroup.pm
@@ -138,7 +138,15 @@ sub Run {
                 UserID     => $Self->{UserID},
             );
         }
-        return $LayoutObject->Redirect( OP => "Action=$Self->{Action}" );
+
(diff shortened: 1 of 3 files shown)


s7design:
Hi @zottto

PR is updated. Notifications are added for Role-Group relations update and for add group because after group adding in AdminGroup module there is redirection to AdminRoleGroup screen.

Regards
Zoran

s7design:
Hi @zottto

The PR is updated, the notifications are deleted.

Regards
Zoran

zottto:
Merged with 0127a57.

Thanks,
Marc

PR #1455, opened 2016-09-07
Master - bug - AdminSession and logging ref() in concat

s7design:
Hello @mgruner ,

While working on breadcrumb task for AdminSession found two small bugs. 

Firstly, when you enter to see session for customer, you get this logging error:

screen shot 2016-09-07 at 11 23 37 am

Tracing steps back found problem on line 101 in AdminSession.pm :

`$Data{$Key} = $LayoutObject->Ascii2Html( Text => $Data{$Key} );`

Specifically only for Config and CompanyConfig keys. These values are hashes and Ascii2Html() function is set to return empty string in such case. Now this had no influence on the remaining of module functionality when on line 115 :

`if ( $Key eq 'Config' || $Key eq 'CompanyConfig' ) {
                    $Data{$Key} = 'HASH of data';
                }`

Data of those keys is set to be 'HASH of data' and this is result we see in detail view of session. Moving this in "elsif" statement prevents Ascii2Html() to handle those keys.

Secondly as you can see from Fred log, log message is not complete. This was caused by ref() in concatenation with other string :

`'Invalid ref "' . ref $Param{Text} . '" of Text param!'`

I searched a bit for the explanation and cause of this issue to no avail. My best guess would be that ref() function is not stopping on ' . ' concat and is taking the rest of the string as param. When expression is placed inside braces problem is solved and log shows whole message.

Please enlighten me regarding second point and let me know if you have any questions.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AdminSession.pm
+++ b/Kernel/Modules/AdminSession.pm
@@ -97,6 +97,9 @@ sub Run {
                 if ( $Key =~ /Password|Pw/ ) {
                     $Data{$Key} = 'xxxxxxxx';
                 }
+                elsif ( $Key eq 'Config' || $Key eq 'CompanyConfig' ) {
+                    $Data{$Key} = 'HASH of data';
+                }
                 else {
                     $Data{$Key} = $LayoutObject->Ascii2Html( Text => $Data{$Key} );
                 }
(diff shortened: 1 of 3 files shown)


mgruner:
Merged with slight changes in 600b3897c622d8058e41311b9873d2c17dfd308d, thanks @s7design!

PR #1458, opened 2016-09-07
Errors on entering AdminCustomerUser with RunInitialWildcardSearch disabled

pboguslawski:
When AdminCustomerUser::RunInitialWildcardSearch is disabled (i.e. for
performance reasons) and agent opens AdminCustomerUser (which opens with
empty list) errors like

    Message: Need Search, UserLogin, PostMasterSearch, CustomerIDRaw or CustomerID!

occur in apache error.log (for DB and similar one for LDAP backend if present).

This mod resolves issue as above.

Related: https://dev.ib.pl/ib/otrs/issues/88
Author-Change-Id: IB#1055644


mgruner:
Applied and backported, thanks a lot @pboguslawski!
http://bugs.otrs.org/show_bug.cgi?id=12297

PR #1468, opened 2016-09-13
Optimized user and customer lists generation

pboguslawski:
When searching for users in AdminUser, customer users in AdminCustomerUser
and customer companies in AdminCustomerUser OTRS executes two queries
against all backends: one with limit and second with limit + 1 just
to decide if "more available" should be displayed in list header.

This mod changes this behaviour to execute only one query against each
backend (with limit+1) and trim results to limit which avoids redundant
backend (i.e. DB) queries.

This mod also changes way how limit passed to function call is compared
against limits defined in backend configuration - now the lesser limit
wins which is safer (admin when configuring limit in backend hopes
it won't be exceed; the only exception is Limit = 0 which works
without changes and searches without limits).

This mod also limits number of items returned in functions above to 400
(hardcoded) even if many backends are present (without this mod OTRS
limits display to 400 * # of backends which may produce large web pages
in case of many backends with lots of items).

Related: https://dev.ib.pl/ib/otrs/issues/89
Related: http://bugs.otrs.org/show_bug.cgi?id=7708
Author-Change-Id: IB#1055556

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-09-13 Optimized user and customer lists generation.
  - 2016-09-12 Fixed bug#[11365](http://bugs.otrs.org/show_bug.cgi?id=11365) - ACLs editor shows actions where ACLs does not apply.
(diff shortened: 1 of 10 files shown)


mgruner:
Dear contributor,

PR #1471, opened 2016-09-14
Fixed error Argument "" isn't numeric in bitwise xor (^) at [...]MatchDBSource.pm

pboguslawski:
When there is postmaster filter defined in OTRS with negated EMAILADDRESS:test@test.com
condition against From header, postmaster throws an error like

    Argument "" isn't numeric in bitwise xor (^) at /opt/otrs/Kernel/System/PostMaster/Filter/MatchDBSource.pm [...]

on every e-mail message import.

This mod fixes this issue.

https://dev.ib.pl/ib/otrs/issues/91
Author-Change-Id: IB#1054236


mgruner:
Merged and backported to 5 and 4, thanks @pboguslawski!

PR #1476, opened 2016-09-19
Master - Fixed code if RichText param is not defined

s7design:
Hi @mgruner 

There is small improvement IMHO. I saw that there is some message in Fred if RichTextWidth and RichTextWidth are not passed in SetRichTextParameters function.
There is our proposal for fixing, if you like that you can merge it in master.

This is log from Fred:
```
[Mon Sep 19 12:58:18 2016] -e: Use of uninitialized value in numeric gt (>) at /opt/otrs//Kernel/Output/HTML/Layout.pm line 5583.
[Mon Sep 19 12:58:18 2016] -e: Use of uninitialized value in numeric gt (>) at /opt/otrs//Kernel/Output/HTML/Layout.pm line 5591.
```

You can reproduce this for example if you edit signature in AdminSignature and similar ones screens.

Regards
Zoran

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -5576,25 +5576,8 @@ sub SetRichTextParameters {
     my $ConfigObject   = $Kernel::OM->Get('Kernel::Config');
 
     # get needed variables
-    my $ScreenRichTextHeight;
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Merged in 5174e62, thanks!

PR #1477, opened 2016-09-19
Master - Fixed spam log error in AdminSysConfig

s7design:
Hi @mgruner 

Working on frontend usability I saw strange log error. IMHO it is spam error message with small changes from this PR it is fixed.

P.S.
I am not sure if you want that we fix such things. It does not a problem for us to fix that. It takes a little our time but I think it is better if there are not such messages.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AdminSysConfig.pm
+++ b/Kernel/Modules/AdminSysConfig.pm
@@ -1005,7 +1005,7 @@ sub Run {
 
         $LayoutObject->Block( Name => 'OverviewResult' );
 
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @s7design,

In general, there should be a bug report for such things in bugzilla. Currently I don't even know what exactly this PR is fixing. So I'd ask you to create a bug entry with more information and re-create the PR afterwards with a reference to the bug entry. Thanks!

Marc

s7design:
Hi @mrcbnsls 

No problem I will apen the bug and then fix it.

Regards
Zoran

PR #1478, opened 2016-09-20
Master frontend usability - AdminRegistration

s7design:
Hi @zottto 

Hereby is refactored AdminRegistration. Breadcrumb is added. When system is not registered, in Overview screen controller sets up subaction to 'OTRSIDValidate' (parameter for breadcrumb in TT), but our proposal is that breadcrumb 'OTRSIDValidate' might not be drawn in this case.
Selenium test is modified.

Regards
Zoran


mrcbnsls:
Merged in d83b021, thanks!

PR #1479, opened 2016-09-20
Marked process states as translatable

urbalazs:
Hi @mgruner,
This PR marks process states as translatable. Branch rel-5_0 is also affected.


mrcbnsls:
Thanks, merged in 0fb55dd and 08e3c21.

PR #1480, opened 2016-09-21
Fix SysConfig Selenium test

s7design:
Hi @mrcbnsls 

We found that Selenium test has an error and fixed that with adding 'a' tag. Please let me know what do you think about this.

Regards
Zoran


mrcbnsls:
Merged in 7c8c7b4, thanks!

PR #1485, opened 2016-09-22
Backup script: Fixed check for article directory

jepf:
Checking if the article directory is separate from the home directory of OTRS to decide if article data has to be backed up is broken. If Home is set to /opt/otrs and ArticleDir is set to /opt/otrs_data, /opt/otrs_data won't be backed up.

Please have a look at the fix.


balazj:
Hi Jens,

Thank you very much for this PR, this was a very special use case. I merged it with master, rel-5_0 and rel-4_0 branches.

Kind Regards,
Jaroslav

PR #1490, opened 2016-09-23
Update nb_NO.pm

thoughtbox:
Wrong translation of "My locked tickets" - should be the same as "My Locked Tickets" (see earlier in the file).


mgruner:
Hello @thoughtbox. Please submit any updates via transifex (https://www.transifex.com/otrs/OTRS) only (you can get a free account).

PR #1495, opened 2016-09-26
Fixed SchedulerTaskWorker::AsynchronousExecutor POD

thorsteneckel:


mgruner:
Applied, thanks a lot @thorsteneckel. You might have seen that we worked on the POD last week. Manuals and API docs now also have syntax highlighting (see http://otrs.github.io/doc/api/otrs/6.0/Perl/Kernel/System/ObjectManager.pm.html). We still have to work on most of the content, though. Any help will be appreciated.

thorsteneckel:
Thanks for sharing @mgruner ! That looks promising to me

PR #731, opened 2015-09-28
[Rel-5_0 10117 ] - Filter in queue view for customer userID shows only customers from given backend

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10117

Hi @mgruner 

We made again PR for the bug #10117. Now, we provided solution for  branch rel-5_0.  As you can see, we added new SysCongig item, which enable this type of search by unknow customers or doesn't.

There are typos in:
<pre> 
/opt/otrs/var/httpd/htdocs/js/Core.Agent.Dashboard.js:
   43                      Subaction: 'SearchCustomerID',
   44                      Term: Request.term,
   45:                     MaxResults: Core.Config.Get('CustomerAutocomplete.MaxResultsDisplayed')
   46                  };
   47  

/opt/otrs/var/httpd/htdocs/js/Core.Agent.TableFilters.js:
   50                      Subaction: 'SearchCustomerID',
   51                      Term: Request.term,
   52:                     MaxResults: Core.Config.Get('CustomerAutocomplete.MaxResultsDisplayed')
   53                  };
</pre>
because these param are set here:

<pre>
/opt/otrs/Kernel/Output/HTML/Templates/Standard/AgentDashboardTicketGeneric.tt:

  141: Core.Config.Set('CustomerIDAutocomplete.QueryDelay', "[% Data.queryDelay | html %]");
  142  Core.Config.Set('CustomerIDAutocomplete.MaxResultsDisplayed', "[% Data.maxResultsDisplayed | html %]");
  143  Core.Config.Set('CustomerIDAutocomplete.MinQueryLength', "[% Data.minQueryLength | html %]");

/opt/otrs/Kernel/Output/HTML/Templates/Standard/AgentTicketOverviewSmall.tt:

  147: Core.Config.Set('CustomerIDAutocomplete.QueryDelay', "[% Data.queryDelay | html %]");
  148  Core.Config.Set('CustomerIDAutocomplete.MaxResultsDisplayed', "[% Data.maxResultsDisplayed | html %]");

</pre>

There should be CustomerIDAutocomplete instead of CustomerAutocomplete. We fixed it also.

Please let me know if you have any suggestion.

Regards
Zoran

PR #1496, opened 2016-09-27
[Rel-4_0 - bug 10117 ] - Filter for customer user ID shows only customers from given backend

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10117

Hi @mgruner 

We made PR for bug #731  for rel-4_0 as Ulrike Schneider suggested. New SysConfig item is added and could be enabled or disabled for searching unknown customer users.

Regards
Zoran


dvuckovic:
Merged in c92d47e, with some small changes. Thanks @s7design!

PR #1497, opened 2016-09-27
Update TicketInformation.tt

nilei:
Pending until time always belongs to the ticket state and should be displayed next to it.


mrcbnsls:
Merged in dad059a, thanks.

PR #1500, opened 2016-09-27
Added sort criteria to TicketSearch call

capeit-tto:
If more than 10k tickets are in pending states, it might happen that due to missing sort criteria and -order, tickets which should be updated will not experience this update because they're returned at last. The first 10k resulting tickets might be in pending auto states, but their until time is in the future.


dvuckovic:
Merged in 96f49ef, thanks @capeit-tto!

PR #1501, opened 2016-09-27
Blocking external content using CSP

pboguslawski:
OTRS tries to filter external content in HTML e-mails using perl
regexps; this way its difficult to predict and block all HTML/JS/CSS
quirks, i.e. OTRS does not block remote image loading if used
in CSS like this

```
<div style="background:url('http://badspammer.com/tracking/1/TrackRead.aspx?mail_id=1F4CA3DD-24E0-4C14-B631-A7567EAA5D83')"></div>
```

which may allow spammers to notice that their junk was read by receiver.

Using `Content Security Policy` described on
http://www.html5rocks.com/en/tutorials/security/content-security-policy/
seems to be proper way of blocking external content in iframes.
According to https://content-security-policy.com/ it should work fine
in modern browsers (in IE from v12 probably).

This mod adds

```
Content-Security-Policy: default-src 'self'; style-src 'unsafe-inline'
```

HTTP header for iframes with HTML e-mails which forces modern web browsers
to block all external and inline content except inline styling (used often
in html e-mails). CSP header is not added if user explicitly unlocks
external content using link provided by OTRS.

This mod renames also `LoadExternalImages` parameters to `LoadExternalContent`
to better match its purpose and message displayed by OTRS.

In future it may be good idea to drop OTRS own filtering to offload this task
from busy application servers (parsing huge HTML content may cause problems...)
to web browsers.

Related: https://dev.ib.pl/ib/otrs/issues/92
Related: http://www.html5rocks.com/en/tutorials/security/content-security-policy/
Author-Change-Id: IB#1047017

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -2466,10 +2466,12 @@ sub ReturnValue {
 returns browser output to display/download a attachment
 
     $HTML = $LayoutObject->Attachment(
-        Type        => 'inline',        # optional, default: attachment, possible: inline|attachment
-        Filename    => 'FileName.png',  # optional
-        ContentType => 'image/png',
-        Content     => $Content,
+        Type                => 'inline',        # optional, default: attachment, possible: inline|attachment
+        Filename            => 'FileName.png',  # optional
+        ContentType         => 'image/png',
+        Content             => $Content,
+        LoadExternalContent => 1,               # optional, 0 - don't allow page to load resources from external sites;
+                                                # 1 (default) - allow allow page to load resources from external sites
     );
 
     or for AJAX html snippets
@@ -2535,6 +2537,15 @@ sub Attachment {
 
     if ( !$ConfigObject->Get('DisableIFrameOriginRestricted') ) {
         $Output .= "X-Frame-Options: SAMEORIGIN\n";
+    }
+
+    # use CSP to disallow external content to be loaded if not explicitly enabled;
+    # don't allow inline scripts but allow inline styles for messages and
+    # blocking info box to look nice; for CSP details see
+    # http://www.html5rocks.com/en/tutorials/security/content-security-policy/
+    # https://www.w3.org/TR/CSP2/#directive-default-src
+    if ( defined $Param{LoadExternalContent} && ( $Param{LoadExternalContent} == 0 ) ) {
+        $Output .= "Content-Security-Policy: default-src 'self'; style-src 'unsafe-inline'\n"; 
     }
 
     if ( $Param{Charset} ) {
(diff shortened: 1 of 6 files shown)


Kernel/Output/HTML/Layout.pm
```diff
@@ -2537,6 +2539,15 @@ sub Attachment {
         $Output .= "X-Frame-Options: SAMEORIGIN\n";
     }
 
+    # use CSP to disallow external content to be loaded if not explicitly enabled;
+    # don't allow inline scripts but allow inline styles for messages and
+    # blocking info box to look nice; for CSP details see
+    # http://www.html5rocks.com/en/tutorials/security/content-security-policy/
+    # https://www.w3.org/TR/CSP2/#directive-default-src
+    if ( defined $Param{LoadExternalContent} && ( $Param{LoadExternalContent} == 0 ) ) {
```
mgruner: This line will never be effective, because LoadExternalContent is not explicitly set to 0 anywhere. So I believe the patch will not work as it stands, correct?

pboguslawski:
Hi,

We've already merged your changes

https://dev.ib.pl/ib/otrs/blob/master-ib/Kernel/Output/HTML/Layout.pm#L2539

but with fixed CSP, because your policy

https://github.com/OTRS/otrs/blob/master/Kernel/Output/HTML/Layout.pm#L2554

still allows external images to be loaded i.e. using background tricks in CSS like in original commit comment above.

Please verify and fix and close this PR because it does not match current master.

Regards,
Pawel

mgruner:
Hello @pboguslawski, could you please rebase this PR or open another one that works with the current state in master. Then I'll review it again. Thanks and best regards, Martin

pboguslawski:
Hi,

According to [The Golden Rule of Rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/) we don't rebase in public repos; preparing new commit/PR costs time, please consider fixing it manually - the only difference is in CSP definition; yours is

    Content-Security-Policy: default-src *; img-src * data:; script-src 'none'; object-src 'none'; frame-src 'none'; style-src 'unsafe-inline'; referrer no-referrer;\n

which does not block external images (img-src *); it also contains obsolete frame-src which should probably be changed to  child-src; we use

    Content-Security-Policy: default-src 'self'; img-src 'self' data:; script-src 'none'; object-src 'none'; child-src 'none'; style-src 'unsafe-inline'; referrer no-referrer;\n

and it seems to work ok.

Regards,
Pawel

mgruner:
Hello @pboguslawski. This PR is not in a mergeable state, so I'll close it for now. Please feel free to submit another one. I believe that LoadExternalImages could stay, as this is what is really changed by the PR. It should be solved in a way that the existing CSP rule stays active, and another one just blocking images is added. The most restrictive one will win.

PR #1506, opened 2016-09-28
Master - AgentTicketProcess - Customer autocomplete fix

s7design:
Hello @zottto ,

Hereby is fix for customer autocomplete field in AgentTicketProcess. There has been error from our side while JS refactoring. Events were loaded on initial page load but not when process is selected which resulted in no events for customer autocomplete field. 

Regards,
Sanjin


zottto:
Merged with be2a5da.

PR #1507, opened 2016-09-28
Column list validation before saving agent preferences

pboguslawski:
It was noticed in OTRS 3.3, that some agents managed to save their column
filter preferences with duplikcated column names i.e.

```
mysql -e 'SELECT * FROM user_preferences where length(preferences_value)>100' otrs
[...]
5       UserFilterColumnsEnabled-AgentTicketQueue       ["TicketNumber","Title","State","Lock","Queue","Owner","Age","Changed","CustomerID","Lock","Queue","CustomerID","TicketNumber", [...], "TicketNumber","Age","Title","State","Lock","Queue","CustomerID"]
```

The whole preference string was over 9kB long (all filled with duplicated
column list as above) and caused apache RAM consumption problems (removing
this record from DB was necessary to avoid out of memory on OTRS
application server). The problem is not reproducible and the cause of
the problem has not been identified (might be some kind of race condition
or web browser bug).

Regardless of cause OTRS should not allow to save duplicated columns
(i.e. to prevent attacks).

This mod adds validation of column list just before saving preferences
- duplicates are removed which should protect OTRS against the problem
  described above.

Related: https://dev.ib.pl/ib/otrs/issues/94
Author-Change-Id: IB#1058144

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-09-28 Column list validation before saving agent preferences.
  - 2016-09-21 Fixed bug#[12065](http://bugs.otrs.org/show_bug.cgi?id=12065) - queue and state not mandatory.
  - 2016-09-12 Fixed bug#[11365](http://bugs.otrs.org/show_bug.cgi?id=11365) - ACLs editor shows actions where ACLs does not apply.
  - 2016-09-08 Made possible to define ServiceIDs and SLAIDs as default shown ticket search attributes, thanks to Paweł Bogusławski.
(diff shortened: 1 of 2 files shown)


mgruner:
@mrcbnsls IIRC you already worked on this in the past, didn't you? Can you please review this?

mrcbnsls:
Hi @pboguslawski,

When I was looking into this PR, it turned out that this module is no longer in use and I removed it in 7c5daca. What you're trying to achieve is actually already done in Dashboard/TicketGeneric.pm (line 339@master).

I'll close this PR.

Marc

mrcbnsls:
Well, actually there seem to be dependencies to the configuration. I reverted the commit (7cc39e3). I will have another look...

PR #1508, opened 2016-09-29
Rel-5_0 - Selenium - AgentTicketZoomArticleFilter - update

s7design:
Hello @mrcbnsls ,

Saw on public test summary that this test is failing on few different system environments. It must be that somehow filter was not triggered or with incorrect paramaters. Included some additional wait functions to give more time to open dialog.

Hope this will fix those issues.

Regards,
Sanjin


zottto:
Hi @s7design,

at the moment all unit tests look fine, so I think we don't need that adjustement here.

Thanks,
Marc

PR #1509, opened 2016-09-29
[ Master Bug 10691 ] - No CustomerID shown after TicketCreate

s7design:
http://bugs.otrs.org/show_bug.cgi?id=10691

Hi @mgruner 

There is proposal how we can fix the issue that reporter mentioned in the bug report. This bug was closed and we commented as such behavior is expected. 
What do you think about this solution. I only provide it as a proposal, but honestly I am not sure if we should change anything there.
In any case we can speak about new solution or to close the bug with some comment.

Regards
Zoran


mgruner:
We must solve this in another way, please see my comment on the bug. Thanks Zoran!

PR #1510, opened 2016-09-29
Fixed escaping values passed to LDAP filters

pboguslawski:
OTRS does not escape all special characters in values passed to LDAP
filters. For example, when Active Directory is configured as customer
backend, searching for customer user "()" throws "bad filter"
LDAP error.

Passing unescaped strings may be dangerous (LDAP injections).

This mod introduces escaping values passed to LDAP filters by OTRS
with escape_filter_value from Net::LDAP::Util. Asterisk
in customer searches stays unescaped (acts as wildcard).

Related: https://dev.ib.pl/ib/otrs/issues/95
Related: https://www.praetorian.com/blog/how-to-identify-and-prevent-ldap-injection-part-2
Author-Change-Id: IB#1058224

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-09-30 Fixed escaping values passed to LDAP filters.
  - 2016-09-21 Fixed bug#[12065](http://bugs.otrs.org/show_bug.cgi?id=12065) - queue and state not mandatory.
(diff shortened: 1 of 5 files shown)


mgruner:
http://bugs.otrs.org/show_bug.cgi?id=12316

dvuckovic:
Merged in d6f1b0b, cbabe783 and a017c1d. Thanks @pboguslawski!

PR #1512, opened 2016-09-30
Removed the default group 'users' from QueuePermissionGroup setting, …

MBRG:
…as this can lead to confusion as to why "new queues are not visible in the Queue Overview Dashlet right away".


mgruner:
Merged for master only in 6a17b76fe2e1bb6636123f77f1041fffc23baa45, thanks a lot  @MBRG!

PR #1514, opened 2016-10-02
Add a missing word at the end of a sentence

issyl0:
- While translating strings for OTRS on Transifex, I came across a
  couple of source strings that had typos or missing words. This is
  one I've managed to find in the code to fix. I hope I've fixed it in
  the right places - I have no idea about Perl, I just figured that I
  should fix it in the source rather than every one of the i18n files
  in this repo.


mgruner:
Hello @issyl0. Thanks for your suggestion. After internal review it was decided that the wording should stay as it is in this particular case. Sorry!

PR #1520, opened 2016-10-10
Disabled warnings on internal unicode chacter printing

pboguslawski:
OTRS throws warning

```
Unicode non-character U+FFFF is illegal for open interchange at [...]Kernel/Output/HTML/Layout.pm [...]
```

to apache error log on displaying article containing unicode 0xFFFF
character (i.e. content comming from UTF-8 e-mail with character 0xFFFF).

Problem with such warnings in perl was discussed on
https://rt.perl.org/Public/Bug/Display.html?id=121226

OTRS should not pollute apache logs with such warnings if allows
such private chars to leak from e-mails to web browser.

This mod disables such warnings in OTRS on printing articles
to web browser.

Related: https://dev.ib.pl/ib/otrs/issues/97
Related: https://rt.perl.org/Public/Bug/Display.html?id=121226
Author-Change-Id: IB#1036717

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-10-10 Disabled warnings on internal unicode chacter printing.
  - 2016-09-21 Fixed bug#[12065](http://bugs.otrs.org/show_bug.cgi?id=12065) - queue and state not mandatory.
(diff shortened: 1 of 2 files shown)


mgruner:
Applied and backported to OTRS 5 and 4. Thanks!

PR #1524, opened 2016-10-13
Configurable resolver timeout

pboguslawski:
OTRS uses hardcoded 3s timeout for DNS queries while checking MX
(if CheckMXRecord is enabled). 3s may be too short in some scenarios
(sporadic errors in MX validation of remote addresses). DNS timeout
should be configurable to allow admin to fine-tune this setting.

This mod introduces new SysConfig parameter ResolverTimeout that
allows one to set custom timeout for DNS queries fired by OTRS
(now only for CheckMXRecord algo).

Related: https://dev.ib.pl/ib/otrs/issues/99
Author-Change-Id: IB#1058919

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-10-13 Configurable resolver timeout.
  - 2016-09-21 Fixed bug#[12065](http://bugs.otrs.org/show_bug.cgi?id=12065) - queue and state not mandatory.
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Pawel, this has been in OTRS for a very long time, and so far I was not aware of problems with that. I think that such an issue is rather a problem with the IT infrastructure and should be solved at this level.
We have to have a balance between flexibility in configuration and simplicity. In this particular case I'd like to avoid introducing yet another config setting.
Thanks a lot for your suggestion!
Best regards, Martin

PR #1525, opened 2016-10-15
Wrong command in suggestion

urbalazs:
Hi @mgruner 
There is a wrong command in a suggestion in the Support Data Collector. Branch rel-5_0 is also affected.


mgruner:

PR #1527, opened 2016-10-17
Master - fixed modernize checkbox DF in search screens

s7design:
Hi @mgruner, @mrcbnsls 

We found that there is small inconsequence in Search screens. Checkbox DF is not modernize field. There is fix, if you like you can marge in master. 
I saw this when I work on TME and HSDF AddOns.

Regards
Zoran


dvuckovic:
Merged in 8bacfe6 and 2550bcb, thanks @s7design!

PR #1528, opened 2016-10-19
[Rel-5_0 Bug 12367] - * in CustomerID breaks all CustomerUser.CustomerIDs

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12367

Hello,

Hereby is our proposal for fixing reporters issue. There was a problem with CustomerSearch() function after updating CustomerCompany. All available CustomerUsers were found when using asterisk symbol as search parameter to see which CustomerUsers needs to be updated to reflect CustomerCompanyID update.

Added parameter which is used in CustomerSearch() in DB.pm so we don't treat asterisk symbol as '%' in SQL bind. Updated UT to check for this type of behaviour.

Please let me know if you have any issues or questions regarding this fix.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/CustomerCompany.pm
+++ b/Kernel/System/CustomerCompany.pm
@@ -120,6 +120,15 @@ sub CustomerCompanyAdd {
         }
     }
 
+    # check if special characters '*' or '%' are used alone in CustomerID param
+    if ( $Param{CustomerID} =~ /^[%*]+$/ ) {
+        $Kernel::OM->Get('Kernel::System::Log')->Log(
+            Priority => 'error',
+            Message  => "Invalid CustomerID $Param{CustomerID} !",
+        );
+        return;
+    }
+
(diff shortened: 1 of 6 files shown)


s7design:
Hello,

Changed approach of resolving reporters issue in latest commit so we forbid input of special characters '*' and '%' as CustomerID when used alone or in combination with one another without other characters. Fix is applied for CustomerCompanyAdd() and CustomerCompanyUpdate().

Modified UT and Selenium tests to check for these restrictions.

Please let me know if you have any questions or issues regarding this bug fix approach.

Regards,
Sanjin

mgruner:
Hello @s7design. I don't agree with this proposal as it only works around the problem instead of solving it by using the CustomerIDRaw parameter that CustomerSearch offers. I applied this fix now, please see the bug for details.

PR #1529, opened 2016-10-19
Fixed use of uninitialized value $Param{"To"} in string ne

pboguslawski:
This mod fixes use of uninitialized value $Param{"To"} in string ne
in Kernel/Modules/AgentTicketCompose.pm.

Related: https://dev.ib.pl/ib/otrs/issues/100
Author-Change-Id: IB#1024517


mgruner:
I added a `defined` and applied and backported this, thanks @pboguslawski!

PR #1530, opened 2016-10-19
[Rel-5_0 Bug 12369 ] - Errors in deleting tickets

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12369

Hi Dennis,

There is fix for the problem that a reporter mentioned in the bug report. It was our proposal. There was a case that if ticket is deleted there are a lot log errors. It could be solved if we add Silent param in TicketGet function in the event TicketDynamicFieldDefault.

I felt free also to improve the code. Testing this case I saw that this event is called more times. I used idea from the event TicketProcessTransitions.pm.

Please let me know what do you think about this solution.

Regards
Zoran


dvuckovic:
Merged in 23a965d, 20282a4 and 7fcfdc9, with an accompanying unit test. Thanks Zoran!

PR #1532, opened 2016-10-24
Fixed e-mail address validation while sending autoresponses and article creation on send errors

pboguslawski:
OTRS does not check if e-mail specified in From/Reply-To message headers
is valid before using it for sending autoresponse which throws sending
errors in case of invalid addresses like

```
test <test@>
```

present in From or Reply-To.

In case of e-mail (i.e. autoresponse, forwarding, reply) sending error
(i.e. bad address like above or other SMTP error like too big message)
ArticleSend() creates article in DB which is misleading because
no e-mail message was sent in fact. ArticleSend() should not create
article in DB if no message was successfully sent to allow agent
to clearly notice the problem.

This mod fixes issues described above.

Related: https://dev.ib.pl/ib/otrs/issues/101
Author-Change-Id: IB#1023814

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-11-08 Fixed quotation in log message.
  - 2016-10-24 Fixed e-mail address validation while sending autoresponses and article creation on send errors.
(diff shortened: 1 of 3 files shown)


mgruner:
Hello Pawel. This seems to be related to https://bugs.otrs.org/show_bug.cgi?id=12407, I'll add a note there as well. I don't want to merge the PR in its current form as it changes the way OTRS creates tickets: it first sends the mail and then creates the article.

PR #1534, opened 2016-10-25
[ Rel-5_0 Bug 12371 ] - StorageModule ArticleStorageFS fails on attachments with long filenames

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12371

Hello,

Hereby is our proposal for fixing reporters issue. There was an issue when filename is ~250 character long including extension having each character 1 byte in size. Almost all systems allow up to 255 bytes per Filename when trying to save them in system as ArticleStorageFS is doing.

Problem is that ArticleStorageFS is creating additional files of same attachment in system with added extension e.g. ( .content-type , .disposition, .content-alternative). With already reaching maximum Filename length, byte wise, with these extensions we excess limit and attachments are unable to create.

Firstly added check if perhaps some different charset is used in attachment name, resulting in more then 1 byte per character in Filename. This way we can distinguish and treat different encoding separately when shortening Filename in order for it to be acceptable by system Filename byte size limitations.

Secondly added limit of 222 bytes per Filename without extensions, leaving some 'free space' for additional extensions that OTRS is adding in ArticleStorageFS. If this limit is exceeded we shorten by characters appropriately by how much bytes are excess.

Additionally escaped FilenameCleanUp() function for additional files that OTRS is creating to keep file integrity, this function has already been called on main Filename. Created UnitTest ArticleStorageNameLenght.t where we test this behaviour.

My apologise for such a long explanation, but this required such approach. Please if you have any issues or comments regarding this fix let me know, I will be more then glad to clarify anything.

P.S. This fix is also solving bug #11132 http://bugs.otrs.org/show_bug.cgi?id=11132, where issue was Cyrillic letters which use 2 byte per character.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/EmailParser.pm
+++ b/Kernel/System/EmailParser.pm
@@ -695,6 +695,13 @@ sub PartsAttachments {
             String => $Part->head()->recommended_filename(),
             Encode => 'utf-8',
         );
+
+        # cleanup filename
+        $PartData{Filename} = $Kernel::OM->Get('Kernel::System::Main')->FilenameCleanUp(
+            Filename => $PartData{Filename},
+            Type     => 'Local',
+        );
+
         $PartData{ContentDisposition} = $Part->head()->get('Content-Disposition');
         if ( $PartData{ContentDisposition} ) {
             my %Data = $Self->GetContentTypeParams(
@@ -724,11 +731,11 @@ sub PartsAttachments {
         my ($SubjectString) = $Part->as_string() =~ m/^Subject: ([^\n]*(\n[ \t][^\n]*)*)/m;
         my $Subject = $Self->_DecodeString( String => $SubjectString );
 
-        # trim whitespace
-        $Subject =~ s/^\s+|\n|\s+$//g;
-        if ( length($Subject) > 246 ) {
-            $Subject = substr( $Subject, 0, 246 );
-        }
+        # cleanup filename
+        $Subject = $Kernel::OM->Get('Kernel::System::Main')->FilenameCleanUp(
+            Filename => $Subject,
+            Type     => 'Local',
+        );
 
         if ( $Subject eq '' ) {
             $Self->{NoFilenamePartCounter}++;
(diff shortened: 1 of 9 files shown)


carlosfrodriguez:
Hi @s7design 

Thank you for your pull request, the solution looks very interesting, thank you for the unit test, the only thing that I'm not sure is what happens if we have also a long extension?

I wonder if it would be better to take into account the extension length while doing the substring of the filename? I mean to reduce the name even more leaving the extension  intact.

((enjoy))
cr

s7design:
Hello @dvuckovic and @carlosfrodriguez ,

Updated this PR with different logic and approach to the problem at hand. With Carlos feedback, now we calculate both file name and extension into consideration for the maximum byte size for complete file name. Used perl chop() function in infinite while loop removing character by character ( leaving extension intact ) until we get acceptable hardcoded byte size ( 220 bytes with extension ) when we finish loop.

Additional testing on Oracle DB showed some different behaviour when files are longer then 255 characters ( including Latin as well ). Included FIleNameCleanup() function for FileStorageDB as well where we shorten string if they are too long.

UnitTest is updated. Please let me know if there are any questions regarding this update..

Regards,
Sanjin

dvuckovic:
Hi Sanjin,

latest state has been merged and it's working for me locally, so I pushed it to our unit test platform to test it further. Let's see if it works this time. Thanks!

Regards,
Dusan

mgruner:
See also #1181. I believe we need to combine the best from both PRs when merging.

dvuckovic:
Hi Sanjin,

I successfully tested all your changes locally and just pushed them to our unit test server. It seems the server is out of order right now, but let's see if we can get some results from it in the coming days.

As Travis CI build suggested, there was an issue with an additional unit test: `WebUploadCache.t`. I included the fix for it too. Let's see if everything is covered for now.

Thanks!

Regards,
Dusan

carlosfrodriguez:
Hi @dvuckovic,

How we should proceed with this PR?

((enjoy))
cr

dvuckovic:
Approach used in this fix is not good, we should aim to refactor this method to a more sensible variant. It's really hard to know which behavior you can expect based on different parameters. Since article storage was refactored for master branch, it makes sense to also do it there only. Because of this, fix would be applicable only in OTRS 6.

Closing for now.

PR #1535, opened 2016-10-26
[Rel-4_0 Bug 12367] - * in CustomerID breaks all CustomerUser.CustomerIDs

s7design:
http://bugs.otrs.org/show_bug.cgi?id=12367

Hello,

Hereby is our proposal for fixing reportes issue for rel-4_0. Forbid input of special characters '*' and '%' as CustomerID when used alone or in combination with one-another or other characters. Fix is applied for CustomerCompanyAdd() and CustomerCompanyUpdate().

Modified UT check for these restrictions.

Please let me know if you have any questions or issues regarding this bug fix approach.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/CustomerCompany.pm
+++ b/Kernel/System/CustomerCompany.pm
@@ -120,6 +120,15 @@ sub CustomerCompanyAdd {
         }
     }
 
+    # check if special characters '*' or '%' are used in CustomerID param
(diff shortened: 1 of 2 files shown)


s7design:
Martin fixed it on different way. See bug report

Regards
Zoran

s7design:
Martin fixed it on different way. See bug report

Regards
Zoran

PR #1538, opened 2016-10-27
Remove redundant and wrong permission check

moritz:
There does not seem to be a permission type "customer", and there is a check for a configurable permission type right above it.


mgruner:
Applied. There is a built-in permission type "customer" (see System::Permission in Ticket.xml) that is also the default value of the configured permission type. Thanks @moritz!

dvuckovic:
This proved to be a partial fix to the [bug#8661](https://bugs.otrs.org/show_bug.cgi?id=8661).

PR #1540, opened 2016-10-30

urbalazs:
Hi @mgruner 
I found some strings that are not marked for translation. Branch rel-5_0 is also affected.


mgruner:

PR #1541, opened 2016-10-30
Update/add REST API Sample config

frennkie:
This PR adds a new route to the existing example config for the GenericTicketConnectorREST setup.
Additionally this PR adds a REST sample for the the Link API.

### DIFF
--- a/development/webservices/GenericFAQConnectorREST.yml
+++ b/development/webservices/GenericFAQConnectorREST.yml
@@ -34,19 +34,19 @@ Provider:
         LanguageList:
           RequestMethod:
           - GET
-          Route: LanguageList
+          Route: /LanguageList
         PublicCategoryList:
           RequestMethod:
           - GET
-          Route: PublicCategoryList
+          Route: /PublicCategoryList
         PublicFAQGet:
           RequestMethod:
           - GET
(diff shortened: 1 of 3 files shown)


mgruner:
Hi @frennkie, thanks for your CLA and this proposal. I believe that for the LinkObject we don't have operations yet in OTRS, so that this example would not make sense, right?

frennkie:
Dear @mgruner, there is an extension to which the OTRS Company will grant access to for Business Solution customers (on request) that implements the Link API.
I am using this at the office and it is working. 
I also have a 27 page PDF documentation (Title: "OTRSGenericInterfaceLinkMechanism") for the extension but I'm not sure whether I can share it.
I also have no clue why OTRS does not offer this as a open extension all OTRS Free customers.

jensoliver:
Hi Martin,

we should integrate at least the PR in the Framework

Thanks

Jens

dvuckovic:
Merged changes to `GenericTicketConnectorREST.yml` only in 67ae032, 54e15ac and 4e770d9, since this functionality is part of the framework. Other two files will be added to their respective add-ons (OTRSGenericInterfaceLinkMechanism and FAQ). Thanks @frennkie!

PR #1543, opened 2016-11-02
Fix relative path.

mstk712:
Don't access customer.pl and public.pl running on PSGI/Plack environment.
Relative path is incorrect, it does not appear customer.pl and public.pl.
I had to correct the path because it is.


mgruner:
Hello @mstk712. Thanks for the hint, I simplified this a bit and merged it in bee53a8a244fcefc7c97fadf05b847991b36b689 and backported to OTRS 5 and 4.

PR #1544, opened 2016-11-02

urbalazs:
Hi @mgruner
Since OTRS 4 doesn't support Transifex, this is the only way to update the translation.


mgruner:
Merged, thanks!

PR #1547, opened 2016-11-06
Marked error messages as translatable

urbalazs:
Hi @mgruner 
I found some messages, that are not marked for translation. Branch rel-5_0 is also affected.


mgruner:

PR #1548, opened 2016-11-08
Marked JS messages as translatable

urbalazs:
Hi @zottto 
I found some messages in JS files, which are not marked for translation.


zottto:
Merged with 450cf93.

PR #1549, opened 2016-11-10
leftover dtl tag from older frameworks

straubr:
Updated from dtl to tt.

Well this is only a tiny minor thing...


mgruner:
Applied with slight changes, thanks a lot @straubr!

PR #1551, opened 2016-11-12
Refix Wrong variable <OTRS_CUSTOMER_REALNAME> in AutoResponses #4640

akosiaris:
This is a one-line partial revert of 82d55d6 which introcuded a condition where
the From field in an Auto Response would be badly generated. This is documented
in Bug #4640 comments 15 and 16. The field that should be used instead is the To
field and not the From since the generated email already has those headers set

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -1577,7 +1577,7 @@ sub _Replace {
 
                 # generate real name based on sender line
                 if ( !$From ) {
-                    $From = $Data{From} || '';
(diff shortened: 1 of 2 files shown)


mgruner:
@mrcbnsls can you review this please, since you worked on the original commit?

akosiaris:
FWIW, I 've updated the tests as well and now all pass

akosiaris:
Hi,

any chance of this getting reviewed ? Thanks!

dvuckovic:
This has been superseded by more complete fix in #1635. Thanks for your patience!

PR #1552, opened 2016-11-13
Fixed wrong sysconfig subgroups

urbalazs:
Hi @mgruner 
These two sysconfig options are in wrong subgroups. Branch rel-5_0 is also affected.


mgruner:
Merged, very nice, thanks @urbalazs.

PR #1553, opened 2016-11-13
Superfluous plural form

urbalazs:
Hi @mgruner 
There is only one title field, so plural form is unnecessary here. Branch rel-5_0 is also affected.


mgruner:

PR #1557, opened 2016-11-14
Fixed error argument "" isn't numeric in numeric eq (==) at [...] AdminUserGroup.pm line 153

pboguslawski:
In AdminRoleGroup when saving permissions added for all agents using
column header (i.e. RO), POST request contains empty value in first
array item, i.e.

    Action=AdminUserGroup&Subaction=ChangeGroup&ID=8&ro=&ro=1&ro=2&ro=3&ro=4

which throws warnings like

    -e: Argument "" isn't numeric in numeric eq (==) at /opt/otrs//Kernel/Modules/AdminUserGroup.pm line 153.

to apache error log. Permissions are saved correctly but OTRS should
not pollute logs.

Related: https://dev.ib.pl/ib/otrs/issues/103
Author-Change-Id: IB#1061101


dvuckovic:
Merged in ed7b722, 06c4ebd and 4d0b6d0, thanks @pboguslawski!

PR #1558, opened 2016-11-14
Marked notification messages as translatable

urbalazs:
Hi @mgruner 
This patch makes it possible to translate the notification messages, like select a time zone, etc.


mgruner:
Hello Balazs, "Data" is meant to insert raw HTML data which is not translatable. "Info" is for translatable text; I fixed the time zone notification to use "Info" instead of "Data" (109d75127ebc3b504d5c92b0e31c588ccfb81bae). If you find other places, we should fix them too.

mgruner:
Thanks!

PR #1561, opened 2016-11-15
Put Save buttons into a separate widget

urbalazs:
Hi @mgruner 
Now the save buttons are inside the "Edit ACL structure" widget. If the user collapses the widget, the buttons disappear. This pull request puts these buttons into a separate widget (as it is in [AdminProcessManagementProcessEdit.tt](/OTRS/otrs/blob/master/Kernel/Output/HTML/Templates/Standard/AdminProcessManagementProcessEdit.tt)).


mrcbnsls:
Thanks, merged in 22a0bac and 27f9d74.

PR #1565, opened 2016-11-16
Fixed race condition in ArticleCreate

pboguslawski:
OTRS searches for ID of article inserted to database using

    TicketID
    MessageID
    From
    Subject
    IncomingTime

and highest ID in table; because IncomingTime resolution is 1 s, if two
articles with the same MessageID arrive to database in the same second
(i.e. poor web browser sends duplicated POST creating same note or two
e-mail messages arrive with the same Message-ID /if send To more than
one OTRS system address/) , race may occur and OTRS process may append
article data (attachments, plain and html files) to other ArticleID
than inserted by this process.

Related: https://dev.ib.pl/ib/otrs/issues/105
Author-Change-Id: IB#1021899

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2016-??-??
+ - 2016-11-16 Fixed race condition in ArticleCreate.
  - 2016-11-04 Make it possible to search for emtpy dynamic fields via the TicketSearch API, thanks to Rolf Schmidt and Moritz Lenz.
  - 2016-11-02 Added sort criteria to TicketSearch call in PendingCheck console command. Thanks to Torsten Thau.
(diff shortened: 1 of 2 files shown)


pboguslawski:
Additional idea: in the future consider also introducing "transactions" to OTRS, i.e. postmaster should not leave orphaned ticket in case of articlecreate() failure. This may require using SQL transactions also to ensure consistency. Using perl DBI exceptions (eval, begin_work, commit, rollback) for catching any SQL failures may simplify error detection also. Deadlock exceptions and replying transactions may be also nice feature if possible.

mgruner:
Fix applied with small modifications like including the process id in the insert fingerprint. Thanks a lot @pboguslawski!

PR #1566, opened 2016-11-17
Superfluous colon in table heading

urbalazs:
Hi @mgruner 
There are no colons in table headings - except in this table. This proposal removes the superfluous colons. I also apply a small wording improvement. Branch rel-5_0 is also affected.


dvuckovic:
Merged in be86ca7, 675059d and 5ff7975, thanks @urbalazs!

PR #1568, opened 2016-11-21
Followup for bug#12418

urbalazs:
Hi @mrcbnsls 
According to [admin manual](http://otrs.github.io/doc/manual/admin/stable/en/html/what-is-the-queue-overview.html) only the current queue will be bolded.


mrcbnsls:
Thanks, this has been fixed in 24dc9881cbd24eb6dc284031583c25c75f7ddb66.

PR #1569, opened 2016-11-21
Fixed standard available permissions settings

urbalazs:
Hi @mgruner 
The setting "System::Permission" is in both Framework.xml and Ticket.xml, so always the latter will be applied. Therefore I delete it from Framework.xml, and also moved the setting "System::Customer::Permission" into the Ticket.xml


mgruner:
Hi @urbalazs, thanks a lot. The original state is as designed. If you remove Ticket.xml (to deactivate the ticket engine), the system has less permissions by default. I would leave it as it is.

PR #1570, opened 2016-11-21
Mandatory Owner removes empty selection

capeit-tto:
If Owner is set to mandatory the empty selection must not be removed in order to force an explicit selection of value. If no empty selection is possible, the user is likely to forget to actually pick a new owner.

### DIFF
--- a/Kernel/Modules/AgentTicketProcess.pm
+++ b/Kernel/Modules/AgentTicketProcess.pm
@@ -3051,15 +3051,12 @@ sub _RenderOwner {
         ValidateRequired => '',
     );
 
-    my $PossibleNone = 1;
-
     # if field is required put in the necessary variables for
     #    ValidateRequired class input field, Mandatory class for the label
     #    do not allow empty selection
     if ( $Param{ActivityDialogField}->{Display} && $Param{ActivityDialogField}->{Display} == 2 ) {
         $Data{ValidateRequired} = 'Validate_Required';
         $Data{MandatoryClass}   = 'Mandatory';
-        $PossibleNone           = 0;
     }
 
     my $SelectedValue;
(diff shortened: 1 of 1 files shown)


capeit-tto:
...responsible is affected as well.

carlosfrodriguez:
Hi @capeit-tto,

This was also solved for OTRS 6 in this bug report: https://bugs.otrs.org/show_bug.cgi?id=11028, can you please take a look into it, if it also satisfies the issue as well as your proposal I can back port it for OTRS 5, to have a consistent solution.

mgruner:
@carlosfrodriguez can you please clarify with @dvuckovic why this was done for master only? There might be a good reason.

dvuckovic:
@carlosfrodriguez This has been implemented for master because we didn't feel comfortable with changing default behavior in patch level, and in an effort to maintain full backwards compatibility with current processes. But feel free to re-asses it: if it makes sense to you to backport this fix for OTRS 5 and 4 as well, then by all means.

Looking now at the implementation, we realize that it can be further improved, by simplifying the `BuildSelection()` call, similar to the proposal in this pull request. I expect there will be a follow-up applied on master in coming days.

carlosfrodriguez:
Hi @dvuckovic @capeit-tto,

I wasn't considering the impact that this could provoke in a patch level release, so I guess we should keep it just for OTRS 6.

@dvuckovic, The bugzilla team will apply the follow-up? that was not fully clear.

((enjoy))
cr

carlosfrodriguez:
Hi @dvuckovic,

Never mind I see the bug was updated.
((enjoy))
cr

carlosfrodriguez:
Hi @capeit-tto,

This pull request inspired us to create a better fix to the one we had already in OTRS 6 please take a look at https://github.com/OTRS/otrs/commit/3537be32edea0e7e031921f5b4ab833c747b9192

I guess we can close this pull request

((enjoy))
cr

PR #1571, opened 2016-11-21
Counters incremented transactionally in database

pboguslawski:
OTRS uses text file to store counter for ticket numbers (in AutoIncrement,
Date and DateChecksum modules). This is file is not protected with locking
during updates which may cause ugly race conditions (concurrent processes
trying to generate new ticket number) and errors like one below caught
in OTRS 3:

    DBD::mysql::db do failed: Duplicate entry '100000034' for key 'ticket_tn' at /opt/otrs/Kernel/System/DB.pm line 475.
    ERROR: OTRS-otrs.PostMasterDaemon.pl-4 Perl: 5.14.2 OS: linux Time: Sun Oct 20 11:23:34 2013

Managing counter outside of database is not elegant nor easy
to maintain in cluster environment. Counters should be incremented
and read transactionally with locking to avoid races.

This mod extends OTRS interface to DBI with transaction handling
and introduces new counter module AutoIncrementDB
for Ticket::NumberGenerator setting in SysConfig, which stores
counter in new SQL table otrs.counter and updates it transactionally.

Related: https://dev.ib.pl/ib/otrs/issues/106
Author-Change-Id: IB#1015965


pboguslawski:
Counter routines introduced in this mod are ment to be common for extra modules like in https://github.com/OTRS/ITSMChangeManagement/pull/41

ITSMConfigItem module uses its own counter table configitem_counter without transactions (races possible); this may be also migrated in future to common counter table.

mgruner:
Hello @pboguslawski, we are currently working on a similar solution for OTRS 6 that works DB based too and enhances the existing ticket counters, but does not require transactions in the database. I'll close this PR therefore. Thanks a lot for your proposal!
@carlosfrodriguez FYI.

PR #1576, opened 2016-11-24
Bugfix for 12293: Added PostMaster::PreCreateFilterModule

dennybresch:
Hi Martin,

here is discussed Bugfix for 12293 - Added PostMaster::PreCreateFilterModule as new Postmaster layer before article creation.

https://bugs.otrs.org/show_bug.cgi?id=12293

Best regards, Denny


mgruner:
Fix applied with small modifications, thanks a lot @dennybresch and @thorsteneckel!

PR #1582, opened 2016-12-08
Renamed Favourite to Bookmark

urbalazs:
Hi @mgruner 
Here is an other wording change proposal (like in https://github.com/OTRS/otrs/pull/1395).
In the redesigned admin panel of OTRS 6 the admin can mark some widget as "Favourite". These widgets will display in the menu, so the word "Bookmark" will be better choice. Furthermore the star icon means bookmark in many other places (e. g. in browsers).
I changed only the visible strings. There are many inner variables in the source code with name "favourite" (partly or fully). If you agree with my proposal, then it also would be great to rename the related variables in the source code.

### DIFF
--- a/Kernel/Config/Files/Framework.xml
+++ b/Kernel/Config/Files/Framework.xml
@@ -9111,7 +9111,7 @@ via the Preferences button after logging in.
         </Setting>
     </ConfigItem>
     <ConfigItem Name="Frontend::NavBarModule###2-AdminFavourites" Required="0" Valid="1">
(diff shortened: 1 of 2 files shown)


mrcbnsls:
Hi @urbalazs, I disagree here. What you can do in the admin screen is not actually 'bookmarking'. Bookmarking means that you want to return later to this exact place (like adding a piece of paper to a certain page in a book that you're reading) or website (in terms of browsers where you can bookmark websites). On the admin overview, you want to mark certain items as favourite for quicker access. So at least from my perspective it makes perfect sense to call this 'favourite'.

PR #1585, opened 2016-12-11
Improved dialog buttons

urbalazs:
Hi @mrcbnsls,

This PR adds "`CallForAction`" class to dialog buttons. I'm not sure about to also add "`Primary`" class for the buttons. If only "`CallForAction`" is added, the deletion of process element has strange look and feel. See this screenshot:

dialog-buttons

Where can I find further information about the use of CSS classes in OTRS? What does `CallForAction`, `Primary`, etc. means? I think, the look and feel chapter is missing from the developer manual. [Bootstrap](http://getbootstrap.com/) has very good documentation, maybe the developer manual would need something similar.

Please, apply this patch also to rel-5_0.

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.Dialog.js
+++ b/var/httpd/htdocs/js/Core.UI.Dialog.js
@@ -362,7 +362,7 @@ Core.UI.Dialog = (function (TargetNS) {
                 Type: 'Close',
                 Function: Params.OnClose
             }];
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @urbalazs, thanks, your changes have been merged into master and rel-5_0. Due to these changes, many places in the code which added the 'CallForAction' class manually, got obsolete now, which I fixed in separate commits.

The 'Primary' class basically does nothing more than adding bold style to the button text to visualize that this would probably be the button the user would want to click after calling the dialog.

Currently there is no documentation on things like those, but there are plans to add a full-blown css reference some time soon.

PR #1587, opened 2016-12-13
[ Rel-5_0 Bug 10691 ] - No CustomerID shown after TicketCreate and modified Unit test

s7design:
https://bugs.otrs.org/show_bug.cgi?id=10691

Hello @mgruner 

This is proposal for no auto-assigning CustomerID and CustomerUser for unknown incoming emails in accordance with your suggestion. Because of that, if there are no CustomerID and CustomerUser found, parameters with their values which are used later for ticket creating stay undefined.

IF condition is modified because there is the same condition in parent IF and '$GetParam{'X-OTRS-CustomerUser'}'  can not change 
(https://github.com/s7design/otrs/blob/5241e87234e22ba647dac196f57253926e7d18e3/Kernel/System/PostMaster/NewTicket.pm#L187).

Regards,
Milan

### DIFF
--- a/Kernel/System/PostMaster/NewTicket.pm
+++ b/Kernel/System/PostMaster/NewTicket.pm
@@ -206,6 +206,11 @@ sub Run {
         }
     }
 
+    # if there is no customer user found!
+    if ( !$GetParam{'X-OTRS-CustomerUser'} ) {
+        $GetParam{'X-OTRS-CustomerUser'} = $GetParam{SenderEmailAddress};
+    }
+
     # get ticket owner
     my $OwnerID = $GetParam{'X-OTRS-OwnerID'} || $Param{InmailUserID};
     if ( $GetParam{'X-OTRS-Owner'} ) {
--- a/Kernel/System/PostMaster/NewTicket.pm
+++ b/Kernel/System/PostMaster/NewTicket.pm
@@ -184,7 +184,7 @@ sub Run {
         }
 
         # take CustomerID from customer backend lookup or from from field
-        if ( $CustomerData{UserLogin} && !$GetParam{'X-OTRS-CustomerUser'} ) {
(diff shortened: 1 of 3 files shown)


scripts/test/Console/Command/Maint/PostMaster/Read.t
```diff
@@ -12,15 +12,85 @@ use utf8;
...
+$Self->True(
+    $CustomerUserID,
+    "CustomerUserID $CustomerUserID is created",
+);
+
+my @Tests = (
+    {
+        EmailAddress => 'Unknown' . $CustomerUserEmailAddress,
+        Result       => {
+            CustomerUserID => undef,
```
mgruner: Actually this is not the requested behaviour. CustomerUserID should still be filled with the email address to have at least something to identify in the ticket. But CustomerID should stay empty, just like it is if you assign an unknown CustomerUser in AgentTicketCustomer.

s7design:
Hi @mgruner,

Thanks for your follow-up. No problem it will be updated as you suggested.

Regards
Milan

s7design:
Hello @mgruner 

PR is updated in accordance with your suggestions. Please inform me if you have any comment for these changes.

Regards,
Milan

Kernel/System/PostMaster/NewTicket.pm
```diff
@@ -184,7 +184,7 @@ sub Run {
         }
 
         # take CustomerID from customer backend lookup or from from field
-        if ( $CustomerData{UserLogin} && !$GetParam{'X-OTRS-CustomerUser'} ) {
+        if ( $CustomerData{UserLogin} ) {
```
mgruner: I omitted this change, I believe it is still valid.

mgruner:
Merged, thanks!

PR #1588, opened 2016-12-14
Rel-5_0 - Extended JS validation for radio button

s7design:
Hello @mgruner 

This is our proposal for extending of JS validation for radio button. First, we rolled back fix in OTRSAdvancedGenericAgent package. On Rel 5_0 branch extending for radio buttons is done for 'Validate_DateInFuture' and 'Validate_DateNotInFuture' validation methods. On Master branch there are some differences and it will be done separately.

There is not the most generic solution because finding by value but there are more than one radio buttons in parent HTML element and they can be distinguished only by value (0, relative, 1). Maybe it could be found like the last radio button in this parent element but this is also not general solution.
JS Unit test is extended too.

This is input radio button in OTRSAdvancedGenericAgent
> <input type="radio" value="1" name="${Prefix}Used" $UsedAbsolute>

because of that in Core.Form.Validate.js the element is found on this way

> input[type=radio][value=1]');

If you have any comments or suggestions for these changes, please let me know.

Regards,
Milan

### DIFF
--- a/var/httpd/htdocs/js/Core.Form.Validate.js
+++ b/var/httpd/htdocs/js/Core.Form.Validate.js
@@ -395,7 +395,8 @@ Core.Form.Validate = (function (TargetNS) {
     }, "");
 
     $.validator.addMethod("Validate_DateInFuture", function (Value, Element) {
-        var $DateSelection = $(Element).parent().find('input[type=checkbox].DateSelection');
+        var $DateSelection = $(Element).parent().find('input[type=checkbox].DateSelection, input[type=radio][value=1]');
+
         // do not do this check for unchecked date/datetime fields
         // check first if the field exists to regard the check for the pending reminder field
         if ($DateSelection.length && !$DateSelection.prop("checked")) {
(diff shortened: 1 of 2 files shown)


var/httpd/htdocs/js/Core.Form.Validate.js
```diff
@@ -395,7 +395,8 @@ Core.Form.Validate = (function (TargetNS) {
     }, "");
 
     $.validator.addMethod("Validate_DateInFuture", function (Value, Element) {
-        var $DateSelection = $(Element).parent().find('input[type=checkbox].DateSelection');
+        var $DateSelection = $(Element).parent().find('input[type=checkbox].DateSelection, input[type=radio][value=1]');
```
mgruner: As you suggested, locating by value=1 is not so elegant. Since we already use .DateSelection for checkboxes, I'd suggest using the same for radio buttons and adding it to the selector as well, as you proposed: `var $DateSelection = $(Element).parent().find('input[type=checkbox].DateSelection, input[type=radio].DateSelection');`

s7design:
Hi @mgruner,

Thanks Martin :) We will change it here and of course I will update the package OTRSAdvancedGenericAgent. In the package we will add class DateSelection.

Regards
Milan

s7design:
Hello @mgruner 

PR is updated as you suggested. Selector is modified here and class 'DateSelection' is added in the package OTRSAdvancedGenericAgent.

Regards,
Milan

mgruner:
Hello Milan, I'm getting errors in the JSUnitTest with this. Did you forget to update the test cases maybe?


bildschirmfoto 2016-12-23 um 11 46 35

s7design:
Hi @mgruner 

Thank you for your feedback. It will be solved very soon.

Regards,
Milan

s7design:
Hi @mgruner 

JSUnitTest is modified as you suggested.

Regards,
Milan

mgruner:
Thanks Milan! Applied to OTRS 5 and master.

PR #1590, opened 2016-12-16
[ Rel-5_0 Bug 12480 ] - Fixed bug that Bulk function ignores agent/role permissions

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12480

Hello @dvuckovic  

This is our proposal for ignoring queues - agent/role permissions in Bulk.
JS code for Owner field update on queue change is added in Core.Agent.TicketAction.js and subaction block for AJAX update processing is placed in AgentTicketBulk PM module.
Order of fields in Bulk screen is changed in accordance with suggestions in bug report.
Selenium test is extended for updating of owner field.
Please let me know if you have any suggestion or comment for this PR.

Regards,
Milan

### DIFF
--- a/Kernel/Modules/AgentTicketBulk.pm
+++ b/Kernel/Modules/AgentTicketBulk.pm
@@ -32,6 +32,7 @@ sub Run {
     my $ParamObject  = $Kernel::OM->Get('Kernel::System::Web::Request');
     my $LayoutObject = $Kernel::OM->Get('Kernel::Output::HTML::Layout');
     my $TicketObject = $Kernel::OM->Get('Kernel::System::Ticket');
+    my $ConfigObject = $Kernel::OM->Get('Kernel::Config');
 
     if ( $Self->{Subaction} eq 'CancelAndUnlockTickets' ) {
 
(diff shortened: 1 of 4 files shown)


dvuckovic:
Hi Milan,

thanks for your proposal, please find comments below.

In new AJAX update block you implemented, different permission level is being used compared to when the Owner field is printed on page for the first time. IMO, you should either consider changing `move_into` to `rw` like in the render code below, or changing both of them to `move_into` (which makes more sense, for sure).

I am getting different results when I first open the Owner field, and when I then reset the Queue field to be empty (no change). I suspect this has something to do with the permission level used as already stated, but it should be checked nonetheless. IMO, Owner field should be identical if user decides they won't change the Queue field after all.

There is also a possible field on the page for ticket Responsible, which is an optional feature that can be turned on in SysConfig. I think the check should be implemented for this field too, since it behaves almost identically to the Owner. Can you please try to add this use case too?

Feel free to update this PR with outlined changes, I will then check it out again.

Thanks again for your code, keep up the good work!

Regards,
Dusan

Kernel/Modules/AgentTicketBulk.pm
```diff
@@ -88,8 +89,49 @@ sub Run {
...
+            Valid => 1
+        );
+
+        # Put only possible rw agents to owner list.
+        if ( !$ConfigObject->Get('Ticket::ChangeOwnerToEveryone') && $QueueID ) {
+            my %AllGroupsMembersNew;
+            my $GroupID = $Kernel::OM->Get('Kernel::System::Queue')->GetQueueGroupID( QueueID => $QueueID );
+            my %GroupMember = $Kernel::OM->Get('Kernel::System::Group')->PermissionGroupGet(
+                GroupID => $GroupID,
+                Type    => 'move_into',
```
dvuckovic: This permission level is different than what is used below.

s7design:
Hello @dvuckovic 

Thank you for follow-up. PR is modified as you suggested.
AJAX update of queue field is changed because of case when queue field is empty, responsible field update is added and Selenium test is modified accordingly.
Please inform me if you have any comment for these changes.

Regards,
Milan

dvuckovic:
Hi Milan,

your PR has been merged in de2ef77, 3d3575 and 1a9773e. There were some small style changes applied to your code, please see commits above. Other than that, great job!

Thanks!

Regards,
Dusan

PR #1591, opened 2016-12-20
[ Rel-5_0 Bug 12482 ] - Fixed showing link of DF in Customer interface

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12482

Hello @mrcbnsls 

This is our proposal for fix this bug. 'a' tag with 'href' link attribute is added in CustomerTicketZoom.tt.
Selenium test is created for checking this.
Please let me know if you have any feedback for this PR.

Regards,
Milan

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
+++ b/Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
@@ -225,7 +225,7 @@
 [% RenderBlockStart("TicketDynamicField") %]
                     <li>
                         <span class="Key">[% Translate(Data.Label) | html %]:</span>
-                        <span title="[% Data.Title | html %]">[% Data.Value %]</span>
+                        <span title="[% Data.Title | html %]"><a href="[% Data.Link | Interpolate %]" class="DynamicFieldLink" target="_blank">[% Data.Value %]</a></span>
                     </li>
 [% RenderBlockEnd("TicketDynamicField") %]
 # example of how to use fixed dynamic field blocks for customizations
--- a/Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
+++ b/Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
@@ -404,9 +404,7 @@
 [% WRAPPER JSOnDocumentComplete %]
 <script type="text/javascript">
     Core.Customer.TicketZoom.Init();
-    if (typeof Core.UI.Floater !== 'undefined') {
-        Core.UI.Floater.Init();
-    }
+    Core.UI.Floater.Init();
     $('a.AsPopup').bind('click', function (Event) {
         Core.UI.Popup.OpenPopup($(this).attr('href'), 'TicketAction');
         return false;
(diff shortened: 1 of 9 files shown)


mgruner:
Hi Milan. Looking nice! In the agent interface, there is also a 'link preview' feature, which will enable a floater with a preview for a configured URL. Could you add that in the same manner as well? Otherwise we'd wait for the next bug about this. 😁

s7design:
Hi @mgruner 

Thank you for your follow-up. It will be done as soon as possible.

Regards,
Milan

s7design:
Hi @mgruner 

'Link preview' feature is enabled in customer interface as you suggested.
Core.UI.Floater.js is used for floater in the agent interface and also added to customer interface. Usage for interfaces is separated by added IF condition because HTML page structures are not identical for agent and customer interfaces.

Style for floater is added in Customer/Core.Default.css as 'subsection Floaters'. It is similar to agent CSS file.
Settings for left and right floater positions are modified with substraction of 50 pixels to avoid lateral floater edges too close to the left or right side of display.
Link preview inspection is added to Selenium test.

Please let me know if you have any comment for the last commit.

Regards,
Milan

s7design:
Hi @mgruner 

PR is updated with config change.

Regards,
Milan

var/httpd/htdocs/js/Core.UI.Floater.js
```diff
@@ -184,7 +184,13 @@ Core.UI.Floater = (function (TargetNS) {
                 $('[data-trigger="floater"]').removeClass('FloaterOpen');
                 $TriggerObj.addClass('FloaterOpen');
 
-                $FloaterObj = $('#AppWrapper > .MetaFloater.Hidden').clone();
+                // Separate creating of $FloaterObj for agent and customer interface.
+                if (Core.Config.Get('CGIHandle').search(/customer.pl/) > 0) {
```
mgruner: I believe we should consider simplifying that by removing the body/#AppWrapper element from the selector, making this evil switch unneccessary. @mrcbnsls would you agree?

var/httpd/htdocs/js/Core.UI.Floater.js
```diff
@@ -196,15 +202,15 @@ Core.UI.Floater = (function (TargetNS) {
                 FloaterWidth = parseInt($FloaterObj.outerWidth(), 10);
 
                 // set left position
-                if (TriggerOffset.left < FloaterWidth / 2) {
+                if (TriggerOffset.left - 50 < FloaterWidth / 2) {
```
mgruner: @mrcbnsls what do you think about this?

Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
```diff
@@ -404,9 +404,32 @@
...
+    </a>
+    <i class="fa fa-spinner fa-spin"></i>
+    <div class="NoPreview">
+        <i class="fa fa-warning"></i>
+        [% Translate("A preview of this website can't be provided because it didn't allow to be embedded.") | html %]
+    </div>
+    <div class="Content" style="display: none;">
+        <iframe sandbox="allow-scripts allow-forms"></iframe>
+    </div>
+</div>
```
mrcbnsls: I think this part should go into an include as it is identical to whats included in AgentTicketZoom, right, @s7design?

var/httpd/htdocs/js/Core.UI.Floater.js
mrcbnsls: I'm not convinced this is necessary. If possible, things like this should be calculated dynamically from the boundaries of given elements. @s7design, please check if it is possible to calculate it here. And please make sure it's still working in the agent interface, too.

Kernel/Output/HTML/Templates/Standard/CustomerTicketZoom.tt
```diff
@@ -404,9 +404,32 @@
 [% WRAPPER JSOnDocumentComplete %]
 <script type="text/javascript">
     Core.Customer.TicketZoom.Init();
+    if (typeof Core.UI.Floater !== 'undefined') {
+        Core.UI.Floater.Init();
+    }
```
mrcbnsls: Why would Core.UI.Floater possibly be undefined? I don't think this is necessary, is it, @s7design?

s7design:
Hi @mrcbnsls 

PR is updated in accordance with your suggestions.

MetaFloater.tt is created and included to AgentTicketZoom and CustomerTicketZoom. Did you think for this in your comment?

Checking if Core.UI.Floater is undefined is removed.

Calculations for floater classes are changed to dynamic. Other proposal is that this subtraction could be removed.

If you have any comment, please inform me.

Regards,
Milan

mrcbnsls:
Merged successfully (with some small changes) for both rel-5_0 and master. Thanks!

PR #1592, opened 2016-12-21
Update Ticket.xml

florianedlhuber:
The blinking queue setting is annoying and should be disabled by default.


mgruner:
@mrcbnsls can you review this please? We should not change this in a stable branch, I guess?

mrcbnsls:
Though it could probably be discussed whether this is just a matter of taste or not, I agree that it's probably more annoying than helpful for most users of OTRS, so I applied the suggested change (for master only) and also changed the relevant place in the admin documentation to reflect it. Thanks!

PR #1583, opened 2016-12-10
Improved GenericAgent screen

urbalazs:
Hi @mgruner 
This pull request contains the following changes:

- marked strings as translatable
- replaced old dropdowns with modern ones
- added "week" as selectable pending time unit

Please also apply this patch to branch rel-5_0.

PR #1593, opened 2016-12-23
Generic agent

urbalazs:
Hi @mgruner 
I removed some code from https://github.com/OTRS/otrs/pull/1583. Now this only marks strings as translatable.


mgruner:

PR #1595, opened 2017-01-06
_Dump JSON::PP::Boolean reference types

richieri:
Sometimes somesystem like Elasticsearch an Jira sends this type of Reference. We got errors because _Dump function is not trading this type of reference.


mgruner:
See #1594

mgruner:
Implemented, closing, see #1594.

PR #1598, opened 2017-01-18
[ Rel-5_0 Bug 11422 new ] - List in the Dashboard: standard priority is ignored if you use a filter

s7design:
http://bugs.otrs.org/show_bug.cgi?id=11422

Hi @mgruner 

New PR is created because there were a lot of changes for rebase (#955).

First commit [572ec7](https://github.com/s7design/otrs/commit/572ec7122395751f645f2f7e2f72d1d3cec6fb3e) separates filtering and ordering. OrderBy is not changed after filtering table by any column but when user clicks on column header, OrderBy is switched from 'Up' to 'Down' and vice versa.
Block parameter OrderBy is determined before passing through the 'for' loops of arrays of columns and all Core.AJAX.ContentUpdate JS events will have the same OrderBy argument.
In AgentDashboardTicketGeneric.tt Refresh block, OrderBy values 'Up' and 'Down' are changed. I think this is more natural that OrderBy is not changed when page is reloaded. In other case, it is much less comfortable.
Selenium test is added appropriately.

Our proposal is in second commit [afa7fa](https://github.com/OTRS/otrs/commit/afa7fac289bdca4501b2d5ac461c4698c7bd16ca) - modified OrderBy changing. Only when user clicks on the same column header, OrderBy will be change from 'Up' to 'Down' and vice versa (or 'Asc' to 'Desc' and 'Desc' to 'Asc'). In other cases, when user clicks on other columns or filters table, OrderBy will not be changed. New parameter 'SortingColumn' is needed for this.
Selenium tests is modified for new 'logic'.

Please inform me what do you think about this PR.

Regards,
Milan

### DIFF
--- a/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketGeneric.pm
@@ -40,6 +40,9 @@ sub new {
         $Self->{$Item} = $ParamObject->GetParam( Param => $Item ) || $Param{$Item};
     }
 
+    # get add filters param
+    $Self->{AddFilters} = $ParamObject->GetParam( Param => 'AddFilters' ) || $Param{AddFilters} || 0;
+
     # set filter settings
     for my $Item (qw(ColumnFilter GetColumnFilter GetColumnFilterSelect)) {
         $Self->{$Item} = $Param{$Item};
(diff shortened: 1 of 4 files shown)


s7design:
Hi @mgruner 

Ordering after refresh of the widget is set that order does not change.

Regards,
Milan

s7design:
Hello,

Closing this PR, fix proposal is in appropriate bug fix branches.

Regards,
Sanjin

PR #1600, opened 2017-01-25
Missing "UserID" parameter in AgentTicketActionCommon.pm

StCyr:
Added a missing "UserID" parameter to the call to _GetSLAs in AgentTicketActionCommon.pm.

Without this additional parameter, the call will failed when no CustomerUser is assigned to the ticket.

Cyrille


mgruner:
Hello Cyrille, thanks for the suggestion. I applied a similar fix in 0ad16a74c74bd97ecf136c44011cea4ea9ca9e95 (directly in the _GetSLAs function). Please let me know if you have any problems with that.

PR #1602, opened 2017-01-25
Add info about Responsible to AgentTicketQueue (M/L)

frennkie:
In the AgentTicketQueue "Large" view there are a few columns listed.
One example is the Ticket Owner. This commit adds an additional row
below Owner that also prints the Ticket Responsible.
The Responsible is only shown if the "Ticket::Responsible" feature is
enabled in Sysconfig.


frennkie:
I noticed that I was just showing the same Agent info for Responsible as for Owner. I fixed that now.
But at the moment I can only print the LoginID of the Responsible. Could somebody please point me to where I can add the First and Lastname of the Responsible.

frennkie:
@zottto: I added a RenderBlock to get the "FullName" to the Template.

This is the result for both "M" and "L" view:

### M before:
otrs_m_owner

### M after:
otrs_m_owner_responsible

### L before:
otrs_l_owner

### L after:
otrs_l_owner_responsible

frennkie:
@zottto I think I addressed the changes you requested. Do you still need input on this from me? If not please remove the "needs update by contributor" label.

mrcbnsls:
Thanks, finally merged in 01c3a2dcae!

frennkie:
@mrcbnsls @mgruner \o/ thanks! :-)

PR #1603, opened 2017-01-26
[ Rel-5_0 Bug 12558 ] - Filtering of "Available Columns" in Allocation List is not working well when columns are changed

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12558

Hi @dvuckovic 

This is our proposal for fix. When some field is moved to 'Available Columns', container has to be re-calculated before filtering.
Please let me know if you have any comment for this PR.

Regards,
Milan


dvuckovic:
Merged in a759ba6, 3e9bb1b and 8d01faa. Thanks @s7design!

PR #1607, opened 2017-01-31
[ Rel-5_0 Bug 12582 ] - Select All Checkbox is shown in Ticket Overview screens when table is empty

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12582

Hi @dvuckovic 

There is our proposal for change behavior described in report on above link. 'Select all' checkbox was shown in Medium and Preview (Large) view when table has no data and it is changed that this checkbox is hidden when table is empty.

Regards,
Milan

### DIFF
--- a/var/httpd/htdocs/js/Core.UI.ActionRow.js
+++ b/var/httpd/htdocs/js/Core.UI.ActionRow.js
@@ -177,6 +177,12 @@ Core.UI.ActionRow = (function (TargetNS) {
             TicketView = 'Small';
         }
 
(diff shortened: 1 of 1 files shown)


zottto:
Merged to rel-5_0 (cb641b9) and master (0a4c2fd) in a slightly different way.

@s7design 
It is safer to search for parent elements with ``closest()`` instead of parent chaining (``parent().parent()``). By adding a class to the list item which should be invisible, it is possible to define the CSS for the neighbor element in the css file instead of using inline CSS.

PR #1608, opened 2017-01-31
[rel-5_0] Improved LinkObject buttons

urbalazs:
Hi @mrcbnsls 
This PR fixes the layout of the LinkObject buttons. Only branch rel-5_0 is affected, the branch master was fixed in the meantime.


mrcbnsls:
Thanks, merged in 3452c40 and added the missing CSS for inline buttons aswell.

PR #1609, opened 2017-01-31
Fixed wrong SysConfig descriptions

urbalazs:
Hi @mgruner 
These SysConfig descriptions are wrong, because the SysConfig options have value "Yes" by default, and the associated settings set CustomerIDs as read-only. So the setting _Controls if CustomerID is read-only in the agent interface._


mgruner:

PR #1611, opened 2017-02-07
Improved SysConfig descriptions

urbalazs:
Hi @mgruner 
These SysConfig settings don't require adding a note by the agent. The ticket note functionality can be turned off below, and these settings will still work. So admins can be set up a special ticket note screen, where the agents can only change the ticket type without adding a note.
Branch rel-5_0 is also affected. Can you please sync with Transifex after merging?


mgruner:
Merged in 08adf61235efd52c77d342172cb7d912023938d0 and backported, thanks! The transifex sync will be done some time before the next release.

PR #1612, opened 2017-02-07
Improved button

urbalazs:
Hi @mrcbnsls 
This button is still old-style designed. Both branch rel-5_0 and master are affected.


mrcbnsls:
Thanks, merged in ec35285. I did it only for rel-5_0, because in 6 the whole sysconfig will be redesigned.

PR #1617, opened 2017-02-15
add support for X-OTRS-FollowUp-…

reneeb:
where … is {Owner|OwnerID|Responsible|ResponsibleID}

For new tickets those headers are already supported, but not for followups.


mgruner:
Merged for OTRS 6, thanks a lot @reneeb!

PR #1619, opened 2017-02-17
[ Master ] - Improved Selenium tests News and ProductNotify

s7design:
Hi @mgruner , @dvuckovic 

Selenium tests News.t and ProductNotify.t are updated for greater stability. Tests will be finished if cloud services are disabled.

If you have any comment, please let me know.

Regards,
Milan


dvuckovic:
We probably want to preserve the check of displayed links from feeds, it's the only way to be sure fetching really works. Errors we noticed in unit tests have probably something to do with unavailability of our website server at that exact moment. Also, `CloudServicesDisabled` check is probably not needed (on by default), and we don't override it on test platforms. I'll close this for now, we'll see if there is a way to improve these tests in the future, maybe based on some sample data which is available locally.

PR #1621, opened 2017-02-20
[rel-5_0] Changed customer to customer user

urbalazs:
Hi @mgruner 
This screen is about customer user and not about customer. This was partially fixed in master with https://github.com/OTRS/otrs/commit/d540c452be8fb355b8b7ac28fdd45abe0671b131
I will create a separate pull request for master. If it is possible, it would be good to backport this into rel-4_0.


urbalazs:
Now I see, that the same issue is for Customer User-Services screen.

mgruner:
I just merged #1622 for master. To avoid string changes in rel-5, I'll close this one now. OTRS 6 is coming!

PR #1623, opened 2017-02-20
[ Rel-5_0 Bug 12628 ] - Sorting of columns in Dashboard-Ticket-Widgets is toggling with every refresh of the widget

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12628

Hi @dvuckovic 

This is our proposal for fixing reported issue. In Refresh block in AgentDashboardTicketGeneric.tt function Core.AJAX.ContentUpdate must get opposing values for described behavior.

Function 'last()' is added for the case when table is sorted by Priority because there are two columns with priority - first is default with flags and second is normal column - and second (or last) column is needed for AJAX update.

Regards,
Milan


dvuckovic:
Merged in 969f56e, 3f12199 and d7def6e, with some changes. Thanks @s7design!

PR #1630, opened 2017-02-27
Fixed pod.

rolfschmidt:
Fixed pod.


mgruner:
Hello Rolf,

you are right in that `Send()` returns an array. But if called in scalar context, this will still work, as it contains `2` in case of success. So calling in scalar context makes sense.

If we want to document the calling in list context, we should do it right and document *what* is returned. Feel free to improve this if you like!

Best regards, Martin

PR #1632, opened 2017-03-02
[ Master ] - Improved Selenium test stability for News and ProductNotify

s7design:
Hi @dvuckovic 

There are News and ProductNotify Selenium tests modified for increased stability.
In both tests Request function is overridden to prevent any cloud service instability.

Regards,
Milan

### DIFF
--- a/scripts/test/Selenium/Output/Dashboard/News.t
+++ b/scripts/test/Selenium/Output/Dashboard/News.t
@@ -12,30 +12,30 @@ use utf8;
 
 use vars (qw($Self));
 
-# get selenium object
 my $Selenium = $Kernel::OM->Get('Kernel::System::UnitTest::Selenium');
 
 $Selenium->RunTest(
     sub {
 
-        # get helper object
-        my $Helper = $Kernel::OM->Get('Kernel::System::UnitTest::Helper');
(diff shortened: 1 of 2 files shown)


dvuckovic:
@mgruner Please take a look and let me know what you think about this. For me, this is working great locally. Method for redefining subroutines follows that for changing settings during UTs, we might think about adding it to the `UnitTest.pm`.

mgruner:
I also like this. We should add a `Kernel::System::UnitTest::Helper::CustomCodeActivate()` method which takes the raw code as a parameter and works similar to `ConfigSettingChange()`. `ConfigSettingCleanup()` should then perhaps be renamed to `CustomFileCleanup()` or similar. Backport until OTRS 4 or so, I guess, depending on where the Selenium test changes are needed.

Of course, this is only needed for Selenium or HTTP based tests. For in-process unit tests, we should use Perl's built-in override methods like `local *Kernel::System::WebUserAgent::Request = sub { ... }`, right?

dvuckovic:
Merged as part of `cce0c6d`, thanks @s7design, great job! Will backport it to `rel-5_0` and `rel-4_0`, pending QA.

PR #1633, opened 2017-03-02
Fixed pod.

rolfschmidt:
Fixed pod.


mgruner:
Hello Rolf.
I agree that *Object is the more common syntax in OTRS, but in case of the Helper we actually do use `$Helper` in most places, so it is correct AFAICT. Changing that in all Selenium tests would cause quite a huge diff, I guess, and I'm not sure if that's worth it.
Best regards, Martin

PR #1639, opened 2017-03-06
[Master - Bug 12661] - In overview views wrong dates are displayed in the Created column

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12661

Hi @dvuckovic,

  Hereby is fixed the issue in small overview screens. There was a problem with wrong data for Created column. There is used logic as well as  in Medium.pm. So the code is tidied a little bit; TicketGet is called outside statement, and it not need to call it again. 

Regards
Zoran
S7design


dvuckovic:
Merged in d8d942e, 055a0a0 and d35f955. Thanks @s7design!

PR #1640, opened 2017-03-07
[ Master Bug 12657 ] - Fixed case sensitive searching for text type dynamic field values

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12657

Hi @dvuckovic

There is a proposal for fix a bug on above link.

Check for SearchCaseSensitive param and database case sensitive property are added due to SQL requests are not the same. SearchSQLGet Unit test is modified accordingly.

Need for operators is for discussion. Is text value could have just 'Like', 'Equals' and 'Empty' operators?

Regards,
Milan

### DIFF
--- a/Kernel/System/DynamicField/Driver/BaseText.pm
+++ b/Kernel/System/DynamicField/Driver/BaseText.pm
@@ -149,19 +149,13 @@ sub SearchSQLGet {
     }
 
     my $DBObject = $Kernel::OM->Get('Kernel::System::DB');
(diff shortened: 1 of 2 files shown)


s7design:
Hi @dvuckovic 

Some modifications are added.

Regards,
Milan

dvuckovic:
Merged in 3d4083e, 302c533 and 37729ba, with some small changes. Thanks @s7design!

PR #1641, opened 2017-03-08
[Master - Bug 12669] ToolBar icon 'Ticket in my Services' is shown without access to ticket

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12669

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. Implemented same logic as it is for AgentTicketQueue where it is checked for user permission in order to show count of ticket in AgentTicketService. If Config  'ViewAllPossibleTickets' is enabled, icon will be visible as will ticket in AgentTicketService.

Updated Selenium test to check for this type of behaviour and additionally improved and updated comments. 

Please let me know if there are any questions regarding this bug fix.

Regards,
Sanjin

### DIFF
--- a/Kernel/Output/HTML/ToolBar/TicketService.pm
+++ b/Kernel/Output/HTML/ToolBar/TicketService.pm
@@ -28,7 +28,7 @@ our @ObjectDependencies = (
 sub Run {
     my ( $Self, %Param ) = @_;
 
-    # check needed stuff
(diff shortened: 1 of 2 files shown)


dvuckovic:
Merged in f917b8b, 9bd1439 and 9a808c3. Thanks @s7design!

PR #1642, opened 2017-03-09
[ Master Bug 12663 ] - Fixed jumping to the first unread article in AgentTicketZoom

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12663

Hi @dvuckovic 

There is a proposal for fix reported bug on above link.
Case with only TicketID parameter (without ArticlePage and ArticleID) is modified in AgentTicketZoom PM module. First, there is searching for the latest not seen article and based on this result, page is determined. If all of articles are seen, page jump to the latest created article.
Selenium test is added to check these changes.

Regards,
Milan


dvuckovic:
Merged in 6a49b8f and 35b033a, thanks @s7design!

PR #1643, opened 2017-03-10
Fixed: ACL to restrict Status based on InputFields (TicketProperties) not working as expected

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12671

Hi @carlosfrodriguez 

There is our proposal for fixing this issue. With the fix is solved only ACL for State as it is reported in the bug, but in my opinion we should check if there is needed some changes for  other fields (e.g. Owner, Priority ... )

I suppose ACL for Services was fixed with 
https://github.com/OTRS/otrs/commit/6522d6287c9a4cfa6beb72bf8000d61c8f06f284

Your expert knowledge with ACL will be welcomed.

Regards
Zoran

### DIFF
--- a/Kernel/Modules/AgentTicketActionCommon.pm
+++ b/Kernel/Modules/AgentTicketActionCommon.pm
@@ -1351,6 +1351,7 @@ sub Run {
 
         # create html strings for all dynamic fields
         my %DynamicFieldHTML;
+        my $DynamicFieldACL;
 
         # cycle trough the activated Dynamic Fields for this screen
         DYNAMICFIELD:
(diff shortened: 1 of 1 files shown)


carlosfrodriguez:
Hi @s7design 

Thank you for your pull request, As far as I remember if you send the TicketID to the ACL (which it seams to be the case as it should be included in %Ticket) then TicketACL.pm should gather the values of the dynamic fields already. If this is not working correctly then we should fix it there.

((enjoy))

s7design:
Hi @carlosfrodriguez,

You are right for TicketID and with this Param we could get ticket data, but it is not necessary in this case, or better to say it is not needed always. 

I thought in this direction at the beginning and tried to solved this in TicketACL but it is not good solution because this is Database properties and always would be restricted some data.  In the reporter case we should restrict by customer selected properties. 

I will close this PR for now. There should be provide better solution or maybe you can improve our proposal. We fixed it in the controller because there should be restricted data on loading in any other cases AJAXUpdate subsection handle ACL restriction ( it works fine )

We also check ACL with Database properties, it works fine as well.

Regards
Zoran

PR #1644, opened 2017-03-10
Fixed: Customer Menu can be clicked through in a certain instance.

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12679

Hi @dvuckovic,

Herby is solved the issue in Safari. There is different behavior in Safari and  other browsers. Our proposal is to set z-index property for Action menu items. 

Regards
Zoran


dvuckovic:
Merged in 5b68bd5 and b9e34e6. Thanks @s7design!

PR #1645, opened 2017-03-10
[Master - Bug 12676] - Ticket delete slows exponentially with dynamic fields

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12676

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. Created new function ObjectValuesDelete() which deletes directly based on object_id param. This way we avoid iterating through out all dynamic fields when deleting ticket and causing exponential time increase with more dynamic fields in the system.

Updated unit test DynamicFieldValue.t to test new function.

Please let me know if you have any questions regarding this PR.

Regards,
Sanjin

### DIFF
--- a/Kernel/System/DynamicFieldValue.pm
+++ b/Kernel/System/DynamicFieldValue.pm
@@ -19,6 +19,7 @@ our @ObjectDependencies = (
     'Kernel::System::DB',
     'Kernel::System::Log',
     'Kernel::System::Time',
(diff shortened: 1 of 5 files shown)


dvuckovic:
Merged in f6325ca, dfbfcf4 and ab0a46b, albeit with some small changes. Thanks @s7design!

PR #1646, opened 2017-03-14
[rel-5_0] Marked strings as translatable

urbalazs:
Hi @mgruner 
I found some strings that are not marked for translation. Only rel-5_0 is affected.


mgruner:
Merged, thanks @urbalazs!

PR #1647, opened 2017-03-14
[ Master - Bug 12681 ] - In agent ticket search, profile field is not modernize.

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12681

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. Implemented 'Modernize' class for SearchProfile field. This was causing issues in past since Core.UI.Dialog.js is calling FocusFirstElement() function which is finding first field and placing focus on it. When that field is modernize, it auto-expand showing available selectable options. Included additional check to not place focus if found field is part of modernize field 'input'.

Additionally added 'id' and 'for' as parts of label for 'Search template'. This way focus is on field when label is clicked.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketSearch.pm
+++ b/Kernel/Modules/AgentTicketSearch.pm
@@ -1815,21 +1815,13 @@ sub Run {
             Base      => 'TicketSearch',
             UserLogin => $Self->{UserLogin},
         );
-        delete $Profiles{''};
-        delete $Profiles{'last-search'};
-        if ($EmptySearch) {
-            $Profiles{''} = '-';
-        }
-        else {
-            $Profiles{'last-search'} = '-';
-        }
         $Param{ProfilesStrg} = $LayoutObject->BuildSelection(
-            Data       => \%Profiles,
-            Name       => 'Profile',
-            ID         => 'SearchProfile',
-            SelectedID => $Profile,
-
-            # Do not modernize this field as this causes problems with the automatic focusing of the first element.
+            Data         => \%Profiles,
+            Name         => 'Profile',
+            ID           => 'SearchProfile',
+            SelectedID   => $Profile,
+            Class        => 'Modernize',
+            PossibleNone => 1,
         );
 
         $Param{StatesStrg} = $LayoutObject->BuildSelection(
--- a/Kernel/Output/HTML/Templates/Standard/AgentTicketSearch.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentTicketSearch.tt
@@ -15,7 +15,7 @@
         <input type="hidden" name="ShownAttributes" value="" id="ShownAttributes">
         <fieldset class="TableLike">
             <legend><span>[% Translate("Templates") | html %]</span></legend>
-            <label>[% Translate("Search template") | html %]:</label>
+            <label for="SearchProfile" id="LabelSearchProfile">[% Translate("Search template") | html %]:</label>
             <div class="Field">
                 [% Data.ProfilesStrg %]
                 <div id="SearchProfileAddBlock">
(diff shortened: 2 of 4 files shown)


s7design:
Hello @dvuckovic ,

PR is updated with proposal for fixing navigation and focus on hidden field. In case when modernize field is found in dialog as first element, new field is created with same parent before modernize field and focus is set on it. This field have position: absolute and z-index: -1 which is allowing it to remain behind modernize selection field and serve as 'dummy' focus field.

Please let me know if there are any issues or questions regarding this commit. 

Regards,
Sanjin

var/httpd/htdocs/skins/Agent/default/css/Core.Dialog.css
```diff
@@ -157,6 +157,11 @@
     width: 98%;
 }
 
+.Dialog .Field .FocusField {
+    position: absolute;
+    z-index: -1;
+}
+
```
mrcbnsls: I think there might be cases where just having a z-index of -1 is not enough, e.g. when the element which is on top of the "hidden" element has no background color. In that case the "hidden" element would still be visible in a weird way. We should better add positioning, too (similar to ``.InvisibleText``, ``left: -9999px`` and ``top: -9999px``).

var/httpd/htdocs/js/Core.UI.Dialog.js
```diff
@@ -88,10 +88,17 @@ Core.UI.Dialog = (function (TargetNS) {
...
-        $('div.Dialog:visible .Content')
-            .find('a:visible, input:visible, textarea:visible, select:visible, button:visible')
-            .filter(':first')
-            .focus();
+        var FirstElement = $('div.Dialog:visible .Content').find('a:visible, input:visible, textarea:visible, select:visible, button:visible').filter(':first'),
+            FocusField;
+
+        // Check if found first element is modernize input field, prepend input field which is behind modernize field and set focus on it allowing navigation.
+        if (FirstElement.closest('.Field').find('.Modernize').length) {
+            FocusField = FirstElement.parent();
```
mrcbnsls: I didn't try the code out, but it seems wrong or at least unnecessarily complex to me. If a ``.Field`` container contains more than one field of which one is of type ``.Modernize``, the hidden field would be added without any need. Why not just check if ``FirstElement.hasClass('InputField_Search')``?

Kernel/Output/HTML/Templates/Standard/AgentTicketSearch.tt
```diff
@@ -15,7 +15,7 @@
         <input type="hidden" name="ShownAttributes" value="" id="ShownAttributes">
         <fieldset class="TableLike">
             <legend><span>[% Translate("Templates") | html %]</span></legend>
-            <label>[% Translate("Search template") | html %]:</label>
+            <label for="SearchProfile" id="LabelSearchProfile">[% Translate("Search template") | html %]:</label>
```
mrcbnsls: What's the use of the ``id`` here? I think ``for``should be enough, no?

Kernel/Modules/AgentTicketSearch.pm
```diff
@@ -1815,21 +1815,13 @@ sub Run {
...
             UserLogin => $Self->{UserLogin},
         );
-        delete $Profiles{''};
-        delete $Profiles{'last-search'};
-        if ($EmptySearch) {
-            $Profiles{''} = '-';
-        }
-        else {
-            $Profiles{'last-search'} = '-';
-        }
```
mrcbnsls: Is this part no longer needed?

dvuckovic:
Merged in d8fe430 and 48f0009. Thanks to @s7design for pull request and @mrcbnsls for insights.

PR #1649, opened 2017-03-15
[ Master bug#12684] The format buttons are missing from the stacked area chart on the dashboard if the language is not English

s7design:
Hi @dvuckovic,

Herby is solved translation of the controls labels in stats. There is used Core.Config.Get instead of Core.Language.Translate because  Dev::Tools::TranslationsUpdate doesn't check thirdparty  JavaScript files.
There is also translate Format name in preview screen. 

Regards
Zoran

### DIFF
--- a/var/httpd/htdocs/js/thirdparty/nvd3-1.7.1/models/OTRSLineChart.js
+++ b/var/httpd/htdocs/js/thirdparty/nvd3-1.7.1/models/OTRSLineChart.js
@@ -45,7 +45,7 @@ nv.models.OTRSlineChart = function() {
 // OTRS
 // ---
 //       , noData = 'No Data Available.'
-        , noData = Core.Language.Translate('No Data Available.')
+        , noData = Core.Config.Get('NoDataAvailable')
 // ---
         , dispatch = d3.dispatch('tooltipShow', 'tooltipHide', 'stateChange', 'changeState', 'renderEnd')
         , duration = 250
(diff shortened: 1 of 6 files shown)


var/httpd/htdocs/js/thirdparty/nvd3-1.7.1/models/OTRSLineChart.js
```diff
@@ -45,7 +45,7 @@ nv.models.OTRSlineChart = function() {
 // OTRS
 // ---
 //       , noData = 'No Data Available.'
-        , noData = Core.Language.Translate('No Data Available.')
+        , noData = Core.Config.Get('NoDataAvailable')
```
dvuckovic: @zottto We would need to go back to old API for translations in this case. The reason is that `TranslationsUpdate` console command will not mark strings translatable for JS in thirdparty files, only core. What do you think, is this acceptable? Translations would be set by Perl `AddJSData()`, of course.

mrcbnsls: @dvuckovic, as of today, the TranslationsUpdate command will also scan 3rd party files (which contain custom markers), so we should go back to use Core.Language.Translate there.

s7design:
Hi @dvuckovic 

This PR is rebased because of changes that has been done on TranslatUpdate. There was left only translate in Preview screen for Format name of graph stats. 

There is another PR for OTRS5.

Regards
Zoran

dvuckovic:
Merged in 4e20d01, thanks @s7design!

PR #1650, opened 2017-03-15

urbalazs:
Hi @mgruner 
I found two hardcoded strings, that are not marked for translation. Both branch rel-5_0 and master are affected.


mgruner:

PR #1651, opened 2017-03-15
[ Master - Bug 12683 ] - Fixed difference between key and content in EventsTicketCalendar

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12683

Hi @dvuckovic 

There is proposal for reported bug fix. Some keys are taken and its values are modified in accordance with logic used in AgentTicketStatusView screen as reporter wrote.

Regards,
Milan


dvuckovic:
Merged in e8d6456, ae37abc and 33f854a. Thanks @s7design!

PR #1653, opened 2017-03-15
Add column to show pending date (not hours till pending time)

reneeb:


mgruner:
Hi @reneeb. While this is an interesting feature addition, I don't agree with how it is implemented. Here an undocumented ticket parameter with the name `NotUsed` is used. To solve this properly we'd have to change the Ticket API probably. I'll close this PR therefore. Please feel free to make additional suggestions or changes.

PR #1654, opened 2017-03-16
Added config option for the RichText width/height

straubr:
The .pm file looks for these values in the config (as do all the others), but for this view the options were missing in the config.


mrcbnsls:
Nice, thanks. Merged in b793bdb.

PR #1656, opened 2017-03-17
[ Master Bug 12611] - In Overview the 'Order by' and the 'Move' buttons are old style

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12611

Hello @dvuckovic ,

Reporters issues could not be resolved because of the current technical limitation. Nevertheless while investigating this issue, included modernize input field for 'Reply' in Overview screens and Customer history tables.

Please let me know if there are any questions regarding this PR.

Regards,
Sanjin


dvuckovic:
Merged in 246a495 and 7113936, thanks @s7design!

PR #1657, opened 2017-03-17
[ Master - Bug 12695 ] - Fixed collapse/expand event after submit table setting

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12695

Hi @dvuckovic 

There is a proposal for reported bug fix. Toggle event has to be loaded after table redrawing.

If you have any comment please inform me.

Regards,
Milan


dvuckovic:
Merged in 4a1fe15 and 1aa0e90, thanks @s7design!

PR #1658, opened 2017-03-18
Improved wording for the redesigned preferences

urbalazs:
Hi @mgruner 
There is no longer "Update" button in the redesigned User preferences, only a save button with a tick mark.


mrcbnsls:
Thanks, merged!

PR #1659, opened 2017-03-20
[ Rel-5_0 bug#12684] The format buttons are missing from the stacked area chart on the dashboard if the language is not English

s7design:
Hi @dvuckovic,

Herby is solved translation of the controls labels in stats for OTRS5. There is also translate Format name in preview screen.

Regards
Zoran


dvuckovic:
Merged in 2cb9eb7, thanks @s7design!

PR #1660, opened 2017-03-21
Fixed file input overflowing

urbalazs:
Hi @mrcbnsls 
This is a proposal for fixing the file input overflow when importing a web service (in case of a Hungarian locale).

**Before:**
overflow-before

**After:**
overflow-after

If this proposal can be merged, please backport it also into the OTRS 5.


mrcbnsls:
Thanks, merged in master and rel-5_0. Do you think there are other places which should also be fixed regarding this issue?

urbalazs:
We found only this one. The other file inputs are in the left column or in the main column.

PR #1663, opened 2017-03-21
[Master Bug 12703 ] Fixed: dynamic list and dynamic matrix interpret agent/user Userfirsname

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12703
Hi @dvuckovic,

Hereby is fixed the bug. There is also fixed comment for function UserName, that is used in the fix. There was wrong param in example UserLogin insted User. In this function is called GetUserData which expect User or UserID.

Lock param in list of ticket attributes is also changed in upper case first latter. 

Regards
Zoran


dvuckovic:
Merged in 82c227a, 8e328a3 and 16d13e0. Thanks @s7design!

PR #1664, opened 2017-03-21
[ Master - Bug 12704] - Ticket age values in statistic export is in seconds (not human readable)

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12704

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. Modified 'Age' param in statistics from seconds to format ( 100 d 12h ) which is applied in different OTRS screens (e.g. Overview's, AgentTicketZoom...).

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentStatistics.pm
+++ b/Kernel/Modules/AgentStatistics.pm
@@ -1008,18 +1008,6 @@ sub RunAction {
         };
     }
 
-    # Change ticket param 'Age', if selected, to human readable format see bug#12704.
-    my $Age = $LayoutObject->{LanguageObject}->Translate('Age');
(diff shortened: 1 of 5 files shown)


s7design:
Hello @dvuckovic ,

Removed code from 1st commit as we discussed and copied CustomerAge() to Stats.pm as function without translatable strings. Updated Result.t UT to check for new type of 'Age' column output and sorting by that column is done as well.

Please let me know if there are any issues or questions regarding this commit.

Regards,
Sanjin

dvuckovic:
Merged in f1b8c79, pending QA. Thanks @s7design!

PR #1665, opened 2017-03-21
[rel-5_0] Removed obsolate translatable strings

urbalazs:
Hi @mgruner 
These dummy files is no longer needed, as all the affected strings are marked as translatable somewhere in the source code (I checked all of them).


mgruner:
Hello @urbalazs and thanks for your proposal. These files have been removed in master and thus all future versions, that's enough  I believe.

PR #1667, opened 2017-03-22
[ Master - Bug 12716 ] - Activity dialog causes an Error in customer interface.

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12716

Hello @dvuckovic ,

Hereby is fix for reporters issue. Solving bug #12433 added code which check for activities with ActivityGet() function for which in this case (  CustomerTicketProcess.pm ) we need param Interface to match CustomerInterface.

Please let me know if there are any questions regarding this PR.

Regards,
Sanjin


dvuckovic:
Merged in e30e5ca, c0c0692 and 7545ce2. Thanks @s7design!

PR #1668, opened 2017-03-23
[ Master - Bug 12718 ] - There is no check for the similar names when creating a mail filter 

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12718

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. Added check for existing PostMasterFilters on create or edit action, triggering error when check is matched. Updated Selenium test to check for this behaviour and additionally improved comments on it to match new style.

Regards,
Sanjin


dvuckovic:
Merged in 03c80ed, 48c3d8d and 27044c5. Thanks @s7design!

PR #1669, opened 2017-03-23
[Master Bug 12649] - The long and short description of the process ticket are not displayed at the first activity

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12649

Hi @dvuckovic,

This is our solution for adding long description on initial dialog for Activity Process. We have chosen this simple one but we are open for any other suggestions from our UI Experts.

Regards
Zoran


dvuckovic:
@mrcbnsls What do you think about this approach? Currently, long description is not used anywhere in the process front-end.

Since this change might be desirable only in `master`, the idea for `rel-5_0` is to try to fix it by better explaining the fields in admin interface (`FieldExplanation`).

mrcbnsls:
Hi @dvuckovic and @s7design,

thanks for the suggestions. I honestly don't think adding the long description to the title of the short description is appropriate because it could contain multiple lines of text which might get cut in some browsers (e.g. in Firefox). Also, it is IMO not very likely that users understand they need to hover over the short description in order to read the long description. I'd suggest displaying the long description as a separate widget below the process information widget:

bildschirmfoto 2017-03-27 um 09 53 29

@dvuckovic, I'm not sure I understand your suggestion for rel-5_0. Would you explain to the admin who's creating the process that the long description is not shown anywhere? :-D I think it could be okay to backport the new widget also to rel-5_0, don't you?

mgruner:
@mrcbnsls, displaying a hint was my suggestion for a fix for the original bug: it’s not clear what the fields are actually meant for, so IMHO the fix should address that issue. 

Adding new widgets is possible as a feature enhancement, but I would be careful with back porting that, as this will cause information to be shown where it was not shown before.

Martin Gruner
Team Lead R&D

OTRS AG
Bahnhofplatz 1a
94315 Straubing

T: +49 (0)6172 681988 0
F: +49 (0)9421 56818 18
I:  www.otrs.com/

Integrierter Kalender, komplett verschlüsselte Kommunikation und vieles mehr! Die neue OTRS 5s mit Features für OTRS Free, OTRS Business Solution™ und OTRS Business Solution™ Managed ist jetzt verfügbar! 

Mehr erfahren: https://www.otrs.com/otrs-5s/?lang=de&utm_id=569

mrcbnsls:
@mgruner, I agree on giving more information on the purpose of each field to the user, but what would be shown for the description long field in rel-5_0 when it's not displayed anywhere?

dvuckovic:
@mrcbnsls I agree with you regarding the visibility, it's much better to have it as a separate block of text, instead of a tooltip.

But please note that this exact behavior is applicable for initial process activity only. Every other activity once process is started is actually being displayed in a popup window:

screen shot 2017-03-27 at 10 21 58

Short description is then concatenated to the activity name, and long description is output without line breaks in line below. So we do actually show it right now, but only if activity is not the first one.

@mgruner While I agree that showing new information might be undesirable in OTRS 5, the bug report is essentially about the missing info. And it is being used in other places, so I think it's valid to try to display it on first activity too.

s7design:
Hi @dvuckovic 

PR is modified in accordance with suggestions. If you have any comment please let me know.

Regards,
Milan

PR #1670, opened 2017-03-23
Update Defaults.pm

nilei:
LDAP customer user data backend is read only.


mgruner:
Applied to master, thanks @nilei!

PR #1671, opened 2017-03-23
[Master Bug 12719] - The result of SQL box displays the unique column headers not with the right character encoding

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12719
Hi @dvuckovic 

Herby is fixed encoding name of the columns names of a database statement. Unit test is updated as well. It is our proposal to fix header names in results table in SQL Box screen.

Regards
S7Design
Zoran

### DIFF
--- a/Kernel/System/DB.pm
+++ b/Kernel/System/DB.pm
@@ -828,10 +828,10 @@ to retrieve the column names of a database statement
 sub GetColumnNames {
     my $Self = shift;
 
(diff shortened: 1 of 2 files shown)


mgruner:
👍

dvuckovic:
Merged in 16d3f07, 84aeedd and 4bd87ff. Thanks @s7design!

PR #1673, opened 2017-03-23
[Master Bug - 12696] - Missing save button from personal preferences in mobile mode

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12696

Hi @dvuckovic 

Hereby is fixed issue with update buttons in AgentPreferences. There is style in Core.Agent.Admin.SystemConfiguration.css for class SettingUpdateBox which set property display to 'none'.

>     .WidgetSimple.Setting.IsLockedByMe .Content .SettingUpdateBox {
>         display: none !important;
>     }

So there is added new class for the block element SettingUpdateBox in AgentPreferences. 

Regards
Zoran


s7design:
Closing, there is no good fix. Look the comment on bug report.

Zoran

PR #1676, opened 2017-03-27
[ Rel-5_0 Bug 12720 ] - Settings window of Complex LinkObject is not translated

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12720

Hello @dvuckovic ,

Hereby is fix proposal for reporters issue in Rel-5_0. Implemented slightly different approach then how it is proposed for master fix. Moved calling function Core.Agent.TableFilters.SetAllocationList() from LinkObject.js to TicketZoom.js, this way we get correctly translated configs-columns when page is loaded. Similar approach is used for dashboard modules.

Regards,
Sanjin


s7design:
Hello,

Fix is applied. Closing this PR.

Regards,
Sanjin

PR #1677, opened 2017-03-28
[Master - Bug 12726] -  Automatic links to CVE numbers don't work in TicketViewZoom.

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12726

Hi,

Herby is fixed an issue with replacing CVE number in article body. There is changed URL to  favicons for US-CERT NVD. 
Unit test is updated with the cases which check if there is links in article body and if there is images for the links that is set in CVE sysconfig item. 

Regards
Zoran


s7design:
Hello,

Regards,
Milan

PR #1678, opened 2017-03-28
Swich to https

urbalazs:
Hi @mgruner 
If an error occurs in OTRS, and the user wants to send the bug report, he will be informed, that the connection is not secure, a third party may see the submitted data. This patch avoids this warning message.

I looked for "http" in the source code, and there are many comments that contains "http" instead of "https" (e. g. all copyright text in the beginning of the files, or many places in the CHANGES.md). Maybe it would be good to change these occurrences, too.

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/Error.tt
+++ b/Kernel/Output/HTML/Templates/Standard/Error.tt
@@ -32,7 +32,7 @@
                 </div>
                 [% END %]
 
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Thanks, applied for 4, 5 and master.

We'll discuss the copyright change, too, thanks for pointing this out.

PR #1680, opened 2017-03-29
[ Master - Bug 12735 ] - Console does not allow adding of sub services when their name exists as a service

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12735,

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. In console command Admin::Service::Add, there was just check every time if option 'name' exists as service, regardless if it's used as subservice. Updated UT to check for these types of behaviour, where we create in multiple sublevels with same name for parent and child.

Regards,
Sanjin


s7design:
Hello,

Closing this PR. Fix is in appropriate bug fix branches.

Regards,
Sanjin

PR #1682, opened 2017-03-30
[ Rel-5_0 Follow-up Bug 12720 ] - Settings window of Complex LinkObject is not translated

s7design:
Hello @dvuckovic ,

Hereby is follow-up for bug fix. With this PR we make LinkObject agnostic about column names. Please let me know if there is any questions or issues regarding this PR.

Regards,
Sanjin


dvuckovic:
Merged in 3c32513, thanks @s7design!

PR #1684, opened 2017-03-30
Update AgentTicketZoom.tt

nilei:
Reposition of article dynamic fields. The current place between From/To and Subject felt wrong and misplaced.


mrcbnsls:
Thanks, merged in 9e995a8.

PR #1685, opened 2017-03-30
[Master Bug 12734] - Dropdown not considered into TicketSolutionResponseTime Statistic 

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12734

Hi @dvuckovic 

Hereby is fixed problem with the issue in TicketSolutionResponseTime stats. Unit test with the test cases for  function GetStatElement from TicketSolutionResponseTime.pm is added as well.

Regards
Zoran


skirsdeda:
+1 Actually, TicketSolutionResponseTime stat ignores all DynamicField search attributes due to that extra "_" in regex.

PR #1686, opened 2017-03-30

urbalazs:
Hi @mgruner 
I found two error messages, that are not marked for translation. Both rel-5_0 and master are affected.


mrcbnsls:
Thanks, merged!

PR #1689, opened 2017-03-31
Improved AttachmentDelete button

urbalazs:
Hi @mrcbnsls 
This button is not modern. Both branch rel-5_0 and master are affected.


mrcbnsls:
Thanks for pointing this out. There are many more places with attachment delete buttons though. Also, there is code in Core.AJAX.js which is capable of loading templates with attachments which also had the old button style.

I made bigger changes for master in 61e4b0f which include your changes. As this includes also changes in the placements of the buttons, I don't want to backport it to rel-5_0 for now. If you would be willing to add another PR which fixes only the button style in rel-5_0 (if possible in all places, see the list of affected files in 61e4b0f for reference), that'd be great!

Marc

PR #1692, opened 2017-04-03
Fixed bug#12743 - ACLs with ticket conditions are not matching on process list in AgentTicketProcess.

rolfschmidt:
https://bugs.otrs.org/show_bug.cgi?id=12743


dominikklein:
Hi @rolfschmidt 

thanks for the pull request :) this will be fixed in the next days.

Best regards,
Dominik

PR #1693, opened 2017-04-03
Marked error message as translatable

urbalazs:
Hi @mrcbnsls 
There is an error message, that is not marked as translatable. I don't know, what would be the best solution for these kind of strings, because it should be translated in the error message (underlined by red), but it should leave untranslated in the Error details (underlined by green). The first one is a user interface, the latter is a log entry. Both branch rel-5_0 and master are affected.

error-message

### DIFF
--- a/Kernel/System/Ticket/Article.pm
+++ b/Kernel/System/Ticket/Article.pm
@@ -20,6 +20,7 @@ use Kernel::System::VariableCheck qw(:all);
 
 our @ObjectDependencies = (
     'Kernel::Config',
+    'Kernel::Language',
     'Kernel::System::Cache',
     'Kernel::System::CustomerUser',
     'Kernel::System::DB',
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @urbalazs, thanks for your PR. I agree with you that this needs to be separated in some way. But though I agree that it would be nice if all messages would be translated everywhere in the system, I think it's acceptable (even for a professional software) to have some messages for hard errors (which probably don't show up too often if the system is configured properly) in english, only.

For messages which can appear to the user, too, on a regular basis (which I don't know exactly for the message which is the subject of this PR), the ```Notify()``` method should be used in addition to a log entry. It's unfortunate that this is mixed up here.

Regarding this PR: I don't think it solves the situation properly (which you probably agree on). This is why I'd close it and postpone this discussion to a later point. I think this would be something that should be clarified with Martin, too.

PR #1694, opened 2017-04-04
[rel-5_0] Improved buttons

urbalazs:
Hi @mrcbnsls 
This pull request is only for branch rel-5_0. In the branch master this is already done.


mrcbnsls:
Thanks, merged in 9e81165. I saw that two buttons were forgotten which only appear when the system is already registered and data sending is enabled. I updated them, too.

PR #1695, opened 2017-04-04
Show old type in ticket history view

moritz:
it has been logged before, but simply not shown in AgentTicketHistory


mrcbnsls:
There have been some wording changes in the meantime, so I took your proposal and changed it to be in line with those changes: 711040b53a - thanks!

PR #1696, opened 2017-04-04
[rel-5_0] Improved delete buttons

urbalazs:
Hi @mrcbnsls 
Here is the patch for branch rel-5_0 to improve the delete buttons, as you requested in https://github.com/OTRS/otrs/pull/1689


mrcbnsls:
Thanks, merged!

PR #1697, opened 2017-04-04
[ Master - Bug 12745 ] - AgentTicketSearch missing 'Extended' param in Article and Ticket get methods

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12745

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issue. There was missing param Extended for both Article and Ticket get methots. This param is required in order to call _TicketGetExtended() and in that function there is _TicketGetClosed()  function which returns "collected attributes of (last) closing for given ticket" from ticket_history table. With this fix, closed tickets with or without any articles will show 'Closed' data in normal, CSV or Excel output.

Regards,
Sanjin

### DIFF
--- a/Kernel/Modules/AgentTicketSearch.pm
+++ b/Kernel/Modules/AgentTicketSearch.pm
@@ -680,6 +680,7 @@ sub Run {
                     # get ticket data instead
                     %Data = $TicketObject->TicketGet(
                         TicketID      => $TicketID,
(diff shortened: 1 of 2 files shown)


s7design:
Hello,

Fix is applied in appropriate bug fix branches. Closing this PR.

Regards,
Sanjin

PR #1700, opened 2017-04-06
[master] Added icon to button

urbalazs:
Hi @mrcbnsls 
An icon is missing from the System Registration button. I also changed the existing one, because this icon is used in other places for the registration buttons.


mrcbnsls:
Merged, thanks!

PR #1704, opened 2017-04-10
[ Master - Bug 12744 ] - Responsible Filter in Queue View does not work

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12744

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issues. Responsible field is changed to be autocomplete as it was before in overview screens.

Please let me know if there is any issues or questions regarding this PR.

Regards,
Sanjin


dvuckovic:
Merged in 159ec6e and 3140250, thanks @s7design!

PR #1707, opened 2017-04-10
Fixed: The format buttons are displayed on the dashboard translated, but during the running of stats they are not translated

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12756

Hi @dvuckovic 

Herby is solved translation for action in SVG screens for stats.
It is solved only in rel-5_0, because in master there is used in js Core.Language.Translate.

regards
Zoran


mrcbnsls:
Thanks, merged!

PR #1709, opened 2017-04-12
[ Master - Bug 12736 ] - Fixed Internal Server Error when signature part misses

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12736

Hi @dvuckovic 

This is a proposal for bug fix. Error was appeared because when signature missed, undefined parts of Entity called "as_string" and "body_as_string" functions. Modification of IF clause solve this wrong calls.

If you have any comment please inform me.

Regards,
Milan


dvuckovic:
Merged in a47271d, 9cce6d6 and f8caa3b. Thanks to Michael and @s7design!

PR #1711, opened 2017-04-13
[Master Bug 12752] Regular expression for dynamic field also affects search

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12752

Hi @dvuckovic 

Hereby is solved an issue with regular expression for text and Textarea dynamic fields. It is our proposal, we can discuss about that.

Regards
Zoran


s7design:
Hello,

Regards,
Milan

PR #1713, opened 2017-04-18
Unified error messages and marked as translatable

urbalazs:
Hi @mgruner 
I noticed, that when adding new items (group, role, SLA, service, etc.) the error messages are different when the name is already exists. I can unified the existing error messages, but the following items have no error message, the SQL error is displayed for the user (both branch rel-5_0 and master are affected):

Attachment:
kep

Salutation:
kep

Signature:
kep

ACL (error screen instead of ribbon):
kep

State:
kep

Priority:
kep

Service (missing name):
kep

**EDIT**: Bug report for this issue: https://bugs.otrs.org/show_bug.cgi?id=12778

### DIFF
--- a/Kernel/System/ACL/DB/ACL.pm
+++ b/Kernel/System/ACL/DB/ACL.pm
@@ -157,7 +157,8 @@ sub ACLAdd {
     if ($ACLExists) {
         $Kernel::OM->Get('Kernel::System::Log')->Log(
             Priority => 'error',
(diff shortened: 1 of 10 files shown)


mgruner:
Hello @urbalazs. We typically don't translate system log messages, also because there is not always a properly initialized Layout/Language object available. Could you please consider resending this PR to not include the translation part?

PR #1715, opened 2017-04-18
Added support for Perl modules on Freebsd

felixjogris:
Hi,

attached patch adds support for FreeBSD ports to bin/otrs.CheckModules.pl. The FreeBSD ports system compiles latest software from scratch (it was the inspiration for Gentoo's portage).
A default install of OTRS 5.0.18 on FreeBSD is missing these Perl modules:

Excel::Writer::XLSX
Lingua::Translit
Pod::Strip
HTML::Truncate
Schedule::Cron::Events

Thus, I added extra checks for them as well, and notified the maintainer of the OTRS port.

BR,
Felix


mgruner:
Hello Felix, the modules you mention are all bundled in Kernel/cpan-lib, except for HTML::Truncate, which is no longer needed. I will add your BSD logic though, to aid people installing on that platform. Thanks a lot!

mgruner:
Applied with some changes in master in a4af0004aaaac7fc91ad290ff44a8a2d8721ec86. Thanks a lot @felixjogris!

PR #1717, opened 2017-04-20
 contribute-TransitionAction CallFunction 

johannesbauer1:
Allows you to call a os script with Arguments.
First Option in the Settings Page must be named "FunctionName" and the value is set to the path of the script.
The transitionaction will always add as the first argument the ticketnumber, as the second the ticketid and then the ones specified in the settingspage.
The Options are ordered alphabetically by the value you did set as key.


mgruner:
Hi @johannesbauer1. Thanks a lot for your proposal. I currently don't see this as a feature that would be relevant for many customers, so I hesitate to merge it at this point. We'd also need to apply some changes to the code. If you plan more contributions, please also consider signing our CLA as described on http://doc.otrs.com.

PR #1719, opened 2017-04-21
[ Master - Bug 11028 ] - Owner and responsible always pre-selected in initial process screen

s7design:
https://bugs.otrs.org/show_bug.cgi?id=11028

Hello @dvuckovic ,

Hereby is proposal for fixing reporters issues. Had to add empty selection since it was not possible to handle it via BuilsSelection() - PossibleNone param. Removed also part in which by default if there is no ticket or default values current user is pre-selected.

Regards,
Sanjin


s7design:
Hello,

Fix is applied in appropriate bug fix branch. Closing this PR.

Regards,
Sanjin

PR #1720, opened 2017-04-21
[ Rel-5_0 - Follow-up Bug 12773 ] - In case of big screen resolution the Subject field is too long

s7design:
https://bugs.otrs.org/show_bug.cgi?id=12773

Hello @mrcbnsls ,

Hereby is just short fix for rel-5_0 in which max-width for Subject field is increased so it match default RichText editor width.

Regards,
Sanjin


mrcbnsls:
Merged in acdfca6245, thanks. Should fit better now.

PR #1729, opened 2017-04-25
[ Master ] - Improved Selenium test CustomerAutoCompletion

s7design:
Hello @dvuckovic ,

Hereby is updated Selenium test in order to increase stability of UT. Removed parts where we are checking input values in dialog. IMO this is redundant information since we are later clicking on autocomplete values, which will not be available if input is incorrect. This change is result from manual testing and it happened in few occasions that field lose value after these checks.

Additionally changed some find_element, to try and prevent losing element and getting this error "An element command failed because the referenced element is no longer attached to the DOM.: Element is no longer attached to the DOM" 

Please let me know if there are any questions or issues regarding this PR.

Regards,
Sanjin


dvuckovic:
Merged in 5415782, thanks @s7design!

PR #1732, opened 2017-04-26
[ Rel-5_0 - Bug 11243 ] - Wrong escalation time if response time set to 60 or less minutes

s7design:
https://bugs.otrs.org/show_bug.cgi?id=11243

Hi @dvuckovic 

It is a proposal for fix of reported bug which does not exist on master branch because of Time object changes.
Bug appeared every time when escalation time is set to less than 60 minutes and is greater than the rest of a day working time. Because of that, IF clause is modified.
Unit test Escalations.t is extended for inspecting of described case.

Regards,
Milan


s7design:
Hello,

Regards,
Milan

PR #1735, opened 2017-04-27
[ Master - Bug 10681 ] - Fixed wrong FieldOrder default value at Dynamic Field creation

s7design:
https://bugs.otrs.org/show_bug.cgi?id=10681

Hello @dvuckovic,

It is a proposal for reported bug fix. By default, when dynamic fields were created, they were added to the bottom of the page, not to the bottom of the list. Because of that, MaxFieldOrder value for 'MaxFieldOrder' block is changed.

Selenium tests are modified. Tested dynamic field was deleted by backend function after re-navigating to DynamicField overview screen and because of that default FieldOrder value was not correct in the next adding screen.

Also, check for default FieldOrder value is added to Selenium test AdminDynamicField.t.

If you have any comment about this PR, please let me know.

Regards,
Milan

### DIFF
--- a/Kernel/Modules/AdminDynamicField.pm
+++ b/Kernel/Modules/AdminDynamicField.pm
@@ -313,8 +313,6 @@ sub _DynamicFieldsListShow {
         );
     }
 
-    my $MaxFieldOrder = 0;
-
     # check if at least 1 dynamic field is registered in the system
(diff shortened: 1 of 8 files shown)


s7design:
Hello,

Regards,
Milan

PR #1737, opened 2017-04-28
Fixed: I can't configure SMTP TLS of Office 365.

s7design:
Hi @dvuckovic,

Herby is updated CheckModules.pl with new modules that could be required if Mail::IMAPClient is installed and used. 

There is added only Authen::SASL and Digest::MD5. I am not sure if Digest::HMAC_MD5  and/or Authen::NTLM are also required. 

I made only master branch at this moment. After your follow-up I will be able to make the same for rel-5 and rel-4.

Regards
Zoran

### DIFF
--- a/bin/otrs.CheckModules.pl
+++ b/bin/otrs.CheckModules.pl
@@ -362,6 +362,26 @@ my @NeededModules = (
                     zypper => 'perl-IO-Socket-SSL',
                 },
             },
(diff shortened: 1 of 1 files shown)


dvuckovic:
Merged in 38f6879, fbc4d82 and 2b5d7eb. Thanks @s7design!

PR #1739, opened 2017-04-28
Inconsistent naming of the BackendObject

straubr:
All the other functions use $BackendObject, but the new-POD description declares it as $DynamicFieldObject (which collides with the actual DynamicFieldObject).


mgruner:
Fix applied, thanks a lot @straubr!

PR #1740, opened 2017-04-30
[master] Fixed typos

urbalazs:
Hi @mgruner 
I found some typos. This PR is only for branch master, I will create an other for rel-5_0.

PR #1741, opened 2017-05-01
[rel-5_0] Fixed typos

urbalazs:
Hi @mgruner 
This is for branch rel-5_0. For branch master please see https://github.com/OTRS/otrs/pull/1740


mgruner:
Applied with small fix (into an another), thanks @urbalazs!

PR #1742, opened 2017-05-03
[Master - Bug 12801] Set Role validity state to invalid, doesn't effect permissions.

s7design:
Hi @dvuckovic 

Hereby is solved issue with permission by role if role is set on invalid. There was a problem here
https://github.com/OTRS/otrs/blob/master/Kernel/System/Group.pm#L1933 for the reported issue. 

However, I fixed it on the another two place in Group.pm. I checked the test Role.t and AdminRole.t. These changes doesn't have influence on them. 

Regards
Zoran

### DIFF
--- a/Kernel/System/Group.pm
+++ b/Kernel/System/Group.pm
@@ -1652,7 +1652,7 @@ sub PermissionGroupRoleGet {
     }
 
     # get valid role list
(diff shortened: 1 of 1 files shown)


s7design:
Hello,

Regards,
Zoran

PR #1743, opened 2017-05-03
[Master Bug 12802] ACL -> Process restriction does not work with a Queue Filter

s7design:
Hi @dvuckovic 

This is proposal for fixing reported issue. There was missing ticket data for TicketACL if it is needed check database data.
In the report is mentioned restriction by Queue but the same problem is with other thicket properties ( State, Priority ... )

Regards
Zoran


s7design:
Hello,

Closing this PR. Bug is marked as duplicate of https://bugs.otrs.org/show_bug.cgi?id=12743 and fix is applied in appropriate bug fix branch.

Regards,
Sanjin

PR #1747, opened 2017-05-04
[ Rel-5_0 ] - JS improvements for multiple LinkTable widgets.

s7design:
Hello @dvuckovic ,

Hereby is proposal for fixing JS issues when having multiple different link object tables. Sending new parameter to Core.Agent.TableFilter.SetAllocationList() function. With this param we then on 'Save changes' on widget preference setting update just that specific widget excluding other from the update which resulted in multiplication of allocation lists.

Additionally, added off / on when creating events for registering toggle between containers.

Regards,
Sanjin


dvuckovic:
Merged in 848a5c6, thanks @s7design!

PR #1749, opened 2017-05-08
Improved wording

urbalazs:
Hi @mgruner 
In Hungarian language usually the parameter is before the noun (e. g. the English "Page 3" is "3. oldal" in Hungarian), and there is no plural form when a number is before a noun (e. g. the English "1 Article", "2 Article**s**", ... are "1 bejegyzés", "2 bejegyzés", ... in Hungarian).
I sent this pull request, because the current wording makes it impossible to correctly translate the whole sentence into Hungarian, because the order of the words are fixed. With placeholder parameters it would be possible to make the correct Hungarian wording.
All supported versions (rel-4_0, rel-5_0 and master) are affected.

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AgentTicketZoom.tt
@@ -251,7 +251,7 @@
         [% IF Data.ZoomTimeline %]
             [% Translate("Ticket Timeline View") | html %]
         [% ELSE %]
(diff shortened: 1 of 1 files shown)


mgruner:
Merged for master, thanks @urbalazs. I would like to avoid a back port to not require retranslation, is that acceptable for you?

urbalazs:
Thank you for the merge @mgruner. If you back port to rel-5_0 and make a translation sync to Transifex, then translators have time to correct it. Since rel-4_0 has no translation update, it is not so important at the moment (because the upcoming OTRS 6 will make it unsupported).

PR #1758, opened 2017-05-12
Allow to send email to system addresses

gravasio:
In older OTRS versions was possible to answer and forward a ticket to a system address.
In OTRS5 this is forbidden and an error message is given to the user.
Actually our workflow needs to be able sometimes to put in CC a system address.

This patch adds a setting in Ticket::Frontend::AgentTicketCompose allowing the user to disable system address protection.

Thanks
Giuseppe


mgruner:
Hi Guiseppe, thanks for your proposal. The PR seems to be broken, can you rebase your work  please on master? It could be caused by an inconsistent state that we had some time ago.
To accept contributions from you, I would have to ask you to sign our cla as described on http://doc.otrs.com, please.
In this case I am not sure about the merging, however. OTRS would catch all followups with the "References" follow-up detection module. Did you turn that off?
Best regards, Martin

mgruner:
Hello @gravasio, I'm closing this PR for now. Please feel free to follow up with more information!

PR #1759, opened 2017-05-13
[rel 5_0] Avoid admin to leave OTRS

urbalazs:
Hi @mgruner 
This proposal makes it possible not to leave OTRS by clicking the external links.


mgruner:
Applied to OTRS 5, thanks @urbalazs! @balazj, can you please add this also to master in your branch?

PR #1762, opened 2017-05-18
Follow-up for PR1740 - fixed punctuation

urbalazs:
Hi @mgruner 
This is a follow-up for https://github.com/OTRS/otrs/pull/1740 (master) and https://github.com/OTRS/otrs/pull/1741 (rel-5_0).
The superfluous dot is also present is these files. Both branch rel-5_0 and master are affected.


mgruner:
Applied in 7187fcc558ec543fdd534b62fa09ae1a38552a65, not relevant for master any more. Thanks @urbalazs!

PR #1765, opened 2017-05-23
Console Script UserUnLink.pm

jensoliver:
Created Console Script UserUnLink.pm


mgruner:
Hi @jensoliver, thanks a lot, but we're not adding new features to OTRS 5 any more.

PR #1770, opened 2017-06-09
Useless logical expression

k-mx:
More elegant construction without useless '!defined $CustomQueueIDs[0]'


mgruner:
Hello, thanks for your proposal. I removed the useless if condition now in a1051c3274cc83cc98e1623c5122857467afb0e9 for OTRS 5, and also for master. Thanks!

PR #1772, opened 2017-06-12
Fix for regex to be able to match long GeoTLD.

m00derp:
Currently in the proposed to be changed file we have the following regex for domain checking

( $FQDN !~ /^([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,6}$/ ) 
since we are using a "simple" regex for matching 2-6 chars for .com .biz etc.
the problem is not being a geotld in our case but the characters being too many.
I suggest increasing this value 2,6 or 2,10 since the biggest geotld I found was vlaanderen.
The regex would then be
elsif ( $FQDN !~ /^([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,12}$/ ) enabling a match even in the long GEOTLD domain names.

### DIFF
--- a/Kernel/System/SupportDataCollector/Plugin/OTRS/FQDN.pm
+++ b/Kernel/System/SupportDataCollector/Plugin/OTRS/FQDN.pm
@@ -38,7 +38,7 @@ sub Run {
     }
 
     # FQDN syntax check.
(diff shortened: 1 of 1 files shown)


mgruner:
Merged in 17faf0650b5e4fe979e9eca133eb9b3cf7a364fd for OTRS 5 and cherry-picked to master and OTRS 4. Thanks a lot @m00derp!

PR #1774, opened 2017-06-20
Text templates selection list is not presented when creating ticket

StCyr:
In OTRS-4.0.20, when creating an email or phone ticket the list of valid templates of type "Create" is always empty.

The related code is the same in branch "master"

The commit in this PR fixes the issue.

BEFORE COMMIT 9c0ea56:
=================

image

AFTER COMMIT 9c0ea56:
================

image

best regards,

Cyrille

### DIFF
--- a/Kernel/Modules/AgentTicketEmail.pm
+++ b/Kernel/Modules/AgentTicketEmail.pm
@@ -2575,10 +2575,10 @@ sub _MaskEmailNew {
     # build text template string
     if ( IsHashRefWithData( \%StandardTemplates ) ) {
         $Param{StandardTemplateStrg} = $LayoutObject->BuildSelection(
-            Data       => $Param{StandardTemplates}  || {},
-            Name       => 'StandardTemplateID',
-            SelectedID => $Param{StandardTemplateID} || '',
-            Class      => 'Modernize',
+            Data         => \%StandardTemplates,
+            Name         => 'StandardTemplateID',
+            SelectedID   => $Param{StandardTemplateID} || '',
+            Class        => 'Modernize',
             PossibleNone => 1,
             Sort         => 'AlphanumericValue',
             Translation  => 1,
--- a/Kernel/Modules/AgentTicketPhone.pm
+++ b/Kernel/Modules/AgentTicketPhone.pm
@@ -2522,10 +2522,10 @@ sub _MaskPhoneNew {
     # build text template string
     if ( IsHashRefWithData( \%StandardTemplates ) ) {
         $Param{StandardTemplateStrg} = $LayoutObject->BuildSelection(
-            Data       => $Param{StandardTemplates}  || {},
-            Name       => 'StandardTemplateID',
-            SelectedID => $Param{StandardTemplateID} || '',
-            Class      => 'Modernize',
+            Data         => \%StandardTemplates,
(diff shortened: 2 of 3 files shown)


mgruner:
@dvuckovic can you pass this to gatekeeping please?

s7design:
Hello @StCyr ,

This PR change intentional OTRS behaviour where Template selection list depends on Queue selection. Once Template is created it must be linked with Queue in AdminQueueTemplates screen in order to be selectable. This allows agents to have specific Templates for desired queue in which ticket is created. When Queue is selected, AJAX call update Template field and show appropriate available selection.

Please if you have any questions regarding this issue feel free to ask.

@dvuckovic My suggestion is to close this PR.

Regards,
Sanjin

StCyr:
Hi, 

I'm currently testing this functionality and have created templates and assigned them to queues.

Templates of type "Answer" can be selected without problem (for replying to an email), but templates of type "Create" don't work at all (without this patch).

I agree that my patch is quick and dirty but I mainly wanted to raise a concern related to this part of the code which doesn't seem to make any sense:

First, OTRS gets the list of "Create" standard templates regardless of the queue:
```perl
    # check if exists create templates regardless the queue
    my %StandardTemplates = $Self->{StandardTemplateObject}->StandardTemplateList(
        Valid => 1,
        Type  => 'Create',
    );
```

Then, OTRS creates the selection list based on a parameter given by the user ($Param{StandardTemplates}) rather than  this list of "Create" standard templates
```
    # build text template string
    if ( IsHashRefWithData( \%StandardTemplates ) ) {
        $Param{StandardTemplateStrg} = $Self->{LayoutObject}->BuildSelection(
            Data       => $Param{StandardTemplates}  || {},
            Name       => 'StandardTemplateID',
            SelectedID => $Param{StandardTemplateID} || '',
            PossibleNone => 1,
            Sort         => 'AlphanumericValue',
            Translation  => 1,
            Max          => 200,
        );
        $Self->{LayoutObject}->Block(
            Name => 'StandardTemplate',
            Data => {%Param},
        );
    }
```

I do believe there's something wrong there.

dvuckovic:
Hi @StCyr,

I do believe that @s7design is right, since I cannot reproduce reported behavior. I tested in both latest OTRS 4 and OTRS 5 systems.

There is a feature in place that is intentionally leaving template field empty, until queue is selected above. The reason you found that parameter will be used as list data instead is also a feature: if queue is pre-selected via URL (for example: `index.pl?Action=AgentTicketPhone;Dest=2||Raw`), template list will also get populated correctly.

So, here are the steps for getting correct template list:

* Go to ticket creation screen
* Verify there are no templates in the list
* Select a queue above, making sure it's the one with some "Create" templates assigned to it
* Verify that the template list is refreshed, and said templates shown

I will close your PR for now, please feel free to submit a bug report using our [public Bugzilla tracker](https://bugs.otrs.org/) if you still believe there is a defect in OTRS.

Thanks!

Regards,
Dusan

StCyr:
Hi,

You're right: I got a wrong "Email" template (that I had first created for my tests) assigned to all my queues rather than the "Create" template which I left unassigned to any queue.

It's now working perfectly as it should without the patch.

Thanks for your help.

Cyrille

PR #1775, opened 2017-06-20
Update AdminNotificationEvent.tt

nilei:
"Customer User ID" instead of "Customer"


mgruner:
Merged, thanks @nilei!

PR #1622, opened 2017-02-20
[master] Changed customer to customer user

urbalazs:
Hi @mgruner 
This is only for branch master. For rel-5_0 see: https://github.com/OTRS/otrs/pull/1621

PR #1776, opened 2017-06-20
Update AdminNotificationEvent.pm

nilei:
"Customer user" instead of "Customer"


urbalazs:
Similar PR can be found here: https://github.com/OTRS/otrs/pull/1622

mgruner:
Merged, thanks @nilei!

PR #1777, opened 2017-06-26
Xadt add china name master

WEIBINBINXADT:
This function is provided by "Xian Dian Tong Software Co., Ltd., 'http://www.dian-tong.com/'" on the Chinese name display branch


mgruner:
Seems to be a duplicate of #1778, closing.

PR #1778, opened 2017-06-27
modify the question of the Chinese name

WEIBINBINXADT:
This function is provided by "Xian Dian Tong Software Co., Ltd., 'http://www.dian-tong.com/'" on the Chinese name display branch


Kernel/Config/Files/Framework.xml
```diff
@@ -8202,8 +8202,8 @@ You can log in via the following URL:
         <Setting>
             <Hash>
                 <Item Key="Module">Kernel::Output::HTML::CustomerUser::Generic</Item>
-                <Item Key="Required">UserFirstname;UserLastname;</Item>
```
mgruner: This looks wrong for non-chinese users. Can you revert this please? Same also in the following lines.

Kernel/Config/Files/Framework.xml
```diff
@@ -10988,6 +10988,21 @@ via the Preferences button after logging in.
                 <Item Key="7" Translatable="1">Lastname Firstname (UserLogin)</Item>
                 <Item Key="8" Translatable="1">(UserLogin) Lastname Firstname</Item>
                 <!-- End Hungarian -->
+                <Item Key="9" Translatable="1">LastnameFirstname</Item>
+            </Option>
+        </Setting>
+    </ConfigItem>
+    <ConfigItem Name="CustomerFirstnameLastnameOrder" Required="1" Valid="1">
```
mgruner: Why should this be different from FirstnameLastnameOrder? I don't see that we need a separate setting for this ATM. If it was neccessary, it would need the same options like FirstnameLastnameOrder.

Kernel/Config/Files/Framework.xml
```diff
@@ -11454,8 +11469,8 @@ via the Preferences button after logging in.
         <Setting>
             <Array>
                 <Item>UserLogin</Item>
-                <Item>UserFirstname</Item>
```
mgruner: See my first comment.

Kernel/Modules/AdminRoleUser.pm
```diff
@@ -54,7 +54,7 @@ sub Run {
             Selected => \%Member,
             Data     => \%RoleData,
             ID       => $UserData{UserID},
-            Name     => "$UserData{UserFirstname} $UserData{UserLastname} ($UserData{UserLogin})",
+            Name     => "$UserData{UserFullname} ($UserData{UserLogin})",
```
mgruner: This will not work, because some FirstnameLastname options include the UserLogin too, which would then be doubled. Perhaps we should just remove the hard coded UserLogin here too?

Kernel/Modules/AdminUserGroup.pm
```diff
@@ -65,7 +65,7 @@ sub Run {
             %Types,
             Data => \%GroupData,
             ID   => $UserData{UserID},
-            Name => "$UserData{UserFirstname} $UserData{UserLastname} ($UserData{UserLogin})",
+            Name => "$UserData{UserFullname} ($UserData{UserLogin})",
```
mgruner: See my comment about AdminRoleUser above.

Kernel/System/CustomerUser.pm
```diff
@@ -1299,6 +1342,58 @@ sub CustomerUserCacheClear {
     return 1;
 }
 
+=head2 _CustomerUserFullname()
+
```
mgruner: As said above, this should be implemented like in User.pm, probably also with the same config option FirstnameLastnameOrder.

Kernel/System/Ticket.pm
```diff
@@ -6437,7 +6437,7 @@ sub TicketWatchSubscribe {
         TicketID     => $Param{TicketID},
         CreateUserID => $Param{UserID},
         HistoryType  => 'Subscribe',
-        Name         => "\%\%$User{UserFirstname} $User{UserLastname} ($User{UserLogin})",
```
mgruner: See above comment for AdminRoleUser.pm (about UserLogin), also for following lines.

mgruner:
Excellent! Merged now in 585793398bc200540beebd5fb48d0d32af176eba. Thanks a lot @WEIBINBINXADT.
I also extended FirstnameLastnameOder.t, please see https://github.com/OTRS/otrs/commit/585793398bc200540beebd5fb48d0d32af176eba#diff-00d336f63175952fe5d571d06c417f99.

PR #1781, opened 2017-07-01
Replace hardcoded bcrypt_hash({ cost => 9 with cost => $Cost

melak:
While here, I am also taking the opportunity to very strongly suggest taking a look at your password storage policies. Even offering the ability to store passwords unsalted is highly questionable; **_defaulting to storing passwords unsalted_** is reckless at best.


mgruner:
Hello @melak, thanks for your submission. Applied in 676bda63b3b2651eb4202a1d9a847638d06b22c4. Can you please explain in a bit more detail what you mean with your comment about unsalted storage?

PR #1783, opened 2017-07-04
Marked system types as translatable

urbalazs:
Hi @mgruner 
The system types are not marked for translation, and the dropdown menu is not modern. Maybe this proposal can solve the problem (I couldn't try it). Both branch rel-5_0 and master are affected.


mrcbnsls:
Merged, thanks!

PR #1784, opened 2017-07-04
Update AdminNotificationEvent.tt

nilei:
The subject of notifications was limited to 200 characters. When lots of dynamic fields or other notification tags are in use, a small subject with only a few characters can result into a very large subject with notification tags in the notification admin interface.

Example: <OTRS_TICKET_DynamicField_123_Value> is not <OTRS_TICKET_DynamicField_234_Value> and not <OTRS_TICKET_DynamicField_345_Value> but close to <OTRS_TICKET_DynamicField_456_Value> or <OTRS_TICKET_DynamicField_567_Value>

Result: 1 is not 2 and not 3 but close to 4 or 5

Cheers, Nils

### DIFF
--- a/Kernel/Output/HTML/Templates/Standard/AdminNotificationEvent.tt
+++ b/Kernel/Output/HTML/Templates/Standard/AdminNotificationEvent.tt
@@ -557,7 +557,7 @@
                             [% Translate("Subject") | html %]:
                                         </label>
                                         <div class="Field">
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @nilei,
I'm afraid the current limitation is valid, because also the `notification_event_message.subject` field on the database has 200 characters, and thus cannot store longer strings. If you feel that this is wrong, perhaps you could consider opening a bug report.
Best regards, Martin

PR #1787, opened 2017-07-06
Update Ticket.xml

nilei:
Open tickets do not necessary mean that someone needs to communicate with or answer anyone. They are just open.


mgruner:
Applied in 2ca92c3817e18c4f390febb7c3c6d092db478404. Thanks @nilei! 👍

PR #1789, opened 2017-07-14
[rel-5_0] Improved output message

urbalazs:
Hi @mgruner 
This output message is generated by `print` method, so the colourize is not working:
console

This output message is missing from branch master. Intentionally or accidentally?


mgruner:
Applied, thanks a lot @urbalazs. I believe in master we just reduced the verbosity a bit.

PR #1744, opened 2017-05-03
Improved wording, marked error messages for translation

urbalazs:
Hi @mgruner 
The error messages already have been marked for translation, but the `Translate()` function was missing from the `.tt` file. I also improved the wording by concatenate the words and params into a sentence. There are many languages, where the order of the words is different from English.
Both branch rel-5_0 and master are affected!

PR #1792, opened 2017-07-19
[rel-5_0] Marked error messages for translation

urbalazs:
Hi @mgruner 
This is the backport version of https://github.com/OTRS/otrs/pull/1744 for branach rel-5_0.
You wrote, you don't want to make changes to the strings in OTRS 5. This PR changes a little bit (remove a colon from the string), but the other changes doesn't affect the existing translation, because the translated strings are already in the language files.
If you don't agree with my minor string change, then feel free to remove it from this proposal.


mgruner:
Merged in 518098eb9e74f71daafcbc3d9848dee0d38a6520 without the string change. Thanks a lot @urbalazs!

PR #1794, opened 2017-07-21
Process ticket history modify

WEIBINBINXADT:
This version has modified previous codepolicy error.

### DIFF
--- a/Kernel/Config/Files/ProcessManagement.xml
+++ b/Kernel/Config/Files/ProcessManagement.xml
@@ -486,5 +486,47 @@
             </Hash>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::Frontend::MenuModule###550-ProcessHistory" Required="0" Valid="1">
+        <Description Translatable="1">Shows a link in the menu to add a process history in the ticket zoom view of the agent interface.</Description>
+        <Group>ProcessManagement</Group>
+        <SubGroup>Frontend::Agent::Ticket::MenuModule</SubGroup>
+        <Setting>
+            <Hash>
+                <Item Key="Module">Kernel::Output::HTML::TicketMenu::Generic</Item>
+                <Item Key="Name" Translatable="1">Process History</Item>
+                <Item Key="Description" Translatable="1">Process History</Item>
+                <Item Key="Action">AgentTicketProcessTrace</Item>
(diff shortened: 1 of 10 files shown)


mgruner:
Hello @WEIBINBINXADT, thank you very much for your contribution. Again, this is an important use case.
I see some issues with the current implementation though:

- In the customer panel, we currently don't have a ticket history at all. This feature should not be active by default, only on request, if at all.
- We currently don't reveal agent details by default in the customer panel.
- This could reveal data to the customer that is not wanted (i. e. from times when the ticket was not accessible to the customer because it was in internal queues; and from process details that should not be known to the customer.
- It introduces terms not common in OTRS (operator instead of agent).
- The integration in the customer interface is not yet finished (text link instead of icon).
- In the agent interface, it could be discussed if it would be helpful to integrate this functionality into the regular ticket history screen (maybe with tabs?).

@mrcbnsls, @jensoliver, @nilei can you please also offer your opinion on this suggestion?

mrcbnsls:
Hi @WEIBINBINXADT/@mgruner, could you please add some meaningful description to this PR, explaining what it is about and what it is meant to achieve?

WEIBINBINXADT:
I'm so sorry that I forgot to write description for this PR  and hope that it did not cause you too much inconvenience. This PR is about show a table for process ticket history. This PR consists of the following：
1. For process ticket, a new HistoryView is created, where the process progresses are shown in a table.
2. Each row of the table, the activity change, the agent / the customer, the incoming time and the leaving time of the activity and the duration are shown.
3. The activity are displayed by name instead of Activity-Entity-IDs in the normal history view.
4. This feature will be available both for customer and agent interface.

jensoliver:
Hi @WEIBINBINXADT,
can you please add some more information like screenshots?

I fully agree with the issues @mgruner mentioned already

mrcbnsls:
@WEIBINBINXADT, do you think you could add some screenshots as requested by @jensoliver? Thanks a lot.

@mgruner, I'm not sure how we should proceed on this. As we're feature freeze and this needs probably some more discussion, it's more likely something for OTRS 7, isn't it?

mgruner:
Referring to my previous comment, I'll close this PR for now. We would be glad to discuss this and receive an updated version for OTRS 7!

PR #1795, opened 2017-07-24
Xadt dashboard extension

WEIBINBINXADT:
The version's function is extended dashboard，from Xian Dian Tong Software Co., Ltd., http://www.dian-tong.com/.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -7227,19 +7227,19 @@
             </Hash>
         </Setting>
     </ConfigItem>
-    <ConfigItem Name="DashboardBackend###0120-TicketNew" Required="0" Valid="1">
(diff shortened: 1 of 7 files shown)


mgruner:
Hello @WEIBINBINXADT. Thank you very much for your contribution. It covers important use cases. In the current form, however, I don't want to merge this to OTRS:

- The current "New tickets" widget is just deleted.
- The two new widgets are copies of TicketGeneric.pm, with minimal changes, almost identical to each other. That leads to a lot of unneeded code redundancy.

Why don't we talk about extending TicketGeneric.pm with additional filter capabilities, so that one code module could be used for all widgets? If we combine that with the concept of combinable filters, then we can make the module flexible enough to cover all use cases like (show me all open tickets in my queues). There are already ideas for the required UI changes here, let us know if you are interested in helping with that. We could ask @mrcbnsls to come up with mockups.

Additionally, for future contributions of significant features, I'd ask you to include backend and/or Selenium test cases, please.

Best regards, Martin

PR #1797, opened 2017-07-25
Add Search Attribute by Single Click

feiming769864:
Currently if a search attribute need to be added, this attribute should be selected first and then the + button has to be clicked. This feature supports, if the search attribute is selected, it will be automatically added.


mrcbnsls:
Thanks, nice. Merged in 1269372f4a. I extended this a bit and removed the plus button which is now not needed anymore. This required some changes in the unit test.

I also recognized that adding the attribute on change was not possible for the preselected attribute "TicketNumber". Thats why I added a ```PossibleNone``` to the ```BuildSelection()``` call which builds the attribute selection field.

PR #1799, opened 2017-07-28
Update AgentAppointmentEdit.tt

nilei:
The target of the "a" element should not be a new window as this breaks the workflow for the user.

Example:
1. Zoom into ticket => same window
2. Follow linked appointment => same window
3. Edit appointment => same window (floater)
4. Follow linked ticket in edit form => NEW window
5. Follow linked appointment again => same window
6. ...


mrcbnsls:
Hi @nilei, this PR was wrong regarding the place in the code and also there is no Data.Target at either of the available places I think.

I though agree on your change from a user perspective and removed the target="_blank" from both places in the edit view. See https://github.com/OTRS/otrs/commit/56a15017a14409592157152f885e97316d26f53b.

@dvuckovic, please let me know if you disagree.

PR #1803, opened 2017-08-08
Add GenericAgent Search Attribute

XADT-huyabo:
Currently if a search attribute need to be added, this attribute should be selected first and then the + button has to be clicked. This feature supports, if the search attribute is selected, it will be automatically added.


mrcbnsls:
Thanks for your contribution, merged in a3dbab57d4.

For future contributions (which would be very welcome!), I'd ask you to sign our CLA as described on http://doc.otrs.com/. Thanks in advance!

PR #1805, opened 2017-08-09
New Ticket Quick Operation

XADT-huyabo:
We can use this menu to create a new ticket quickly.


mgruner:
Hello @XADT-huyabo, @fengbo31200, thank you very much for your suggestion. I see some issues here:

- same AccessKey is assigned twice
- same main menu entry shows up twice 

Instead, I recommend using the toolbar modules like `Frontend::ToolBarModule###140-Ticket::AgentTicketPhone` - this also exists for email and allows direct ticket creation even for the different channel types. With OTRS Business Solution™ 6, we will have Phone, Email and SMS.

Best regards, Martin

PR #1806, opened 2017-08-10
Config Possible State for Ticket Views

XADT-huyabo:
This feature supports state config for all ticket views, you can find corresponding system config by typing "Ticket::Frontend::$Action###State" in search input,.

### DIFF
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -5851,6 +5851,28 @@
             </Array>
         </Setting>
     </ConfigItem>
+
+    <ConfigItem Name="Ticket::Frontend::AgentTicketCompose###State" Required="0" Valid="0">
+        <Description Translatable="1">Defines the next possible states after composing / answering a ticket in the ticket compose screen of the agent interface.</Description>
+        <Group>Ticket</Group>
+        <SubGroup>Frontend::Agent::Ticket::ViewCompose</SubGroup>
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @XADT-huyabo and @fengbo31200. Thank you very much for your contribution. I agree with the idea of this feature.

However, there is a problem. The setting names collide with existing settings:

    <ConfigItem Name="Ticket::Frontend::AgentTicketFreeText###State" Required="0" Valid="1">
        <Description Translatable="1">Sets the state of a ticket in the ticket free text screen of the agent interface.</Description>

    <ConfigItem Name="Ticket::Frontend::AgentTicketFreeText###State" Required="0" Valid="0">
        <Description Translatable="1">Defines the next state of a ticket after adding a note, in the ticket free text screen of the agent interface.</Description>

Also, the descriptions are just copied from *###StateType and don't mention that this actually overrides the definition in ###StateType. Both descriptions probably need to be enhanced.

Last but not least, this change lacks backend tests. Please have a look at scripts/test/Ticket/TicketMoveList.t for an example. We could have a similar TicketStateList.t file that defines a list of tests for existing and the new functionality. The tests can define the needed config settings which are then applied before the call to TicketStateList(). Please contact me if anything is unclear.

Thanks again for your contribution, let's work on this to get it improved and merge-able. I'll close the PR in the current form.

Best regards, Martin

PR #1808, opened 2017-08-11
Unified error messages

urbalazs:
Hi @mgruner 
Here are the unified error messages without translation parts, as you requested in https://github.com/OTRS/otrs/pull/1713.


mgruner:
Merged, thank you very much, @urbalazs!

PR #1810, opened 2017-08-14
visibility improvement on AgentTicketZoom

florianedlhuber:
Stripping off after ~30 characters is not good. Extended to 100


mrcbnsls:
Thanks, merged!

PR #1812, opened 2017-08-15
Update Defaults.pm

florianedlhuber:
in CustomerUser stuff (LDAP) changed typos ('value'), instead of ('value)',


mgruner:
Merged, thanks a lot @florianedlhuber!

PR #1815, opened 2017-08-16
Rel 5 0 11 Prevent commas in E-Mail real name

MariusTimmer:
We found out that the real name of an system address can contain commas. The AutoResponses will cause problems at that point because the address will be split by commas. If the real name contains this character one email address will be separated into two broken parts.
I've received some E-Mails from addresses like "institute@domain.invalid". To prevent this, this pull request will not longer accept commas in the real name.


mgruner:
Hello @MariusTimmer,
thanks a lot for your analysis and even proposal for a fix. This is greatly appreciated. I'd like to better understand what really is the problem, i. e. what exactly breaks when special characters are applied in the real name field.
OTRS uses Mail::Address to parse and encode addresses in a safe way. IMHO this should not be prevented in this input field for just one problematic character (and not others like parentheses for example). Rather, we should make sure that addresses are correctly formatted or split.
So I would like to encourage you to dig into this a bit deeper. Please at least open a bug explaining the problem in detail. If you can then also present a PR for this, even better! I would also ask you then to please sign our CLA document as described on http://doc.otrs.com. Thanks!
Best regards, Martin

PR #1818, opened 2017-08-20
Fix CommunicationID matching in GenericInterfaceDebugger.js

skirsdeda:
Changed `parents('tr')` to `closest('tr')` because in some cases click event will be fired by `<tr>` itself and parents would find some `<tr>` very far in ancestors list, then `find('input.CommunicationID')` will get the first matching `<input>` in entire document which is obviously incorrect. `closest('tr')` works correctly when click event is fired by `<tr>` or any of it's children.

### DIFF
--- a/var/httpd/htdocs/js/Core.Agent.Admin.GenericInterfaceDebugger.js
+++ b/var/httpd/htdocs/js/Core.Agent.Admin.GenericInterfaceDebugger.js
@@ -124,7 +124,7 @@ Core.Agent.Admin.GenericInterfaceDebugger = (function (TargetNS) {
             $('#RequestList tbody').html(HTML);
 
             $('#RequestList a').on('click', function() {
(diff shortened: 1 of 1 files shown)


skirsdeda:
My reasoning in the previous comment is incorrect :) The actual reason why I get incorrect behavior is because there is more than one `<tr>` in ancestors of every `<a>` element that is clicked which doesn't happen with standard templates. Anyway, using `closest('tr')` instead of `parents('tr')` is more correct in this case so PR is still valid.

mrcbnsls:
Hi @skirsdeda, thanks for your contribution. I'm unfortunately not able to merge it without any description on how to reproduce the issue you're facing. Although your fix sounds good to me, I'd need to reproduce the issue and see if it really fixes it. Thanks a lot for updating the PR accordingly.

mrcbnsls:
Closing for now. Please reopen with the requested information. Thanks!

PR #1820, opened 2017-08-22
Enable configurable avatar if the senders address equals the current users one.

jensoliver:
So other settings than Mistery Man can be used. E.g. also an external picture. See http://de.gravatar.com/site/implement/images/ for more details


mrcbnsls:
Nice, thanks! Merged with some small wording changes, see ce88e2db4efd7470315527241669d3a145ad01f6

PR #1821, opened 2017-08-24

urbalazs:
Hi @mgruner 
I found some strings in the customer frontend, which are not marked for translation. I don't know, that my proposal would be the best way to translate multiline HTML strings. Please feel free to modify it, if there is a better solution available.
OTRS 5 is also affected, a back port would be needed.


mgruner:
Applied in b1ba8378d34727832e9a40b0d934a44479dfdc63, thanks @urbalazs!

mgruner:
(also back ported)

PR #1826, opened 2017-09-12

urbalazs:
Hi @mgruner 
This is a Hungarian translation update for OTRS 4. Many strings have been changed since my last translation update, and I also synced the translation from OTRS 5 to have the same wording.
As Transifex handles translation for OTRS 5, this is the only way to update the existing translation.


mgruner:
Merged, thanks a lot,  @urbalazs, as usual!

PR #1827, opened 2017-09-12
[rel-3_3] Updated Hungarian translation

urbalazs:
Hi @mgruner
This is a Hungarian translation update for OTRS 3.3. This increases the completeness from 29% to 100%.
I'm not sure if there will be a new release from this branch in the future, but it would be great if the 3.3.19 would be fully translated.


mgruner:
Merged in 1485cf4265f4cdb1a510037b6183b79ba724b7bf, thanks a lot @urbalazs.

PR #1828, opened 2017-09-14
Added Hungarian notification

urbalazs:
Hi @mgruner 
Here is the translation for the new notification (OTRS 6 and 7).


mgruner:
Merged, thanks a lot, @urbalazs.

PR #1831, opened 2017-09-16
Removed test case from Core.Debug js

kmhl:
Core.Debug2 is not defined by module at all


dvuckovic:
Hi @kmhl,

thanks for your pull request!

However, this seems intentional, in order to test different wordings of functions (one with namespace, one without). Please note that the third argument `RequiredLabel` is just a string used in an error message, it does not have to have a meaning. Method is tested if it returns false value, which is expected in this case.

Regards,
Dusan

PR #1832, opened 2017-09-17
Removed duplicated Core.Data.js note

kmhl:


dvuckovic:
Hi @kmhl,

thank you for the pull request, but this seems strange. `RequiredLabel` argument in `CheckDependency()` is not optional: it is used in the error message. While it seems that this might be duplicated code, it really not is because the string is referenced later. Please take a look at http://doc.otrs.com/doc/api/otrs/7.0/JavaScript/Core.Debug.html#.CheckDependency

Regards,
Dusan

PR #1833, opened 2017-09-18
Fixed translatable strings

urbalazs:
Hi @mgruner 
The `Translatable()` function doesn't work, if the parameter contains string concatenation. Because of that now only the first part is containing in the language files.
Branch rel-6_0 and master are affected.


dvuckovic:
Merged in a886ecd3c8 and 5715c4e82b, thanks @urbalazs!

PR #1834, opened 2017-09-18
Mark string as translatable

urbalazs:
Hi @mgruner 
I found 2 strings, that are not marked as translatable. These strings are displayed in the table header of AdminRoleUser screen, when clicking on an agent.
Unfortunately all supported branches are affected: rel-4_0, rel-5_0, rel-6_0 and master. Can you please apply this PR in all branches?


dvuckovic:
Merged in 6508e4b004, ff262653aa and c344fb9970, thanks @urbalazs!

Strings in `rel-4_0` are already frozen, so we will not be applying this patch there.

PR #1835, opened 2017-09-18
Changed FirstnameLastnameOrder subgroup

urbalazs:
Hi @mgruner 
This setting is not only applied to agents, but also to customer users, therefore the `Frontend::Base` would be better subgroup for it.


dvuckovic:
Merged in 24c9833049 and 8a12f7ddd7, thanks @urbalazs!

PR #1838, opened 2017-09-18
Improved AdminCustomerUserService screen

urbalazs:
Hi @mgruner 
This PR changes the following:
- replaced "Customer" with "Customer User"
- marked breadcrumb title as translatable
- replaced apostrophe with colon, as some languages (e. g. Hungarian) uses other character for quotation, but colon is the same in all language (maybe it can be ndash instead of colon)


dvuckovic:
Merged in e00bb68e40 and 97805a15d9, albeit with a small change: introduced colon was dropped from output display. Please consider changing this in all relation screens first (like `AdminUserRole`, `AdminUserGroup`, etc), since with just this change we would lose consistency among screens.

Thanks @urbalazs for your contribution(s)!

PR #1839, opened 2017-09-18

urbalazs:
Hi @mgruner 
This string is still untranslatable. It affects all supported branches. Can you please apply this to all branches?


dvuckovic:
Merged in b3da76803f, e26f40e2e3 and c1a64381c8, thanks @urbalazs!

PR #1840, opened 2017-09-18
Marked mandatory fields

urbalazs:
Hi @mgruner 
The mandatory fields for Services are not marked as mandatory. All supported branches from rel-3_3 to master are affected.


dvuckovic:
Merged in 40cdbc3395 and 92579f9e7f, thanks @urbalazs.

I decided to omit changes for `rel-5_0` and `rel-4_0`, since this is changing default output of already released versions. Someone might complain.

PR #1841, opened 2017-09-18
Added OTRS Blog to Framework.xml

Rooobaaat:


dvuckovic:
While this seems a nice addition for OTRS 7, we first need to check if it makes sense to add a new default widget to the Dashboard as-is. Any widget that is doing remote fetching on the main thread (like `Kernel::Output::HTML::Dashboard::RSS` is), will add to the loading time of the dashboard, which could have adverse effects since this is a first screen after user login (by default). IMO, we should first think in modularizing the Dashboard in the same way as Ticket Zoom screen, so all widgets are loaded asynchronously.

Additionally, we should avoid renaming default options, and only add new ones, since then we can omit the upgrade code for overridden settings.

mgruner:
Dear contributor,

PR #1844, opened 2017-10-01
Update Framework.xml

nilei:
Put OTRS Group Services on top in Admin dashboard.


mgruner:
Hello, Nils,
I tend to prefer it a bit more decent like it is now. There is special highlighting for the OTRS Business Solution™ which draws attention. Thanks for your proposal!
Best regards, Martin

PR #1847, opened 2017-10-05
Configurable access to archive searching in agent panel

pboguslawski:
This mod introduces new SysConfig parameter Ticket::AgentArchiveSystem
that allows admin to disable access for archive searches in agent panel
(does not affect generic agent nor statistics modules). This may help
in protection against overloads in bigger OTRS setups (i.e. fulltext
searches in archive by agents).

Related: https://dev.ib.pl/ib/otrs/issues/113
Author-Change-Id: IB#1073066

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2017-??-??
+ - 2017-10-05 Configurable access to archive searching in agent panel.
  - 2017-03-26 Fixed bug#[12650](https://bugs.otrs.org/show_bug.cgi?id=12650)(PR#1636) - SendCustomerNotification does not respect newly assigned mail address. Thanks to S7!
  - 2017-03-24 Fixed bug#[12720](https://bugs.otrs.org/show_bug.cgi?id=12720)(PR#1672) - Settings window of Complex LinkObject is not translated. Thanks to S7!
  - 2017-03-24 Modernized address book. It is now possible to search for all configured custom user and customer fields.
--- a/Kernel/Config/Files/Ticket.xml
+++ b/Kernel/Config/Files/Ticket.xml
@@ -217,6 +217,17 @@
             </Option>
         </Setting>
     </ConfigItem>
+    <ConfigItem Name="Ticket::AgentArchiveSystem" Required="1" Valid="1">
+        <Description Translatable="1">Activates the ticket archive system search in the agent interface (does not affect generic agent nor statistics modules).</Description>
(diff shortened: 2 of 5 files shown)


mgruner:
Hello, @pboguslawski. Thanks a lot for your suggestion. I'm not sure about this use case and the consistency, though. The naming suggests it is valid for all the agent interface, but it only applies to the ticket search. Statistics can also cause full-text searches. Besides, if this is active, there is no way of accessing archived tickets anymore, even for admins. How do you see that?

@mrcbnsls @jensoliver @nilei would you give your opinion on this, please?

pboguslawski:
Hi,

> The naming suggests it is valid for all the agent interface

It was chosen analogously to Ticket::CustomerArchiveSystem. Feel free to apply better name if needed.

> Statistics can also cause full-text searches

Yes. Generic agent also. But in bigger setups regular agents do not have RW access to stats nor generic agent and both are easier to control (educate key users/developers not to kill DB with fulltext searches if there are not enough resources to do such things in particular setup).

> Besides, if this is active, there is no way of accessing archived tickets anymore, even for admins. How do you see that?

This param for stopping regular agents from searching archive but its still possible to access archived tickets using its number or link present in external sources (i.e. e-mail messages).

The proposed extension is radical but may help in case of bad performance problems, until better solution is implemented in particular setup (external fulltext engine integration like Sphinx or more MirrorDB engines will be available) which may need more time.

Regards,
Pawel

pboguslawski:
Hi,

Just commited enhancement (replaces previous solution): Ticket::ArchiveSearchUserGroup which allowes archive searching only if user does has rw access to group defined in Ticket::ArchiveSearchUserGroup parameter (no restriction if this parameter is not defined).

Should be better than Ticket::AgentArchiveSystem.

Regards,
Pawel

nilei:
Hi @pboguslawski , I would be okay with this if you allow the system configuration for the groups to be more than one group (array). This would not require the admin to create a new group just for this matter. Instead one can add the existing groups into the list. Btw. the default value for the group list should be "users", so if the restriction becomes enabled, still all "users" can access the archive search option.

Cheers, Nils

pboguslawski:
Hi,

This feature is for large OTRS setups with a lot of data. Such systems should rather use roles not groups to set permissions. One group suits our needs (we use it with roles) but feel free to adapt this extension to your needs.

Regards,
Pawel

mgruner:
Dear contributor,

PR #1848, opened 2017-10-05
Update Article.tt

nilei:
Removed confusing field separator. This was only used for the article field and confused several users.

Instead of this static field separator it would be nice to have some kind of pseudo elements available for process dialogs to add field separator, free text information, etc.

Cheers, Nils


mrcbnsls:
Thanks, merged!

PR #1850, opened 2017-10-12
fix little typo

reneeb:


mgruner:
@dvuckovic what do you think? It might be wise to rewrite that sentence.

Plus, we should disallow multiple "!" in a code policy filter for master, right?

mgruner:
Hello @reneeb, thanks for your suggestion. I applied a fix now in 57e3b64cecd01999409eb31379207bd2080cb39e, also removing the second part of the sentence with the multiple exclamation marks.

PR #1852, opened 2017-10-12
integrate toolbar ticketnumber search

reneeb:
This adds a new widget to do a ticket number search. This helps all agents that get a ticket number and want to open it very quickly.


mgruner:
Hello @reneeb. AFAICT OTRS already can do that with the existing full-text search toolbar widget. It checks if the string looks like a ticket number, and then redirects to that ticket. Therefore I believe that this change here is not needed, is it?

mgruner:
I'll close this PR for now. Please feel free to comment with further information!

PR #1854, opened 2017-10-16
Submit with Ctrl+Enter in textareas

Perleone:
Will not work if RichText is enabled!

### DIFF
--- a/var/httpd/htdocs/js/Core.Form.js
+++ b/var/httpd/htdocs/js/Core.Form.js
@@ -259,7 +259,7 @@ Core.Form = (function (TargetNS) {
 
     /**
      * This makes all forms submittable by using Ctrl+Enter inside textareas.
-     * On macOS you can use Command+Enter.
+     * On macOS you can use Command+Enter instead.
      * Does NOT work if Frontend::RichText is enabled!
      */
     $('textarea').keydown(function (Event) {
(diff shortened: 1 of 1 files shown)


mrcbnsls:
Hi @Perleone, thanks a lot for your contribution. I have some remarks:

- This should be improved to be compatible with OS X, too. Event.ctrlKey will not work there.
- I'm not sure if #submitRichtext is present everywhere. I think we should make sure that the submit button of the current form is used rather than one with a hardcoded ID.

Could you please update the PR with the requested changes? Thanks a lot!

Perleone:
> Could you please update the PR with the requested changes?

@mrcbnsls Done and done, checks passed successfully.

Perleone:
_"Once more unto the breach, dear friends, once more!"_

Perleone:
@mgruner The Travis CI bujild against Perl 5.20 and PostgreSQL fails on `/home/travis/build/OTRS/otrs/scripts/test/PostMaster/CommunicationLog.t:` with 
> Message: ERROR:  could not extend file "base/16387/16792": wrote only 4096 of 8192 bytes at block 381
> HINT:  Check free disk space.,
You might want to look at the Travis configuration. Or trigger a rebuild. Or both. :)

mrcbnsls:
Thanks a lot! Merged with a small change: I changed the binding to be a bit more greedy, so dynamically loaded textareas will be considered (e.g. in process tickets), too.

PR #1857, opened 2017-10-20
Marked string as translatable

urbalazs:
Hi @mgruner 
This string is not marked for translation.


mgruner:
Merged, thanks a lot @urbalazs!

PR #1858, opened 2017-10-22

urbalazs:
Hi @mgruner 
I found some strings, that are not marked for translation.


mgruner:
Merged, thanks @urbalazs!

PR #1859, opened 2017-10-23
Improved wording for console commands

urbalazs:
Hi @mgruner 
I reviewed all the console commands, and I found some wording issues. Actually branch rel-6_0 and master are affected, but if you think, you can also backport into rel-5_0.


mgruner:
Applied and backported to OTRS6. Excellent @urbalazs!

PR #1860, opened 2017-10-24
Improved wording for close buttons, marked strings as translatable

urbalazs:
Hi @mgruner 
I found some "Close" button that are ambiguous, because "Close" could mean "Close a dialog" or "Close a ticket". In Hungarian language we have two different words for them.
And I found some strings, that are not marked for translation.


mgruner:
Merged and backported to OTRS 6, thanks a lot @urbalazs!

PR #1861, opened 2017-10-24
Fixed SaveButtons widgets

urbalazs:
Hi @mgruner 
The SaveButtons widget of ACL and Process Management is collapsible, but they shouldn't be collapsible. This PR fixes the layout.
All supported branches (rel-5_0, rel-6_0 and master) are affected.


mrcbnsls:
Thanks, merged!

PR #1865, opened 2017-10-29
Marked field explanation as translatable

urbalazs:
Hi @mgruner 
This field explanation is not marked as translatable.


mgruner:
Applied and backported, thanks so much @urbalazs!

PR #1867, opened 2017-11-06

urbalazs:
Hi @mgruner 
This string is not marked as translatable. The first occurrence is in the Perl doc, so I don't know, if Translatable() is necessary here. If not, please remove it.


mgruner:

PR #1871, opened 2017-11-10
Do not update passwort to prevent rehashing

trciphron:
# Problem:
If `PasswordMaxLoginFailed` is set and an agent logs in more times than allowed, the status is changed to `invalid-temporarily`. The whole `%User` hash is passed to `UserUpdate`  - including the already hashed password.
This changes the password by hashing it again. 
# Solution:
This problem can be prevented by not passing the hashed password.

### DIFF
--- a/Kernel/System/Auth.pm
+++ b/Kernel/System/Auth.pm
@@ -288,6 +288,7 @@ sub Auth {
             %User,
             ValidID      => $ValidID,
             ChangeUserID => 1,
+            UserPw       => '',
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @trciphron, thank you very much. I created a bug entry at https://bugs.otrs.org/show_bug.cgi?id=13407 and applied a similar solution with new test cases, currently in QA. Please track that bug to be informed about the progress.

PR #1872, opened 2017-11-13

urbalazs:
Hi @mgruner 
I found some strings, that are not marked for translation.


mgruner:
Hello @urbalazs, thanks a lot. I merged and backported this, except the last one, as keys are not translated in selects. Please let me know if you see any problems with this.

PR #1873, opened 2017-11-13
Fixed incorrect postmaster filter matching

pboguslawski:
Postmaster filter with conditions like

    Body: non-existing-text
    From: EMAILADDRESS:non-existing-mail@test.com (negation checked here)
    To: EMAILADDRESS:matching-mail@test.com

matches message like

    From: <test@test.com>
    To: <matching-mail@test.com>
    Subject: test
    Message-ID: <test12345678@test.com>
    Content-Type: text/plain; charset="UTF-8"
    Mime-Version: 1.0
    Content-Transfer-Encoding: 8bit

    test

but should not. This mod fixes issue described above.

Related: https://dev.ib.pl/ib/otrs/issues/115
Author-Change-Id: IB#1074203

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 #6.0.0.beta1 2017-??-??
+ - 2017-11-13 Fixed incorrect postmaster filter matching.
  - 2017-03-26 Fixed bug#[12650](https://bugs.otrs.org/show_bug.cgi?id=12650)(PR#1636) - SendCustomerNotification does not respect newly assigned mail address. Thanks to S7!
  - 2017-03-24 Fixed bug#[12720](https://bugs.otrs.org/show_bug.cgi?id=12720)(PR#1672) - Settings window of Complex LinkObject is not translated. Thanks to S7!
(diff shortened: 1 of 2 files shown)


mgruner:
@s7design can you please handle this? We probably need a bug entry here.

milanrakic:
Hi @pboguslawski,

in my system, matching and mismatching work well. I used settings and message from your explanation.
Maybe I don't realised properly, so can you please give me more details for this case.

Regards,
Milan

pboguslawski:
Hi,

Make sure you've checked "Negate" in From condition only.

If still cannot reproduce - please anayse algo - its wrong cause in case of EMAILLADDRESS negation it XOR-s result of all earlier conditions not just this one condition result.

Hope this helps.

Regards,
Pawel

dvuckovic:
Merged in a496b14, 3f9e775 and 8e6bb2d. Thanks @pboguslawski!

Related bug report: https://bugs.otrs.org/show_bug.cgi?id=13361

PR #1876, opened 2017-11-17
added --follow-Outside-symlinks to otrs.SetPermissions.pl

xhoy:
Recently we moved our FS article data to a different partition and symlinked the var/article dir the right location.

It turned out that the SetPermission script skipped  directories that are outside the otrs/ basedir.

- We added a warning notice that a directory is skipped so this doesn't happen unseen
- added a script switch to skip/change the behaviour and follow symlinks outside the basedir.

With kind regards,

Thomas

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 #7.0.0.alpha1 2018-??-??
 
 #6.0.2 2017-??-??
+ - 2017-11-17 Added --follow-outside-symlinks and error for skipping directory to otrs.SetPermissions 
  - 2017-11-17 Fixed bug#[13343](https://bugs.otrs.org/show_bug.cgi?id=13343) - LDAP customer user backends should be implicitly read only.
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @xhoy. Thank you very much for your suggestion!

There seems to be a small Perl issue:
```
Perltidy version is 20170521

50: my ( $Help, $DryRun, $SkipArticleDir, $FollowOutsideSym @SkipReg ...
                                          ----------------- ^           
found Array where operator expected (previous token underlined)
1 file did not pass tidyall check
```

But more generally, I'm not sure about this feature. ATM, `otrs.SetPermissions.pl` is designed to only operate within the OTRS tree. You can also configure your article data and VirtualFS data to be stored elsewhere (without any symlinks), but that will also not be touched by this script. It is normally not necessary to update permissions on these data.

I'll close this PR for now. Please feel free to follow up with more information!

PR #1878, opened 2017-11-21
Improved button text

urbalazs:
Hi @mgruner 
With this button you can add _Customer User_ instead of _Customer_.


mgruner:
Merged with backport to 6 and 5. Thanks @urbalazs!

PR #1879, opened 2017-11-21

urbalazs:
Hi @mgruner 
I found some strings in Ticket.xml that are not marked for translation.


mgruner:
Merged without "Some description". Why should this need to be translated? It is meant to be set by the user, and will not be translated in the admin GUI. Thanks @urbalazs!

PR #1883, opened 2017-12-07
Add missing dependency (Sys::Syslog) in otrs.CheckModules.pl

ymyasoedov:
OTRS Log::SysLog log backend uses Sys::Syslog module. Without it I've got the following error message (from Apache logs):

```
[Thu Dec 07 13:41:32.599880 2017] [:error] [pid 27689] [Thu Dec  7 13:41:32 2017] -e: Can't load log backend module Kernel::System::Log::SysLog! Can't locate Sys/Syslog.pm in @INC (@INC contains: /usr/Custom /us
r/Kernel/cpan-lib /usr /usr/sbin/../../Custom /usr/sbin/../../Kernel/cpan-lib /usr/sbin/../.. /opt/otrs/Custom /opt/otrs/Kernel/cpan-lib /opt/otrs/ /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/
vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 . /etc/httpd) at /opt/otrs//Kernel/System/Log/SysLog.pm line 14.\n[Thu Dec  7 13:41:32 2017] -e: BEGIN failed--compilation aborted at /o
pt/otrs//Kernel/System/Log/SysLog.pm line 14.\n[Thu Dec  7 13:41:32 2017] -e: Compilation failed in require at (eval 67) line 2.\n
```

Where /opt/otrs//Kernel/System/Log/SysLog.pm line 14:

`use Sys::Syslog qw();`

### DIFF
--- a/bin/otrs.CheckModules.pl
+++ b/bin/otrs.CheckModules.pl
@@ -444,6 +444,16 @@ my @NeededModules = (
         },
     },
     {
+        Module    => 'Log::Syslog',
+        Required  => 0,
+        Comment   => 'Required for logging with Log::SysLog OTRS backend.',
+        InstTypes => {
+            aptget => 'libsys-syslog-perl',
+            emerge => 'perl-core/Sys-Syslog',
+            ports  => 'sysutils/p5-Sys-Syslog',
+        },
+    },
+    {
         Module    => 'Template',
         Required  => 1,
         Comment   => 'Template::Toolkit, the rendering engine of OTRS.',
--- a/bin/otrs.CheckModules.pl
+++ b/bin/otrs.CheckModules.pl
@@ -444,7 +444,7 @@ my @NeededModules = (
         },
     },
     {
-        Module    => 'Log::Syslog',
+        Module    => 'Sys::Syslog',
         Required  => 0,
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @ymyasoedov. This module is part of the Perl core distribution which is supposed to be installed in order to use OTRS. So I'm not sure if we need to list this module explicitly? How did you install Perl?

ymyasoedov:
I'm using Orcale 7.4 (RHEL 7.4) and Perl 5.16, Perl was installed with yum from standard RPM repos. Yes, you're right, Sys::Syslog is a part of core Perl distribution, but this module is installing as a separated package (in RHEL it is called perl-Sys-Syslog and it is not required by perl package). Furthermore, perl-Sys-Syslog was not installed by default with perl.

urbalazs:
Confirmed. I'm using CentOS 7.3, and installing Perl doesn't install perl-Sys-Syslog as dependency. Debian and Ubuntu contain it as default (or it is installed as dependency of some package).

mgruner:
Installing the `perl-core` package should fix this.

ymyasoedov:
You're right. Thanks!

PR #1884, opened 2017-12-14
Fixed typo

urbalazs:
Hi @mgruner 
I found this typo during translation. You can cherry-pick this also to rel-5_0, because the correct sentence is already in language files.


mbuc82:
Applied for rel-6_0 and master. Thank you @urbalazs!

PR #1887, opened 2017-12-14
Removed outdated translator credits

urbalazs:
Hi @mgruner 
Since OTRS 4 translator credits are stored in `i18n/otrs/*.po` files and updated automatically from Transifex. Translator credits in `Kernel/Language/*.pm` files are outdated and no longer updated.


urbalazs:
Hi @mgruner 
Can you check this PR?

mgruner:
@urbalazs as discussed, let's achieve that in a different way.

PR #1891, opened 2018-01-06

urbalazs:
Hi @mgruner 
These string are displayed in the Advanced personal settings of OTRS Business Solution. All others are translated, only these are untranslated.


mbuc82:
Applied for rel-6_0 and master. Thank you @urbalazs!

PR #1893, opened 2018-01-12
Add total count of links in block to block header

frennkie:
This PR adds the count of links for a block (e.g. number of linked tickets) to the block header.

Before:
otrs_link_count_before

After:
otrs_link_count_after

The reason for this change is that we sometimes have tickets that really have many linked tickets (e.g. a "master ticket") and it would be great to see at one glance how many tickets are linked in total.

### DIFF
--- a/Kernel/Output/HTML/Layout/LinkObject.pm
+++ b/Kernel/Output/HTML/Layout/LinkObject.pm
@@ -413,7 +413,7 @@ sub LinkObjectTableCreateComplex {
             Name => 'TableComplexBlock',
             Data => {
                 BlockDescription => $BlockDescription,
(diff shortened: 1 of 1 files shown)


Rooobaaat:
Cool! :+1:

frennkie:
@mgruner Why should I create a new pull request? I rebased and changed the base of the PR.. looks fine to me, no?

mgruner:
@frennkie yes, this is correct.

mgruner:
Merged in 14751a2428c69618411d257e7683f0514d8b6c94 even for OTRS 6, as it is really a very small change that should not have any side effects. Thanks!

PR #1894, opened 2018-01-15
Update Ticket.xml

Rooobaaat:
Default value for this setting is wrong. Should be 0 instead of 1


dvuckovic:
Merged in 50a56de and f5c5900, thanks @Rooobaaat!

PR #1897, opened 2018-01-22
Marked AutoResponse names as translatable

urbalazs:
Hi @mgruner
The AutoResponse names are already translated in AutoResponse <-> Queues screen, but they are still in English in the AutoResponse screen. All supported version of OTRS are affected.


s7design:
Hi,

Closing, there is bug report with fix:
https://bugs.otrs.org/show_bug.cgi?id=13672

Regards
Zoran

PR #1904, opened 2018-02-12
Marked headers and tooltips as translatable

urbalazs:
Hi @mgruner and @s7design 
This PR translates the column headers and tooltips of AdminCustomerGroup screen. The strings are already translated, but the translation was not applied yet.


s7design:
Hello @urbalazs ,

Thank you for PR. Fix is merged and bug#13676 (https://bugs.otrs.org/show_bug.cgi?id=13676) is closed.

Regards,
Sanjin

PR #1906, opened 2018-02-20
added option to send lostpassword-mail 

andreashergert1984:
If a customeruser backend has no valid-flag currently no lost-password mail will be send. 
Now if the user in a backend without a valid-flag it will be seen as valid as in the rest of the system.


Rooobaaat:
👍

PR #1908, opened 2018-02-22
Fail.pm is a (Customer)Auth module that always fails

felixjogris:
Hi,

I wrote Fail.pm in order to get a consistent error log message when an agent or a customer user does not provide correct credentials. It is meant to be the last authentication module in your configuration (fall-through). You can use it with Fail2ban (examples provided) or similar tools.
To enable it, add this to Kernel/Config.pm:

```
$Self->{"AuthModule10"} = 'Kernel::System::Auth::Fail';
$Self->{"Customer::AuthModule10"} = 'Kernel::System::Auth::Fail';
```

Syslog messages look like this:

> Feb 22 15:22:07 otrs OTRS-CGI-10[83711]: [Error][Kernel::System::Auth::Fail::Auth][Line:64]: AUTHFAIL10:192.168.23.42 USERNAME:someuser

If the webserver does not provide the REMOTE_ADDR environment variable, then "Got no REMOTE_ADDR env!" will be logged instead of an ip address.

BR,
Felix


mgruner:
Thanks a lot. This will be integrated into OTRS 7 soon, in a slightly different way.

PR #1915, opened 2018-03-14
Added Sysconfig Option to define the default gravatar image for artic…

andreashergert1984:
…lelist if no one is present

configuration:
Frontend::Gravatar::ArticleDefaultImage

### DIFF
--- a/Kernel/Config/Files/XML/Framework.xml
+++ b/Kernel/Config/Files/XML/Framework.xml
@@ -772,6 +772,13 @@
            <Item ValueType="String" ValueRegex="">mm</Item>
         </Value>
     </Setting>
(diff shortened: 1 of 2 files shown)


florianedlhuber:
Great 👍🏻

Rooobaaat:
Like that! 👍

nilei:
Must have !!! 

I am on site right now where we had to disable it because the identicon turned into a "Hakenkreuz". This is not OTRS fault of course, and we already can configure an alternative for agent users, but this PR will also make that possible for customer user senders. The default value btw. should be changed to "mm" instead of "identicon"!

Cheers, Nils

mrcbnsls:
Merged in 5da1d52a277e75eb9d77e5b2269a10ae7bf1b69e, thanks!

PR #1923, opened 2018-04-13
fixed packagedescription html insertion issue in packagemanager

andreashergert1984:
before: 
grafik

after: 
grafik


mgruner:
Hi @andreashergert1984, thanks for the proposal, but this needs to be fixed on a different level, in the cloud service.

PR #1925, opened 2018-04-25
Fix bad regexp for PGP detection, causes backtracking

Perleone:
The regexp contained several greedy matches, which cause backtracking.
The time needed to match grows exponentially with the length of the
string. An 8K string takes about 2 seconds to process, 10K takes over
7 seconds, and 20K more than 15 minutes.

### DIFF
--- a/Kernel/Modules/AgentTicketArticleContent.pm
+++ b/Kernel/Modules/AgentTicketArticleContent.pm
@@ -146,9 +146,9 @@ sub Run {
     # see bug#9672
     if (
         $Data{Content} =~ m{
-        ^ .* -----BEGIN [ ] PGP [ ] MESSAGE-----  .* $      # grep PGP begin tag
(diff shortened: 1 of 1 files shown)


s7design:
Hi,

Thanks for PR. I made feature branch we will see results after testing on out CI.

@sb-otrs please check it. What do you think about this changes in regex for PGP.

This is branch:
feature-pr-pgp-regex-zi-master

Reagrds
Zoran

s7design:
Hi,

It is merged on master branch

https://github.com/OTRS/otrs/commit/afe76838920adf0b11ab53c0491e3b6d6867a642

Please take care about that that this branch will not be used for the future releases and might go away. You should make a copy if you want.

You can see about that in the post on OTRS community web site that explains the new strategy.

https://community.otrs.com/

Regrads
Zoran

PR #1927, opened 2018-05-11
Content-Disposition header / unencoded filename of attachments

felixjogris:
Hi,

when the filename of an attachment contains non-ASCII or non printable characters, Apache 2.4 complains _Response header 'Content-Disposition' value of 'attachment;  filename="..." contains invalid characters, aborting request_.
Suggested patch is to convert the filename to ASCII first, and then to URI encode that as well.

BR Felix

### DIFF
--- a/Kernel/Output/HTML/Layout.pm
+++ b/Kernel/Output/HTML/Layout.pm
@@ -2481,9 +2481,8 @@ sub Attachment {
     if ( $Param{Filename} ) {
 
         # IE 10+ supports this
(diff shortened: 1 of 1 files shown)


s7design:
Hi,

Thanks for proposal, in meantime it is fixed with the following commit:

https://github.com/OTRS/otrs/commit/094ecfb84c08bb65cf95aeede680fd4f5e85e596

There is fixed more issues, and now we have different behavior. 

So I will close this PR.

Regards
Zoran

PR #1939, opened 2018-07-10
Configuration item fix

igor-lobanov:
&lt;Item Key="Module"&gt; usually uses package name. .pm looks needless

### DIFF
--- a/Kernel/Config/Files/XML/Ticket.xml
+++ b/Kernel/Config/Files/XML/Ticket.xml
@@ -13374,7 +13374,7 @@ Thanks for your help!
         <Navigation>Frontend::Agent::LinkObject</Navigation>
         <Value>
             <Hash>
-                <Item Key="Module">Kernel::Output::HTML::LinkObject::Ticket.pm</Item>
+                <Item Key="Module">Kernel::Output::HTML::LinkObject::Ticket</Item>
                 <Item Key="DefaultColumns">
                     <Hash>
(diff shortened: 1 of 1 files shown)


s7design:
Hi Igor,

Thanks for the PR. Actually it is not a bug, it is by design now. You are right that it is not consistent with some other sysconfig items which they are Module item. However, such type of config item is implemented on that way. If we change it in the sysconfig it will be necessary to change in the logic in code. They are such config items in some other OTRS packages. You can see that it is the same for Calendar config item which is part of the framework. 

/opt/otrs/Kernel/Config/Files/XML/Calendar.xml:
  459          <Value>
  460              <Hash>
  461:                 <Item Key="Module">Kernel::Output::HTML::LinkObject::Appointment.pm</Item>
  462                  <Item Key="DefaultColumns">
  463                      <Hash>

/opt/otrs/Kernel/Config/Files/XML/Ticket.xml:
 13376          <Value>
 13377              <Hash>
 13378:                 <Item Key="Module">Kernel::Output::HTML::LinkObject::Ticket.pm</Item>
 13379                  <Item Key="DefaultColumns">
 13380                      <Hash>


So, I will close it now. We can think about refactoring and it could be refactored in future. 

P.S. If you would like to make new PR, what is welcomed always, please subscribe  OTRS contributor agreement on the site http://doc.otrs.com/ 

Regards
Zoran

PR #1940, opened 2018-07-10
OTRS Business Solution™ is not used any more

florianedlhuber:


s7design:
Hi Florian,

   Thanks a lot for the PR. It will be fixed in next releases. We are working on it and it will be changed in some other modules and screens (no only Error.tt),  so I will close this PR it will be done in future. 

Regards
Zoran

PR #1941, opened 2018-07-16
Fix Bug: #13971

TheOneAndOnlyDoctor:
https://bugs.otrs.org/show_bug.cgi?id=13971


s7design:
Hi,

Thanks for proposal. It is fixed. You can follow communication on the bug report. 

Regards
Zoran

PR #1944, opened 2018-07-25
Test for checking bug in StandardTemplateLookup()

igor-lobanov:
Second call of StandardTemplateLookup() with unexisting parameter after successful call with existing parameter returns the value of previous call although we expect undef.


igor-lobanov:
Ok. Really bad idea to check bugs with test :)

PR #1948, opened 2018-08-20
Added ACL Action

florianedlhuber:
Added ACL Action AgentAppointmentCalendarOverview because it is missing.

### DIFF
--- a/Kernel/Config/Files/XML/Ticket.xml
+++ b/Kernel/Config/Files/XML/Ticket.xml
@@ -11047,6 +11047,7 @@
         <Navigation>Core::Ticket::ACL</Navigation>
         <Value>
             <Array>
+                <Item>AgentAppointmentCalendarOverview</Item>
                 <Item>AgentTicketBounce</Item>
                 <Item>AgentTicketClose</Item>
                 <Item>AgentTicketCompose</Item>


mgruner:
@dvuckovic could you please check this?

dvuckovic:
@florianedlhuber Thanks for the proposal.

However, I'm failing to understand the use-case behind this change? What's the expected usage of this new third-level param for ticket ACLs?

When I apply your change to a local system, I do get the action name in the autocomplete list and can choose it, but this screen does not really support ticket ACLs. The same goes for other screens unrelated to tickets in the system, which are also not listed here (e.g. AgentDashboard, AgentCustomerInformationCenter, etc).

I might be missing something obvious, though, so please outline an example for us.

Thanks!

Regards,
Dusan

florianedlhuber:
It's not for the view, It is to remove the new appointment button in ticket menu. Useful if you need to restrict it for a specific role. Any ticket menu action should be available in the ACLs.

mgruner:
@s7design I think you are familiar with this. Can you please check if it is still relevant?

s7design:
Applied thanks to  florianedlhuber.

Regards
Zoran

PR #1949, opened 2018-08-20
sync user ldap attributes to user preferences

felixjogris:
Hi,

Kernel::System::User::UserAdd() and ::UserUpdate() just set a fraction of a user's preferences. Besides first and last name it's just the email address and mobile phone number. So, when Auth::Sync::LDAP creates a new user or updates an existing user from LDAP, attributes like telephone number, fax number, etc. are not being synced to the OTRS database. That behaviour renders tag replacement (signatures, etc.) useless when agent's are defined in an Active Directory or such.

Attached patch introduces a new config variable AuthSyncModule::LDAP::SetPreferences to maintain compatibility. It can be set per AuthSync definition in Config.pm.

The patch also removes an else including indentation where the related if-block returns to the caller.

It also checks whether a user preference is defined before comparing to a maybe new LDAP attribute. This is to avoid an error like "Use of uninitialized value $UserData{$Attribute}".

BR,
Felix

### DIFF
--- a/Kernel/System/Auth/Sync/LDAP.pm
+++ b/Kernel/System/Auth/Sync/LDAP.pm
@@ -215,6 +215,9 @@ sub Sync {
     my $UserSyncMap = $ConfigObject->Get( 'AuthSyncModule::LDAP::UserSyncMap' . $Self->{Count} );
     if ($UserSyncMap) {
 
+        # if defined, sync all user ldap attributes to user preferences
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @felixjogris. There has been an improvement recently in https://github.com/OTRS/otrs/commit/419fc5f2269f2640c2d65952f70000a75bee1e55 which also solves this problem, but with different approach. Best regards, Martin

PR #1888, opened 2017-12-18
Implement StartTLS support for LDAP auth modules

LordAro:
I think I've covered everything that wants to be updated, do let me know if I missed something

I have to say, it's not exactly great that the codebase contains 3 separate LDAP connection objects. I'd definitely recommend merging in some fashon...

PR #1950, opened 2018-08-21
Implement StartTLS support for LDAP auth modules

LordAro:
Rebased version of #1888


s7design:
Hi,

Thanks for PR.

Regards
Zoran

mgruner:
Hi @LordAro, as mentioned and outlined on https://doc.otrs.com, we'll need a CLA to process any non-trivial PR. Besides that, the current implementation could be improved to allow more of the options of start_tls() to be used/configured. Thank you very much for your proposal, which stays available here, and best regards, Martin

PR #1951, opened 2018-08-21
Fix accessiblity issues

edusperoni:
This fixes the submit button not being accessible by keyboard, as well as some submenus in the customer interface.

Port from: https://github.com/OTRS/otrs/pull/1921


s7design:
Hi,

Thanks for PR.

Regards
Zoran

mgruner:
@edusperoni thank you very much for your contribution. Since your code is non-trivial, we cannot apply it without a signed CLA as described on https://doc.otrs.com/. Would you please sign it and then re-open this PR? Thank you very much and best regards, Martin

PR #1954, opened 2018-09-07
Bug 13990

amolinas:
Bug 13990
https://bugs.otrs.org/show_bug.cgi?id=13990

Link the new process ticket to the parent ticket even if there is no article in the first activity dialog.


s7design:
Hello Amolinas,

Thank you for PR. This has been merged in OTRS 6 0b97043e59abfa00602fac482b8e72d5399d7d17 It will be available from 6.0.11 version.

Regards,
Sanjin

PR #1956, opened 2018-10-23
Invalid type of comparison

ymyasoedov:
Numerical comparison is more suitable for this place, because both scalars contain number of seconds.


s7design:
Hello,

Thank you for your PR. Unfortunately development and support for OTRS 4 ceased with release of OTRS 6. For OTRS 4 only security issue updates are still supported. 

This issue does not exists in OTRS 5 or OTRS 6.  Closing this PR.

Regards,
Sanjin

PR #1957, opened 2018-11-07
Update DateChecksum.pm

TheOneAndOnlyDoctor:
Fix ticket number detection if random was used before the update to otrs 6.


s7design:
Hello @TheOneAndOnlyDoctor 

thank you for this pull request, but bug is rejected.

Regards,
Milan

PR #1958, opened 2018-11-08
Use scale in transforamtion for Notification item on hover

Link2Twenty:
Transforms are handed to the GPU so can be processed faster.

### DIFF
--- a/var/httpd/htdocs/skins/Agent/default/css/Core.Header.css
+++ b/var/httpd/htdocs/skins/Agent/default/css/Core.Header.css
@@ -434,14 +434,17 @@ body.Visible-ScreenXL #ToolBar li.UserAvatar div a:hover {
 }
 
 #Navigation > li > a:after {
-    width: 0px;
+    width: 100%;
     height: 3px;
     position: absolute;
     bottom: 0px;
     left: 0px;
     background: #f92;
     content: "";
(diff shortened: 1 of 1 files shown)


var/httpd/htdocs/skins/Agent/default/css/Core.Header.css
```diff
@@ -434,14 +434,17 @@ body.Visible-ScreenXL #ToolBar li.UserAvatar div a:hover {
...
+    width: 100%;
     height: 3px;
     position: absolute;
     bottom: 0px;
     left: 0px;
     background: #f92;
     content: "";
-    transition: width ease 0.3s;
+    transform: scale(0, 1);
+    transform-origin: 0;
```
Link2Twenty: If you remove this line the animation would spread from the centre, which, might look better.

s7design:
Hi,

Thanks for PR.

Regards
Zoran

s7design:
Hi, 

It is applied and it will be available from OTRS 6.0.15. 

There is commit:
https://github.com/OTRS/otrs/commit/1c30d5c96ab2cef26bff1eb4dcd096a739be2521

Regards
Zoran

Link2Twenty:
Thank you, I've sent the CLA now too 😀

PR #1960, opened 2018-12-11
Set NotificationRecipientEmail setting as non-required

ymyasoedov:
There is no sense to make this setting required. It's
better to give the possibility to disable it, if you
don't want to receive mail with error messages from
Daemon.


s7design:
Hi Yuri,

Thanks for PR. It is applied and it will be available form patch level OTRS 6.0.16.

https://github.com/OTRS/otrs/commit/f39318d3a2a5d2be4ade183155ed6598f5b995b1

Regards
Zoran

PR #1963, opened 2019-01-04
Fixed: script name returned by mod_perl is a directory

lermit:
Hi,

As explain here : https://bugs.otrs.org/show_bug.cgi?id=8533#c17 on some linux distro the script returned by mod_perl is limited to the sized of original process name. This is the case for alpine linux.

In this case the apache2 process is named `httpd` (5 char). `$0` will contain `/opt/`.

As you can see the patch for bug 8533 us `-e $0`. If `$0` contain a folder name the test will be true.

This PR test if the `$0` var contain is a directory path.

Regards

### DIFF
--- a/scripts/apache2-perl-startup.pl
+++ b/scripts/apache2-perl-startup.pl
@@ -31,7 +31,7 @@ BEGIN {
     # set $0 to index.pl if it is not an existing file:
     # on Fedora, $0 is not a path which would break OTRS.
     # see bug # 8533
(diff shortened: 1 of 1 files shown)


mgruner:
@s7design can you please handle this? We typically use the high-precedence `||` rather than `or`.

mgruner:
Hi @lermit. Sorry for the long delay. Merged now in 0e1aface13877904dfbeeaa9b73dc17d030a3f5b.

PR #1964, opened 2019-01-22
Make some settings non-required

ymyasoedov:
Give the possibility to disable warning notification about non-verified packages and notification with suggestion to upgrade to OTRS Business Solution.


zilibasic:
Hi,

Thanks for the PR, but we cannot accept it because it is by design and you can always in your instance of intalled OTRS change any of settings as you would like.

Regards
Zoran

PR #1763, opened 2017-05-23
User update custom fields handling

Fess-AKA-DeadMonk:
`GetUserData` returns user fields from both `users` and `user_preferences` tables, whereas `UserUpdate` takes unlimited range of parameters, but only updates few of them. It makes handling custom user parameters clouded. Some packages like **LDAP.pm** send all synced fields directly to `UserUpdate` method, where they are simply dropped.
The suggested patch makes deeper LDAP integration possible with just adding synced fields to **Config.pm** at `AuthSyncModule::LDAP::UserSyncMap` section like in the example shown below:
```
   $Self->{'AuthSyncModule::LDAP::UserSyncMap'} = {
          # DB -> LDAP
            UserFirstname => 'givenName',
            UserLastname  => 'sn',
            UserEmail     => 'mail',
            UserLogin   => 'sAMAccountName',
            UserSalutation => 'title',
            UserComment => 'description',
            UserPhone  => 'telephonenumber'
        };
```
Rows adding to the `users` table is no longer required. Just sync fields with LDAP and use any custom tags you need. If you are going to rely on LDAP solely, it would be the only thing you should do.
Thanks for your work

oh crap, it seems i've chosen the wrong base. I'm busy panicking right now, so please show mercy ^_^

PR #1967, opened 2019-02-20
Update custom user keys in UserUpdate

Fess-AKA-DeadMonk:
`GetUserData` returns user fields from both `users` and `user_preferences` tables, whereas `UserUpdate` takes unlimited range of parameters, but only updates few of them. It makes handling custom user parameters clouded. Some packages like **LDAP.pm** send all synced fields directly to `UserUpdate` method, where they are simply dropped.

The suggested patch makes deeper LDAP integration possible with just 
adding synced fields to **Config.pm** at 
`AuthSyncModule::LDAP::UserSyncMap` section like in the example shown 
below:
```
   $Self->{'AuthSyncModule::LDAP::UserSyncMap'} = {
          # DB -> LDAP
            UserFirstname => 'givenName',
            UserLastname  => 'sn',
            UserEmail     => 'mail',
            UserLogin   => 'sAMAccountName',
            UserSalutation => 'title',
            UserComment => 'description',
            UserPhone  => 'telephonenumber'
        };
```
Rows adding to the `users` table is no longer required. Just sync fields 
with LDAP and use any custom tags you need. If you are going to rely on 
LDAP solely, it would be the only thing you should do.
Thanks for your work

### DIFF
--- a/Kernel/System/User.pm
+++ b/Kernel/System/User.pm
@@ -100,7 +100,7 @@ sub GetUserData {
     if ( !$Param{User} && !$Param{UserID} ) {
         $Kernel::OM->Get('Kernel::System::Log')->Log(
             Priority => 'error',
-            Message  => 'Need User or UserID!',
+            Message  => 'Need User or UserID!'
         );
         return;
     }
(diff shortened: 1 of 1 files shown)


Fess-AKA-DeadMonk:
it's mostly the copy of [pull request](https://github.com/OTRS/otrs/pull/1763) i've made two years ago 
the former PR was broken and closed because of wrong base chosen by a <details><summary>stupid mistake</summary>
<p>
damn this fork button!
it does not make a new fork if it already exists, but just silently sends you to the existing fork
so i've based my work not on the last commit, but on some ancient one, which remained since my last contribution
</p>
</details>

Kernel/System/User.pm
```diff
@@ -585,13 +585,53 @@ sub UserUpdate {
         Value  => $Param{UserEmail}
     );
 
-    # set email address
+    # set user mobile
```
Fess-AKA-DeadMonk: this wrong comment persists for years. can we get rid of it by now?

mgruner:
Hi @Fess-AKA-DeadMonk, this was implemented in the meantime in a slightly different way in https://github.com/OTRS/otrs/commit/419fc5f2269f2640c2d65952f70000a75bee1e55. Thank you very much!

PR #1968, opened 2019-02-27
Add needed param to check

straubr:
Otherwise the DB update fails because of a foreign key constraint.

Note: Would be great if the Objects are named and used consistently.
Example: For User.pm the changing user is passed as "ChangeUserID", here it's named "UserID".
Or: to get detailed data for agents, the function is called "GetUserData" (Verb -> Object), for customerusers it's called "CustomerUserDataGet" (Object -> Verb).

### DIFF
--- a/Kernel/System/CustomerUser.pm
+++ b/Kernel/System/CustomerUser.pm
@@ -977,12 +977,14 @@ sub CustomerUserUpdate {
     my ( $Self, %Param ) = @_;
 
     # check needed stuff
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @straubr. Sorry for the long delay. This PR looks incorrect, it checks for the same key twice now (UserLogin UserLogin). Also, I'm missing a test case that shows what's failing. I'll close this for now, please feel free to follow up with more info.

PR #1970, opened 2019-04-25
Fixed: Net::SSLGlue issues warnings on modern systems (bug#12334).

NeverMin:
fixed [bug#12334](https://bugs.otrs.org/show_bug.cgi?id=12334#c20) and tested pass in Ubuntu 16.04.


milanrakic:
Hi @NeverMin 

thank you for this PR, but it is not enough for fixing the issue. It will be solved in [bug#12334](https://bugs.otrs.org/show_bug.cgi?id=12334).
This PR will be closed.

Milan

PR #1972, opened 2019-05-07
Fixed bug#14534 - Relations for certificates dont get removed in both directions.

rolfschmidt:
https://bugs.otrs.org/show_bug.cgi?id=14534


s7design:
Hi @rolfschmidt 

Thank you for PR. The fix is merged and [bug#14534](https://bugs.otrs.org/show_bug.cgi?id=14534) is closed.

Regards,
Milan

PR #1974, opened 2019-06-08
Fixed: Reply-To overwrites name from customer backend in <OTRS_CUSTOM…

Georgeto:
…ER_REALNAME> template replacement.

This bug was introduced in the fix for bug#4640.

### DIFF
--- a/Kernel/System/TemplateGenerator.pm
+++ b/Kernel/System/TemplateGenerator.pm
@@ -1793,7 +1793,7 @@ sub _Replace {
                 $From //= $Recipient{Realname};
 
                 # get real name based on reply-to
-                if ( $Data{ReplyTo} ) {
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @s7design, can you please review this?

zilibasic:
Hi Georgeto,

Thanks for PR, after QA we merged it here:
https://github.com/OTRS/otrs/commit/0d78a83a2807e6a943e3cc3ebc63e1c0af6a8f7b

There is also extended test case. Please check it and let me know if you have any other suggestion.

Regards
Zoran

zilibasic:
It will be available form OTRS 6.0.20 and OTRS 7.0.9.

Regards
Zoran

Georgeto:
That went pretty fast, thank you!

PR #1978, opened 2019-07-16
New module Kernel::System::Log::Screen and Kernel::System::Log improvement

ymyasoedov:
These two commits increase the convenience for scripting. Now you can override configured minimum log level and used log module in runtime for your scripts.

Generally speaking it's just shortcuts for `$ConfigObject->Set( Key => 'MinimumLogLevel', Value => $Value )` and `$ConfigObject->Set( Key => 'LogModule', Value => $Value )`, but now log objects can be constructed in more pleasant manner:

```perl
local $Kernel::OM = Kernel::System::ObjectManager->new(
    'Kernel::System::Log' => {
	LogModule       => 'Kernel::System::Log::Screen',
        LogPrefix       => 'OTRS-otrs.Test.pl',
	MinimumLogLevel => 'debug',
    },
);
```

Also for scripting I've added new backend `Kernel::System::Log::Screen`, that just puts messages to stdout and colorize them (unless stdout was not redirected to some file). Now you don't need do something like `tail -f /var/log/otrs/otrs.log` and it looks pretty nice: :)

screenshot

### DIFF
--- a/Kernel/System/Log.pm
+++ b/Kernel/System/Log.pm
@@ -42,6 +42,17 @@ create a log object. Do not use it directly, instead use:
             LogPrefix => 'InstallScriptX',  # not required, but highly recommend
         },
     );
+
+you can override configured minimum log level by specifying it with MinimumLogLevel param
+and choose another log module with LogModule param (for runtime only):
+
+    Kernel::System::ObjectManager->new(
+        'Kernel::System::Log' => {
+            LogModule       => 'Kernel::System::Log::File',  # not required 
+            MinimumLogLevel => 'notice',                     # not required
+        },
+    );
+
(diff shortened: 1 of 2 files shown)


ymyasoedov:
I've signed CLA and sent it to cla@otrs.com.

mgruner:
Hello @ymyasoedov. I like your idea and the output it generates a lot, thanks for the suggestion and inspiration. However, I don't agree on the implementation. Using the screen log module would prevent the log from being written to the syslog or file, which would be against the current architecture. 
Maybe in future it would be good to redesign the whole subsystem.
Thanks again and best regards, Martin

ymyasoedov:
Ok, then you need something like log multiplexer (like Log::Dispatch). I'll try to introduce it in the near future.

PR #1981, opened 2019-08-08
Fix submit article multiple times

schroedingersZombie:
https://bugs.otrs.org/show_bug.cgi?id=14723


s7design:
Hi,

As I explained in the bug it is duplicate bug.
It will be fixed very soon.

Regards
Zoran

dvuckovic:
@schroedingersZombie Thank you for your PR, we identified the real culprit and solved in another manner in 2433a13586f6d748eb3c6cd87fa51571c38c9e4f.

PR #1983, opened 2019-08-29
SelectSingleRow() for DB.pm

k-mx:
Hello!

A situation when we need only one row is rather common, and I think that special method for such cases is a good idea.
Following the style [guide](https://doc.otrs.com/doc/manual/developer/6.0/en/html/code-style-guide.html#database-using-while-loop), we have built-in while loop (there is some places where `FetchrowArray() `not wrapped with loop). But regardless [guide](https://doc.otrs.com/doc/manual/developer/6.0/en/html/code-style-guide.html#database-using-limit), I considered to omit hardcoded  `Limit => 1` in new method.
I think it will help to track weird bugs:
```
$Self->{DBObject}->Prepare(
    SQL   => 'SELECT id FROM foo WHERE name = ?',
    Bind  => [ \$Username ],
    Limit => 1,
);
```
If name somehow not unique, we will have random id from call to call. So, `Limit => 1` makes sense only with `ORDER BY`

Thank you for your attention!

### DIFF
--- a/Kernel/System/DB.pm
+++ b/Kernel/System/DB.pm
@@ -886,6 +886,49 @@ sub SelectAll {
     return \@Records;
 }
 
+=head2 SelectSingleRow()
+
+returns single row from a SELECT statement.
+In essence, this calls Prepare() and FetchrowArray() to get single record.
+
+    my $ResultAsArrayRef = $DBObject->SelectSingleRow(
+        SQL  => 'SELECT id, name FROM table WHERE name = ?',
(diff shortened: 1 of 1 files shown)


mgruner:
Hi @k-mx. Thanks for your proposal! I agree that a similar method may be useful. LIMIT => 1 might still be helpful for databases to optimize their queries.
So I won't merge this now, also because we'd need test coverage and apply this method throughout the codebase. Thank you anyway!
Best regards, Martin

PR #1984, opened 2019-09-06
Improved performance of TicketQueueOverview dashboard component.

mback2k:
Running a simple COUNT query per combination of Queue and State is
a lot faster than retrieving all TicketIDs per State and
then counting these TicketIDs per QueueID, since passing in a huge
list of TicketIDs to SQL is quite slow and somewhat redundant.

We have a dashboard queue overview matrix of 24 queues in 9 states
and this solution reduces the initial load time to less than 10 %.

Also, the COUNT queries do not require an upper limit at 100.000.

### DIFF
--- a/Kernel/Output/HTML/Dashboard/TicketQueueOverview.pm
+++ b/Kernel/Output/HTML/Dashboard/TicketQueueOverview.pm
@@ -142,26 +142,16 @@ sub Run {
 
     my %Results;
     for my $StateOrderID ( sort { $a <=> $b } keys %ConfiguredStates ) {
-
-        # Run ticket search for all Queues and appropriate available State.
-        my @StateOrderTicketIDs = $TicketObject->TicketSearch(
-            UserID   => $Self->{UserID},
-            Result   => 'ARRAY',
-            QueueIDs => \@QueueIDs,
-            States   => [ $ConfiguredStates{$StateOrderID} ],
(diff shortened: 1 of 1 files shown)


mgruner:
Hello @mback2k. I cannot apply this to master as the agent interface is being rewritten with another technology at the moment for OTRS 8, but I asked my colleagues to evaluate if your optimization can still be applied there. Thank you very much and best regards, Martin

mgruner:
@mback2k I can confirm this optimization was included in OTRS 8.

mback2k:
@mgruner But what about those who are still running on OTRS 6? I would like to avoid having to maintain this patch locally.

mgruner:
@mback2k our policy for stable versions is to only make required changes (bugs, security issues, performance problems etc.). I haven't received reports so far that the queue overview is a performance problem that needs to be solved (=bug). All changes can have side effects and must be carefully tested, it's not just a simple git merge. 
(In this case, you omitted the Limit, which could cause a change of behaviour in certain circumstances.)
Best regards, Martin

mback2k:
@mgruner I can create a bug report if you require that to fix the performance issue. I can also re-add the limit, but then it would be per queue and status instead of just an overall limit. I understand that all changes can have side effects, but our dashboard loading time went down from >10s to ~1s.

PR #1985, opened 2019-09-20
Update AdminSalutation.tt

ymyasoedov:
Make a comment text translatable.


mgruner:
Hello Yuri, thank you for your proposal. Translating the value before it gets inserted in the input field is definitely incorrect, as it can cause an unwanted update in the database.
Even in the output, in many cases we cannot translate user provided data like the name and the comment in this case. Maybe in a future version of OTRS this is handled differently, but at present I cannot apply this PR.
Thank you very much and best regards, Martin

PR #1986, opened 2019-10-04
Confusing comment in regex

k-mx:


mgruner:
@k-mx nice catch, applied, thanks!

PR #1987, opened 2019-10-04
Cluster suddenly dissapears in AgentTicketZoom

k-mx:
This bug works if cluster is the last element
output


mgruner:
@s7design this seems to be a bug report. Can you handle it, creating a bugzilla entry if neede, please?

s7design:
Hi Maxim,

Thanks for PR. It will be fixed as bug report.
You can follow communication on this bug report

https://bugs.otrs.org/show_bug.cgi?id=14824

Regards
Zoran

PR #1989, opened 2019-10-06
Time of previous history line always an integer

uid66:


mgruner:
Hello @uid66, thank you for your proposal. Which particular problem are you solving in this PR? That's not clear to me yet.

PR #1995, opened 2019-10-06
UserID is not used in HistoryGet

uid66:


mgruner:
Hello @uid66. You are right that the UserID is not currently used in HistoryGet(). But I don't want to change now the API in release branches, and there is still the use case of possible customizations to this method making additional API calls from inside which would require UserID.
Best regards, Martin

PR #2000, opened 2019-10-23
Fix incorrect error message in Kernel::System::PID

ymyasoedov:
TTL is not necessary 1 hour. Error message is not valid.


mgruner:
Dear @ymyasoedov, congratulations to PR#2000 🎉 ! Your change was merged in f585e59959e3ea357ea55f77f0d5f4d5dc5dc292 with small improvements.

PR #2001, opened 2019-10-25
Fix Installation failure. This break occurs, because the  user could …

mpahnke:
… not be

created.
Detail: Fixed the position of the userlogin Parameter for useradd:
Moved the userlogin parameter to the most last commandline position,
where it is expected according the POSIX standard
https://www.unix.com/man-page/posix/8/useradd/
(and  linux works mostly to that standard too)

### DIFF
--- a/scripts/auto_build/spec/templates/includes/redhat-systemd-base.tt
+++ b/scripts/auto_build/spec/templates/includes/redhat-systemd-base.tt
@@ -53,7 +53,7 @@ if id $OTRSUSER >/dev/null 2>&1; then
     [% '#' %] update home dir
     usermod -d /opt/otrs $OTRSUSER
 else
-    useradd $OTRSUSER -d /opt/otrs/ -s /bin/bash -g apache -c 'OTRS System User' && echo "$OTRSUSER added."
+    useradd -d /opt/otrs/ -s /bin/bash -g apache -c 'OTRS System User' $OTRSUSER && echo "$OTRSUSER added."
 fi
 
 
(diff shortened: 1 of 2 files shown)


mgruner:
Hello @mpahnke. On which platform does this fail?

mpahnke:
Hello @mgruner. At least on all OS on which the rpm should be installed, and where the otrs-system user does not exist already. That means on RHEL (CentOS 7 and 8) and SLES maybe. (I've no SLES test system currently) - But I could try to build a test system on the SUSE Studio Build Platform.

mpahnke:
Hello @mgruner. At least on all OS on which the rpm should be installed, and where the otrs-system user does not exist already. That means on RHEL (CentOS 7 and 8) and SLES maybe. (I've no SLES test system currently) - But I could try to build a test system on the SUSE Studio Build Platform.

I tested on opensuse - and it seems, that useradd takes a username login param on these platform regardless of its position, unless it is prefixed with a  '-'. But this is still against its own useradd -help 
which explains as follow:

Usage: useradd [options] LOGIN
       useradd -D
       useradd -D [options]

mgruner:
Hello @mpahnke. Yes, it's against the spec, but it works also. I tested it on RHEL 7, where it also accepts it in first position. We didn't have problems with these RPMs so far. That's why I asked where you saw this failing.

mgruner:
Hello @mpahnke. Thanks again for your proposal here. Since we don't see currently where exactly it is failing, and seems to be working instead, and since RPMs are discontinued with OTRS 7, I'm closing this PR now. Please feel free to provide more info!

PR #2002, opened 2019-10-31
Remove unused array

ymyasoedov:
Array initialized by copying attachments is not used. It can be deleted.


s7design:
Hi,

Thanks it is applied.
There is commit:
https://github.com/OTRS/otrs/commit/6654c62d5f5a851c70bb0958281bfcc0fafcb961 

Regards
Zoran

PR #2003, opened 2019-12-04
No need to create anonimous array

k-mx:
Will work slightly faster


mgruner:
Hi @k-mx. Nice catch, I applied this fix now for the master branch, also in CustomerCompany/DB.pm. It should be semantically equivalent, just a little faster.
Best regards, Martin

PR #2005, opened 2019-12-21
Possible fix for AgentCreateBy event

k-mx:
I think it must be triggered regardless of SenderType

I.e. If we create ticket from AgentTicketPhone, event will always fail (sender type is customer or system)


mgruner:
Hello @k-mx. Sorry, I don't understand the problem (yet) that you are trying to solve. The current code looks ok: agent who created the article is only used if the first article was indeed created by an agent. We'd need a bug report describing what is wrong here.
Additionally, your code uses `1` as a hardcoded user id. Is the purpose to exclude tickets created by root, or by postmaster (the user id of email tickets is configurable via `PostmasterUserID`)?
I'll close this PR for now, please feel free to add more information. Best regards, Martin

PR #2006, opened 2019-12-28
Fix limit log entries in AdminLog module

ymyasoedov:
GetLog() method doesn't accept any parameters, so passing Limit param is useless.


ymyasoedov:
It's small bug fixing. From my point of view it's better to move limit value into settings.

mgruner:
Applied in 76b5176ee86b4f3c6dcc5af2b2dc2fd74e88586c, thanks a lot @ymyasoedov.

PR #2008, opened 2020-02-19
Fix bug in method CheckEmail of Kernel::System::CheckItem

ymyasoedov:


s7design:
Hi Yuri,

thanks for PR. It is fixed here https://github.com/OTRS/otrs/commit/becdabf6d1cb5ed28b23fe28a63c9f933e9ee4b2

Regards
Zoran

PR #2011, opened 2020-03-19
Make user of multiple companys visible in CIC

straubr:
The manual describes how one can add a customer user to multiple customers:
https://doc.otrs.com/doc/manual/admin/6.0/en/html/external-backends.html

They are however not listed in the customer Information center of the "alternative" company. This patch will also list the users if they have an alternative company ID.

Requires "CustomerIDs" to be set in the Config.pm Map and a valid table, as per the manual.

I'm not happy with the replace, but the manual specifies that the IDs have to be semicolon separated and "FIND_IN_SET" only works with comma lists.


mgruner:
Dear @straubr,
Best regards, Martin

PR #2017, opened 2020-06-08
CREATE USER in MySQL 8

bschmalhofer:
Hi, I tried to install ((OTRS)) Community Edition on Ubuntu 20.04 runnint MySQL 8.0.20-0ubuntu0.20.04.1.
In this setup `Kernel::Modules::Installer` complained because of `GRANT ALL PRIVILEGES` statement. Ich changes the user creation statement such that it works with MySQL 8. This should also work with MySQL 5.7 and MariaDB.

Note that FLUSH PRIVILEGES is not needed unless there are direct manipulations of the ,ysql.user table.

### DIFF
--- a/Kernel/Modules/Installer.pm
+++ b/Kernel/Modules/Installer.pm
@@ -505,8 +505,8 @@ sub Run {
 
                 @Statements = (
                     "CREATE DATABASE `$DB{DBName}` charset utf8",
-                    "GRANT ALL PRIVILEGES ON `$DB{DBName}`.* TO `$DB{OTRSDBUser}`\@`$DB{Host}` IDENTIFIED BY '$DB{OTRSDBPassword}' WITH GRANT OPTION",
-                    "FLUSH PRIVILEGES",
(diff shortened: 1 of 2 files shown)


jensoliver:
I think this is obsolete as OTRS 6 does not support MySQL8+

bschmalhofer:
Yes, please discard that PR. I also just ran into an issue where the default authentication method 'caching_sha2_password' gave an error because the database connect isn't over a secure channel.

I can't believe that a CREATE USER statement that works in MySQL 5.7, MySQL 8 and MariaDB is so hard.

s7design:
Hi,

you are right there is an issue with MySQL 8 on OTRS6. You can find more info on the bug

https://bugs.otrs.org/show_bug.cgi?id=13866 and all linked bug in see also and duplicates bug section.

It will work from OTRS7.

Regards
Zoran

s7design:
Hi Bernhard,

There is info about support version for MySQL for OTRS6.

https://github.com/OTRS/otrs/blob/rel-6_0/README.md

Regards
Zoran

PR #2019, opened 2020-06-12
Viewable tickets page

bschmalhofer:
The sysconfig setting ViewableTicketsPage seems to be no longer used. It looks like the number of displayed tickets is now taken from the user preferences.
I also removed manually the now obsolete translation. There is propably a tool for that, but I'm clueless about it.


s7design:
Hi Bernhard,

Thanks for PR you are right this config is not used so long. It is removed in OTRS8. There is no support for OTRS6 except security issue so it will not be changed there.

Regards
Zoran

PR #2021, opened 2020-06-19
improve caching performance of LinkListRaw

felixjogris:
Hi,

on an OTRS installation with thousands of linked and merged (read: parent/child linked) tickets, Devel::NYTProf revealed that most CPU time is spent in _LinkListRaw(). Few seconds after calling "otrs.Console.pl Maint::Cache::Delete", two cache files appeared under var/tmp/CacheFileStorable/LinkObject/, each about 1 MB in disk size. Deserialising them with Storable::thaw and Data::Dump'ing the data showed that these two cache files contain all normally and parent/child linked ticket ids, respectively.
This patch avoids caching the whole link_relation table and looping over all entries in Perl by adding the source_key or target_key of the link object to the CacheKey.

BR,
Felix

### DIFF
--- a/Kernel/System/LinkObject.pm
+++ b/Kernel/System/LinkObject.pm
@@ -552,7 +552,8 @@ sub LinkAdd {
             'Cache::LinkListRaw'
             . '::Direction' . $Direction
             . '::ObjectID' . $Param{ $Direction . 'ObjectID' }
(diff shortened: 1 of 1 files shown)


mgruner:
Dear @felixjogris,
Best regards, Martin

PR #2022, opened 2020-06-26
Avoid $&, $', and $` if you can (perlfaq6)

k-mx:
Similar PR was taken to the [original repo](https://github.com/silnrsi/font-ttf/pull/7) and was discussed there. Usage of mentioned variables can lead to performance issues and increase memory consumption.


mgruner:
Hi @k-mx, thank you very much for your proposal. We don't patch our Kernel/cpan-lib dependencies unless absolutely neccessary. Also, OTRS 6 by policy only receives security fixes at this point.
Best regards, Martin

PR #2028, opened 2020-07-23
Fix HTTP::REST Transport error if request body is empty

sternenseemann:
Previously if you'd perform an request using the HTTP::REST backend that had a empty request body (i. e. { Data => {}, }), it would fail with this error message:

```
Not a SCALAR reference at /opt/otrs/bin/cgi-bin/../../Kernel/cpan-lib/LWP/Protocol/http.pm line 256.
```

This is caused by REST::Client missing its body argument. The solution is to pass undef as a body argument if our body turns out to be empty.

I've sent the CLA a few moments ago via email.

### DIFF
--- a/Kernel/GenericInterface/Transport/HTTP/REST.pm
+++ b/Kernel/GenericInterface/Transport/HTTP/REST.pm
@@ -742,8 +742,16 @@ sub RequesterPerformRequest {
     }
     push @RequestParam, $Controller;
 
-    if ( IsStringWithData( $Param{Data} ) ) {
-        $Body = $Param{Data};
+    # Only POST, PUT or PATCH have a body. If it is empty
+    # (i. e. $Param{Data} = {}), undef is passed to REST::Client.
+    if ( $RestCommand eq 'POST'
+         || $RestCommand eq 'PUT'
+         || $RestCommand eq 'PATCH'
+         ) {
+        if ( IsStringWithData( $Param{Data} ) ) {
+            $Body = $Param{Data};
+        }
+
         push @RequestParam, $Body;
     }
 


Kernel/GenericInterface/Transport/HTTP/REST.pm
```diff
@@ -690,7 +690,8 @@ sub RequesterPerformRequest {
     my $JSONObject   = $Kernel::OM->Get('Kernel::System::JSON');
     my $EncodeObject = $Kernel::OM->Get('Kernel::System::Encode');
 
-    my $Body;
+    # Body defaults to undef, otherwise REST::Client crashes if Param{Data} is {}
+    my $Body = undef;
```
rgci, line 694: Why do you introduce this inconsistency? On all other variable initialisation in the given file undef is not explicitly set. (Because the default is „undef“)

Kernel/GenericInterface/Transport/HTTP/REST.pm
```diff
@@ -744,9 +745,10 @@ sub RequesterPerformRequest {
 
     if ( IsStringWithData( $Param{Data} ) ) {
         $Body = $Param{Data};
-        push @RequestParam, $Body;
     }
 
+    push @RequestParam, $Body;
```
rgci, line 750: Did you test this? You are always pushing the value of $Body even when this is a GET call which would propably cause unwanted behaviour... See https://metacpan.org/pod/REST::Client#GET-(-$url,-%5B%25$headers%5D-)

Kernel/GenericInterface/Transport/HTTP/REST.pm
sternenseemann: I wanted to make it explicit that there are cases in which this variable keeps the `undef` value. The whole behavior of `REST::Client` and `Transport::HTTP::REST` in combination is non obvious in this case, so I was trying to make it clearer from the code.

If necessary I can revert this change.

Kernel/GenericInterface/Transport/HTTP/REST.pm
sternenseemann: Indeed, that is an oversight, I've pushed a fixed version of this commit.

Let me know what you think. I'm not entirely satisfied, since it kind of duplicates some logic, but this way it keeps the diff smaller and avoids restructuring the Body/RequestParams building logic.

mgruner:
Dear @sternenseemann,
Best regards, Martin

PR #2030, opened 2020-09-08
Set autocomplete to 'new-password' for admin interface password fields.

thijskh:
This signals more accurately to password managers that this is not a
place to be putting in any existing passwords.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete


mgruner:
Dear @thijskh,
Best regards, Martin

PR #2027, opened 2020-07-23
Fixed typo error in VariableCheck POD

ymyasoedov:
Fix typo error

PR #2031, opened 2020-09-17
Small fix on VariableCheck documentation

richieri:


ilizoran:
Hi @richieri,

It is fixed in meantime. OTRS6 is not supported so it is fixed only for OTRS7 and onwards.

https://github.com/OTRS/otrs/pull/2027

Regards
Zoran

s7design:
Hi @richieri,

It is fixed in meantime. OTRS6 is not supported so it is fixed only for OTRS7 and onwards.

#2027

Regards
Zoran

